Issue #430, opened 2014-06-26
need option to disable the 'model answer' functionality for questions


moloko:
showing the user the correct answer is not always desirable - at the moment there's no way to turn this off via the json

taylortom:
Merged.

PR #488, opened 2014-12-09
Add ability to turn off model answers for questions

taylortom:
See https://github.com/adaptlearning/adapt_framework/issues/430 for info.


dancgray:
+1

PR #493, opened 2015-01-06
Hard/soft reset implementation

olivermartinfoster:
Each model has two new attributes. _canReset and _isInteractionComplete.

_canReset stops a component from being reset by default. 

Each model has a new function called ‘reset’ which takes two parameters ‘type’ and ‘force’. 
Type can be ‘hard’/true or ‘soft’. 
The optional ‘force’ parameter over-rides the behaviour of _canReset. 
This function will only be performed on the model on which it is called.

_isInteractionComplete and _isComplete now cascade from component to page on both true and false. i.e. if a component is set to _isInteractionComplete: false, then its parents will cascade accordingly down to menu level in the same way it does if _isComplete: true.

Each view has a new function called ‘resetCompletionStatus’ which takes a single parameter ‘type’. 
Type can be ‘hard’/true, ‘soft’. 
This function performs a reset on its highest children. When enacted at a page level it will reset all of the components within that page. These component statuses will then cascade downward according to the cascade behaviour above.

checkIfResetOnRevisit will now force the reset function from the model.
checkIfResetOnRevisit can take model attributes values such as _resetOnRevisit: ‘hard’/true or ‘soft’ keeping its original behaviour intact.

The visual states of the buttons and attempts calculations are dependent upon the _isInteractionComplete attribute instead of the _isComplete attribute.

### DIFF
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -13,7 +13,9 @@ define(function (require) {
 
         defaults: {
             _canShowFeedback: true,
+            _canReset: false,
             _isComplete: false,
+            _isInteractionComplete: false,
             _isEnabled: true,
             _isResetOnRevisit: false,
             _isAvailable: true,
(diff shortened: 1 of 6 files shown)


olivermartinfoster:
From discussion: https://community.adaptlearning.org/mod/forum/discuss.php?d=475

taylortom:
+1 for this implementation

oliverfoster:
Needs reset restriction based on _canReset hierarchy in adaptView resetCompletionStatus.

darylhedley:
-1 as the model data beneath the view is wrong and lies about expectations which has been raised as a concern by a few core team developers here:
https://community.adaptlearning.org/mod/forum/discuss.php?d=475

moloko:
+1

brian-learningpool:
+1

dancgray:
+1

svenlaux:
Thank you very much to Ollie for preparing the documentation/diagrams and talking this through with the team. Thanks also to Brian and Dan for reviewing and giving your +1. We have been stuck on this item for a while (issues identified and discussed but no viable solution put forward) and it's good to see us moving past this as it was now holding up the Framework version 2.0 release.

We will now merge this in and add the work on assessments for the v2.0 release.

PR #502, opened 2015-01-13
Revert grunt-contrib-handlebars to ~0.5.11

taylortom:
Can't run course with the current handlebars version

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "grunt-concurrent": "~1.0.0",
     "grunt-contrib-connect": "~0.9.0",
     "grunt-contrib-copy": "~0.7.0",
-    "grunt-contrib-handlebars": "~0.9.2",
+    "grunt-contrib-handlebars": "~0.5.11",
     "grunt-contrib-less": "~1.0.0",
     "grunt-contrib-requirejs": "~0.4.4",
     "grunt-contrib-watch": "~0.6.1",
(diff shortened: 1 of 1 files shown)


oliverfoster:
Tested and working with the fix.

moloko:
+1

hrajotia:
@taylortom according to this http://github.com/gruntjs/grunt-contrib-handlebars#release-history, they have updated Handlebars to version 2.0. So I thinks this what making trouble here so if we change version of `grunt-contrib-handlebars` to `~0.8.0` then it should work.

what you think?

oliverfoster:
We would have to check if the client and grunt versions are compatible. Have you tested .8.0 with the grunt task to see if the test course works? @hrjotia

oliverfoster:
To be honest I'd like to see us upgrade to version 2 in core libraries and grunt but include the full library rather than just the runtime.

hrajotia:
@oliverfoster - yeah `"grunt-contrib-handlebars": "~0.8.0"`is working for me.

oliverfoster:
also confirmed working with ~0.8.0, looks like we're good

taylortom:
Done.

I'm assuming that this will be revisited with this ticket: https://adaptlearning.atlassian.net/browse/ABU-383?filter=10202, so as long as the build works now, I don't think it will matter much which version we're on.

PR #509, opened 2015-01-20
Set correct version

taylortom:


taylortom:
This'll also need to be done for develop

brian-learningpool:
+1

oliverfoster:
+1

darylhedley:
Sorry - this is the wrong number
should be same as release: Adapt framework v1.1.1 as seen here
https://github.com/adaptlearning/adapt_framework/releases

hrajotia:
+1

Issue #513, opened 2015-01-21
Remove backboneModel.js


oliverfoster:
#507 

There are few use cases where extensions will conflict over variable settings and therefor where the backboneModel code is used. It seems sensible to remove this code.

The removal would improve the large quantity of set operations that occur in Adapt.

Allowing the inverse behaviour seems a more sensible way of achieving an attribute lock, by allowing a plugin to force an attribute lock rather than having the framework check for complex locks on each set operation.

Suggestion for similar behaviour:  
     1. _isLocked { attributeName: true } on the model with a simpler validator (still checking on every set but using a much smaller mechanism)

Suggestions for inverse behaviour:  
   1. trigger "prechange:_attributeName" event and allow for cancellation (allows plugins to take control over the change handling)
   2. change "change:_attributeName" event to allow for cancellation (cannot guarantee event order)

zenduo:
Is this still an issue?

oliverfoster:
Yea. I just need some support on it.

taylortom:
(just updated PR to reference this)

zenduo:
Moved to  https://adaptlearning.atlassian.net/browse/ABU-474

Issue #507, opened 2015-01-16
Performance related fixes


oliverfoster:
Over use of underscore helpers in the core. This leads to an exponential growth of resource requirements when used with course material arrays. I.e. the problem gets more apparent the more course material exists. We need to look at the most used Adapt functions and remove as many utility function calls (such as uneccesary _.each calls) where possible. The core should ideally be as small and fast as possible giving more resources to components and extensions as they are likely to be changed more frequently.

oliverfoster:
Examples:

https://github.com/adaptlearning/adapt_framework/blob/develop/src/core/js/models/backboneModel.js#L27-L34
https://github.com/adaptlearning/adapt_framework/blob/develop/src/core/js/models/backboneModel.js#L49-L66

oliverfoster:
Reference:

https://github.com/jashkenas/underscore/blob/1.5.2/underscore.js#L76-L90

oliverfoster:
What this demonstrates is that a function closure has to be created for each item in the for loop and that the this context has to be carried through.

Simply using native for syntax solves this resource hoging operation speedying up processing and reducing memory consumption. 

_.each should not be used in frequently used function calls on large arrays.

oliverfoster:
http://jsperf.com/jquery-each-vs-underscore-each-vs-for-loops/11  < example from @moloko

moloko:
That's just someone else's I found. here's a slightly tidier one: http://jsperf.com/underscore-each-versus-for-loop

taylortom:
Also worth watching out for any code doing the length calculation like this:
```
for(var i = 0; i < arr.length; i++) {
}
```

oliverfoster:
Yuup, I am also one who fondly writes iterative length calculations... learn something new everyday!

From now on:

```
for (var i = 0, length = arr.length; i < length; i++) {

}
```

oliverfoster:
Backbone collections:

```
for (var i = 0, length = collection.models.length; i < length; i++) {
   var model = collection.models[I];

}
```

zenduo:
Is this one still an issue?

PR #514, opened 2015-01-21
Performance issues

oliverfoster:
Changed the _.each syntax for the for (var i = 0, len = x; i < len; i++) syntax,

<del>Removed backboneModel.js</del>
<br/>

Mapped Adapt._id attributes to stop findWhere / where operations for id's


oliverfoster:
#507

taylortom:
+1 for the concept, will test out when I get a min

chuck-n0rr1s:
#513

darylhedley:
-1 as it removes core functionality that hasn't been replaced. Our plugin architecture relies on this code

oliverfoster:
Added core functionality back in.

oliverfoster:
The plugin architecture does not rely on the previously removed code as removing it doesn't break anything and there are no direct dependencies upon it. No plugins we know of [so far] require or utilise the previously removed code either, hence the suggestion to remove it.

dancgray:
Performance enhancements seem good, but having problems running this in IE8. Just hangs when trying to load pages, no errors. Admittedly it is a VM.

oliverfoster:
missed variable uniqueness requirement should solve that daryl, was getting stuck in a fun loop

dancgray:
+1

taylortom:
+1, Noticeably faster in Win7 IE8

moloko:
+1

PR #517, opened 2015-01-22
Fixes #516 - allows selection of specific menu

### LINKED ISSUE #516: Unable to select a specific menu when more than one is installed
Similar to theme selection, Gruntfile.js should be updated to allow support for selecting a specific menu when more than one is available.  As this is already implemented for themes I'm assuming this is an oversight for menu selection.

brian-learningpool:
This is very similar to a previous change for themes to allow
selection of a specific menu, when more than one is installed.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,7 +1,8 @@
 module.exports = function(grunt) {
     require('matchdep').filterDev('grunt-*').forEach(grunt.loadNpmTasks);
     var outputdir = grunt.option('outputdir') || '',
-        theme = grunt.option('theme') || 'adapt-contrib-vanilla';
+        theme = grunt.option('theme') || 'adapt-contrib-vanilla',
+        menu = grunt.option('menu') || 'adapt-contrib-boxMenu';
 
     if (outputdir) {
         if (outputdir.substring(outputdir.length - 1, outputdir.length) !== '/') {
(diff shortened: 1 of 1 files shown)


oliverfoster:
+1 with amend to allow all menus to be included as before or a specific menu as per pr.

oliverfoster:
```
            menuAssets: {
                files: grunt.file.expand(['src/extensions/<%= menu %>/']).map(function(cwd) {
                    return {
                        expand: true,
                        src: ['**'],
                        dest: '<%= outputdir %>build/assets/',
                        cwd: cwd + 'assets/',
                        filter: "isFile"
                    };
                })
            },
```

Can we try this out for the menuAssets?

brian-learningpool:
It doesn't seem to work as expected.  Can this be raised as a new ticket?

oliverfoster:
i'm alright with that as long as i can get a +1 with a multiple vs single implementation a bit later, this doesn't immediately affect us, but it will do when adapt is released

oliverfoster:
i'd quite like to bring both themes and menus allowing both multiples and singles so that we can eventually have themes which switch out for accessibility and multiple menus depending on device +/ sub-menu requirements, they are awesome features it'd be nice to include.

oliverfoster:
+1 as is.

darylhedley:
+1

brian-learningpool:
I agree with multiple menus and themes being a really nice feature but we should probably spend more time working out the exact requirements here.  

I've raised #520 regarding the hard-coded plugin names.

dennis-learningpool:
+1

PR #518, opened 2015-01-22
Remove backboneModel.js

oliverfoster:
#513 #514


oliverfoster:
https://github.com/adaptlearning/adapt_framework/issues/507

taylortom:
+1 for removing this. If still needed, I'd like to see the functionality re-thought.

moloko:
What we really need is some sort of implementation of Object.freeze() or Object.seal() that is compatible with IE8

moloko:
worth trying this? https://github.com/inexorabletash/polyfill/blob/master/es5.js
we don't necessarily need to include all of it

oliverfoster:
i'd like to have indexOf from that library also

moloko:
hmm sadly it doesn't look like it includes the option to stop a property from being writable. and from what I have read there is no shim for object.freeze/seal

bloody JavaScript! cursèd IE8!!

oliverfoster:
Original Code:

```
lockedAttributes: {

},

validate: function (attrs, options) {  
    //OLLIE: Default processing path for all attributes

    // This is returned and is set on model
    var returnObject = {};
    // Object containing any locked attributes trying to be set on the model
    var lockedAttributes = {};
    var pluginName = options.pluginName;

    // Go through each attribute being set and check if it's in the list of lockedAttributes
    // Add it to the lockedAttributes object
    //OLLIE: Each is expensive
    _.each(attrs, function(value, key) {
        //OLLIE: Has is expensive
        var hasLockedAttribute = _.has(this.lockedAttributes, key);

        if (hasLockedAttribute) {
            lockedAttributes[key] = value
        }

    }, this);

    // Check if there are locked attributes
    //OLLIE: Empty is expensive
    if (!_.isEmpty(lockedAttributes)) {
        //OLLIE: Default processing path for any locked attribute

        // Check if options has a plugin name
        if (!pluginName) {
            var validateError = 'When setting a locked attribute on the model please specify a pluginName in the options object';
            var lockedAttributesKeys = _.keys(lockedAttributes);
            // Leave this console in - is used for debugging
            console.log(validateError, lockedAttributesKeys);
            this.validatedAttributes = _.omit(attrs, lockedAttributesKeys);
            return validateError;
        }

        // Go through each lockedAttributes and set on this.lockedAttributes
        _.each(lockedAttributes, function (value, key) {
            this.lockedAttributes[key][pluginName] = value;
            //OLLIE: Filter is expensive
            var trueAttributes = _.filter(this.lockedAttributes[key], function (value, key) {
                // Return all attributes that have true values
                return value === true;
            });

            // Check if attribute should be set to true
            //OLLIE: Size is expensive
            if (_.size(this.lockedAttributes[key]) === trueAttributes.length) {
                lockedAttributes[key] = true;
                // Check if attribute should be set to false
            } else if (trueAttributes.length === 0) {
                lockedAttributes[key] = false;
                // Check if attribute should not be set
            } else {
                lockedAttributes[key] = this.get(key);
            }
        }, this);

        //return
        //OLLIE: Extend is expensive
        var returnAttributes = _.extend(attrs, lockedAttributes);
        this.validatedAttributes = returnAttributes;
        //return;
    } else {
        this.validatedAttributes = attrs;
        //return;
    }

},

isValid: function () {
    return console.log('Sorry, This is not supported in Adapt');
},

_validate: function (attrs, options) {
    //OLLIE: Default processing path
    // If reset is passed as an option forget validating
    // This is to overcome the initial load of models and collections
    if (options.reset) {
        return true;
    }

    // Check if there's a validateError
    var error = this.validationError = this.validate(attrs, options) || null; //OLLIE: Runs validate function every set call
    if (!error) return true;
    this.trigger('invalid', this, error, _.extend(options, {validationError: error}));
    return false;
},

set: function (key, val, options) {
    var attr, attrs, unset, changes, silent, changing, prev, current;
    if (key == null) return this;

    // Handle both `"key", value` and `{key: value}` -style arguments.
    if (typeof key === 'object') {
        attrs = key;
        options = val;
    } else {
        (attrs = {})[key] = val;
    }

    options || (options = {});

    // Run validation.
    this._validate(attrs, options);

    // If validatedAttributes exist use these instead of the attributes passed in
    if (this.validatedAttributes != null) {
        attrs = this.validatedAttributes;
    }

    // Extract attributes and options.
    unset = options.unset;
    silent = options.silent;
    changes = [];
    changing = this._changing;
    this._changing = true;

    if (!changing) {
        this._previousAttributes = _.clone(this.attributes);
        this.changed = {};
    }
    current = this.attributes, prev = this._previousAttributes;

    // Check for changes of `id`.
    if (this.idAttribute in attrs) this.id = attrs[this.idAttribute];

    // For each `set` attribute, update or delete the current value.
    for (attr in attrs) {
        val = attrs[attr];
        if (!_.isEqual(current[attr], val)) changes.push(attr);
        if (!_.isEqual(prev[attr], val)) {
            this.changed[attr] = val;
        } else {
            delete this.changed[attr];
        }
        unset ? delete current[attr] : current[attr] = val;
    }

    // Trigger all relevant attribute changes.
    if (!silent) {
        if (changes.length) this._pending = true;
        for (var i = 0, l = changes.length; i < l; i++) {
            this.trigger('change:' + changes[i], this, current[changes[i]], options);
        }
    }
    // Clear validatedAttributes cache
    this.validatedAttributes = null;
    // You might be wondering why there's a `while` loop here. Changes can
    // be recursively nested within `"change"` events.
    if (changing) return this;
    if (!silent) {
        while (this._pending) {
            this._pending = false;
            this.trigger('change', this, options);
        }
    }
    this._pending = false;
    this._changing = false;
    
    return this;
}
```

preChanged event suggestion:
```
Backbone.Model.prototype.set = function(key, val, options) {
  var attr, attrs, unset, changes, silent, changing, prev, current;
  if (key == null) return this;

  // Handle both `"key", value` and `{key: value}` -style arguments.
  if (typeof key === 'object') {
    attrs = key;
    options = val;
  } else {
    (attrs = {})[key] = val;
  }

  options || (options = {});

  // Run validation.
  if (!this._validate(attrs, options)) return false;

  // Extract attributes and options.
  unset           = options.unset;
  silent          = options.silent;
  changes         = [];
  changing        = this._changing;
  this._changing  = true;

  if (!changing) {
    this._previousAttributes = _.clone(this.attributes);
    this.changed = {};
  }
  current = this.attributes, prev = this._previousAttributes;

  // Check for changes of `id`.
  if (this.idAttribute in attrs) this.id = attrs[this.idAttribute];

  //OLLIE: use a proxy 'changed' object to stop changes being logged before preChange is handled.
  var changed = {};

  options.cancel = false;

  // For each `set` attribute, update or delete the current value.
  for (attr in attrs) {
    val = attrs[attr];
    if (!_.isEqual(current[attr], val)) {
    	if (!silent) {
    		//OLLIE: Default processing path
    		//OLLIE: Trigger preChange on each changed attribute
			this.trigger('preChange:' + attr, this, val, options);
			//OLLIE: If options object comes back with cancel === true, skip changes and return false to set function
			if (options.cancel === true) return false;
		}
		changes.push(attr);
	}

    if (!_.isEqual(prev[attr], val)) {
      changed[attr] = val;
    } else {
      delete changed[attr];
    }

  }

  //OLLIE: Default processing path
  //OLLIE: Update original object from changed proxy
  //OLLIE: Slightly adds to processing time
  for (var changeAttr in changed) {
  	this.changed[changeAttr] = changed[changeAttr];
  }

  // Trigger all relevant attribute changes.
  if (!silent) {
  	//OLLIE: Default processing path
    if (changes.length) this._pending = options;
    //OLLIE: Perform changes noisily
    for (var i = 0, l = changes.length; i < l; i++) {
      //OLLIE: Does not add to processing time as identical to original behaviour, just in a different order
      var attr = changes[i];
      unset ? delete current[attr] : current[attr] = attrs[attr];
      this.trigger('change:' + changes[i], this, current[changes[i]], options);
    }
  } else {
  	//OLLIE: Not default processing path
  	//OLLIE: Perform changes silently
  	//OLLIE: Slightly adds to the processing time
  	for (var i = 0, l = changes.length; i < l; i++) {
      var attr = changes[i];
      unset ? delete current[attr] : current[attr] = attrs[attr];
    }
  }

  // You might be wondering why there's a `while` loop here. Changes can
  // be recursively nested within `"change"` events.
  if (changing) return this;
  if (!silent) {
    while (this._pending) {
      options = this._pending;
      this._pending = false;
      this.trigger('change', this, options);
    }
  }
  this._pending = false;
  this._changing = false;
  return this;
},
```

Simple lock suggestion:
```
lock: function(attributeName) {
  if (this._lockedAttributes[attributeName] === undefined) this._lockedAttributes[attributeName] = 0;
  this._lockedAttributes[attributeName]++;
}

unlock: function(attributeName) {
  if (this._lockedAttributes[attributeName] === undefined || this._lockedAttributes[attributeName] === 0) return;
  this._lockedAttributes[attributeName]--; 
}

validate: function (attrs, options) {  
  for (var attr in attrs) {
    if (this._lockedAttribute[attr] > 0) return false;
  }
}
```

As implemented, the simple locking method seems a lot more sensible to me.

Either way the Backbone.Model should not be replaced by a Backbone.Model.extend({ anything }) as in the current code. This adds another level to the prototype chain further increasing the resource requirements. Applying the function overrides directly to the Backbone.Model.prototype object is a better method of implementing overriding behaviour.

If the original code is to be retained, I would suggest removing, _.has, _.filter, _.extend, _.each, _.isEmpty and _.size.

oliverfoster:
The plugin architecture does not rely on this code as removing it doesn't break anything and there are no direct dependencies upon it. No plugins we know of [so far] require or utilise the removed code either, hence the suggestion to remove it. Please comment here if you directly depend on this code for extensions if so, please supply use cases or code references so that we can either make this code more efficient or remove it entirely. It causes a huge performance bottleneck in adapt.

oliverfoster:
As a side note, none of the components, extensions, or core from Adapt Learning utilise this code.

oliverfoster:
TODO: performance enhancements to this file. Delayed discussion on need.

oliverfoster:
need to redo this, apologies.

PR #519, opened 2015-01-22
Library updates

oliverfoster:
tested with ie8 and chrome, is much faster.

backbone not updated as options behaviour has changed in instanciation of models and views.

made templates into 'require' module to ensure load sequence and bring handlebars inline with latest version.


oliverfoster:
#507

brian-learningpool:
+1

dancgray:
+1

taylortom:
+1

Issue #506, opened 2015-01-16
_onQuizFailure is incorrect in config.json, should be _onAssessmentFailure


dancgray:
Master branch of adapt-contrib-spoor uses _onAssessmentFailure instead of _onQuizFailure

LN 18 or config.json should be:

            "_onAssessmentFailure": "incomplete"

See https://github.com/adaptlearning/adapt_framework/pull/468

zenduo:
This one looks like its been added to the develop branch now.

PR #523, opened 2015-01-26
hot fix for issue #506

dancgray:
_onQuizFailure is now _onAssessmentFailure in spoor.

See https://github.com/adaptlearning/adapt-contrib-spoor/blob/v1.1.6/js/adapt-contrib-spoor.js#L105

and https://github.com/adaptlearning/adapt-contrib-spoor/blob/v1.1.6/example.json#L12


moloko:
+1

taylortom:
+1

oliverfoster:
+1

Issue #443, opened 2014-07-15
Accessibility tab index reset after drawer opened on IE8


develop-stuartohare:
popupManager.js on 'popup:closed' line24 - also needs to check if '_accessibilty' flag is set in the config model (Adapt.config).

e.g - if ($activeElement && Adapt.config.get('_accessibility')._isEnabled).....

oliverfoster:
Fixed with accessibility merge. Please close

Issue #476, opened 2014-11-20
(Accessibility) Popup manager overwriting tabindex on popup close


oliverfoster:
_From @oliverfoster on October 17, 2014 10:22_

_Copied from original issue: cgkineo/adapt_framework#8_

PR #527, opened 2015-01-28
Accessibility integration

oliverfoster:
Traps tab on accessibility button if accessibility isn't on
Adds accessibility instructions for screen readers
Fixes popupmanager and tabindexes on popup/popdown
Wraps tab around to top of page from the bottom
Integrates jquery.a11y into the handlebars templates so that text blocks can become tabbable
Adds accessibility state rendering to componentView (is currently missing the 'state' partial as it needs to be added to theme)
Allows aria-labels to be read across all devices
Expands LESS focus outline to include all focused elements
Maintains and improves .focused class for ie8
Adds _globals to all popup models
Fixes button states, question view, notify and drawer to allow all to become accessible

Next stage is templates, components and extensions.


oliverfoster:
Directly fixes: #443 #476

oliverfoster:
Tested working in chrome41, ie8 and safari on ios 7.1.2

moloko:
+1. quick check done in IE8 & 9, FF 35, Chrome and - just for fun - Opera

oliverfoster:
err, going to have to redo this

oliverfoster:
apologies

PR #531, opened 2015-01-30
if_value_equals doesn't pass context back

### LINKED ISSUE #530: if_value_equals doesn't pass context back
if _value_equals helper does not pass context back in block.inverse().

will8137:
if_value_equals helper does not pass context back when returning
block.inverse.

fixes #530

### DIFF
--- a/src/core/js/helpers.js
+++ b/src/core/js/helpers.js
@@ -23,7 +23,7 @@ define(function(require){
                 if (value === text) {
                     return block.fn(this);
                 } else {
-                    return block.inverse();
+                    return block.inverse(this);
                 }
             }
         };


hrajotia:
+1

darylhedley:
+1 - good spot!

oliverfoster:
+1

belal-almassri-spongeuk:
Hi, 

This solution I added to one of the projects I worked on, it worked for me. If you add the following code to the app.js page inside the funtion function loadCourseData()

        Handlebars.registerHelper('compare', function(lvalue, rvalue, options) {
        if (arguments.length < 3)
            throw new Error("Handlerbars Helper 'compare' needs 2 parameters");
        operator = options.hash.operator || "==";
        var operators = {
            '==':       function(l,r) { return l == r; },
            '===':      function(l,r) { return l === r; },
            '!=':       function(l,r) { return l != r; },
            '<':        function(l,r) { return l < r; },
            '>':        function(l,r) { return l > r; },
            '<=':       function(l,r) { return l <= r; },
            '>=':       function(l,r) { return l >= r; },
            'typeof':   function(l,r) { return typeof l == r; }
        }
        if (!operators[operator])
            throw new Error("Handlerbars Helper 'compare' doesn't know the operator "+operator);
        var result = operators[operator](lvalue,rvalue);
        if( result ) {
            return options.fn(this);
        } else {
            return options.inverse(this);
        }
    });



Finally add you compare statement in the .hbs page as the following example:
    {{#compare debug "true"operator="=="}}
       
    {{^}}
      
    {{/compare}}

Thanks,
Belal

Issue #538, opened 2015-02-18
add to grunt the fonts folder of components/extensions/menus


chucklorenz:
Recently extensions and menu assets were added to the Grunt file to fix #304. Can the same be done with the fonts folder for components, extensions, and menus? I'm working on an extension (a Drawer item) that uses an icomoon font file to supplement the icons in vanilla.woff. Seems reasonable that other components and menus might need to deliver fonts, too.

hrajotia:
@chucklorenz  thanks for the PR.

PR #540, opened 2015-02-19
Add fonts folders to grunt copy and watch

chucklorenz:
Addresses #538: fonts folders of other plugins (not just theme) need to be copied during build.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,30 +1,7 @@
 module.exports = function(grunt) {
     require('matchdep').filterDev('grunt-*').forEach(grunt.loadNpmTasks);
-    var outputdir = grunt.option('outputdir') || '',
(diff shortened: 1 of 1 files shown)


taylortom:
+1

I think we need to have a look at how we specify the files arr; currently a bit inconsistent. In some places we do:
```
courseAssets: {
   files: [{
      expand: true, 
      src: ['**/*','!**/*.json'], 
      dest: 'build/course/', 
      cwd: 'src/course/'
   }]
},
```
And in others:
``` javascript
menuAssets: {
   files: grunt.file.expand(['src/menu/*/']).map(function(cwd) {
      return {
         expand: true,
         src: ['**'],
         dest: 'build/assets/',
         cwd: cwd + 'assets/',
         filter: "isFile"
      };
   })
},
```

hrajotia:
+1

oliverfoster:
+1

PR #545, opened 2015-03-05
"Required to complete" feature

LukaszGrela:
This adds ability to specify which components should be completed to mark entire block (and other components) as completed. Block model needs to specify `_requireCompletionOf` property which can have following values: 'all', 'any' or valid component id that is child of this block. Below are few examples of use. When `_requireCompletionOf` is missing or empty it defaults to `all`.

    {
        "_id": "b-03",
        "_parentId": "a-02",
        "_type": "block",
        "_requireCompletionOf": "any"
    },
    {
        "_id": "b-03",
        "_parentId": "a-02",
        "_type": "block",
        "_requireCompletionOf": ["all"]
    },
    {
        "_id": "b-03",
        "_parentId": "a-02",
        "_type": "block",
        "_requireCompletionOf": "c-05"
    },
    {
        "_id": "b-03",
        "_parentId": "a-02",
        "_type": "block",
        "_requireCompletionOf": ["c-05","c-10"]
    },

### DIFF
--- a/src/core/js/models/blockModel.js
+++ b/src/core/js/models/blockModel.js
@@ -6,12 +6,145 @@
 
 define(function(require) {
 
-	var AdaptModel = require('coreModels/adaptModel');
+    var Backbone = require('backbone'),
+        Adapt = require('coreJS/adapt'),
+        AdaptModel = require('coreModels/adaptModel'),
+        BlockModel = AdaptModel.extend({
+            _parent:'articles',
+            _siblings:'blocks',
+            _children: 'components'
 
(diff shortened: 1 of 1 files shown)


taylortom:
I really like this idea. I'm wondering if this could be an extension though, rather than being added to core. Also, the PR should be for develop, not master.

LukaszGrela:
I'll move it to develop - but dont know how to do extension :)

LukaszGrela:
Done, https://github.com/adaptlearning/adapt_framework/pull/549

Issue #490, opened 2014-12-16
implement _isOptional


moloko:
Currently this property exists in adaptModel.js, but there is no code that uses it.

My understanding of this is that it's purpose is to allow an author to mark a ContentObject/Article/Block/Component as 'optional' i.e. the user is not required to complete it and, as such, it should be ignored by anything that calculates overall progress/completion (the progress of the item should still be stored, but just for reference).

When speak to @darylhedley about this yesterday he thought it should be the responsibility of a plugin to use _isOptional - but given that the overall course completion is currently calculated by AdaptModel.checkCompletionStatus - which generates the 'change:_isComplete' event that a plugin like contrib-spoor listens for to find out when the overall course is completed - I would have thought it would be best implemented in checkCompletionStatus.

I guess you could do it within contrib-spoor, but you'd essentially be replicating the functionality of AdaptModel.checkCompletionStatus - which seems like wasted effort to me.

Any thoughts on this?

taylortom:
I'm of the opinion that if it's in the core code, it should be implemented. This goes with what @darylhedley was saying about keeping the core as light as possible.

Plugins could just as easily set _isOptional for themselves without it needing to be defined in AdaptModel, this is just confusing for devs that aren't well-versed in Adapt; I know I've been caught trying to use _isOptional assuming it was set somewhere, only to find it wasn't.

darylhedley:
``_isOptional`` is a locked attribute:
https://github.com/adaptlearning/adapt_framework/wiki/Adapt-API#--set--locked-attributes
Removing this from the adaptModel.js will mean it becomes a standard attribute on a model and not use the built in mediator pattern.

taylortom:
Is it reasonable to be able to define new locked attributes from a plugin? I imagine this is something that could be useful in the future. 

I guess if there's no other sensible way to deal with it, we'll just need to make sure it's documented clearly in the code.

olivermartinfoster:
Firstly, do we currently have any community plugins that provide the _isOptional behaviour or any community plugins which use the _isOptional attribute? Is the _is optional attribute behaviour a large enough community requirement that it should be core? Will it break anything else? Have we put the idea to tender on the forum? Do we have a proposed implementation that can be reviewed?

darylhedley:
@taylortom - yes any model extending from Backbone.Model will have the ability to lock the attribute
@olivermartinfoster - this should stay in core as many plugins would need access to this

oliverfoster:
[...]

PR #549, opened 2015-03-12
"Required to complete" feature

LukaszGrela:
This adds ability to specify which components should be completed to mark entire block (and other components) as completed. Block model needs to specify `_requireCompletionOf` property which can have following values: `'all'`, `'any'` or valid component id that is child of this block e.g. `c-02`. Below are few examples of use. When `_requireCompletionOf` is missing or empty it defaults to `all`.

    {
        "_id": "b-03",
        "_parentId": "a-02",
        "_type": "block",
        "_requireCompletionOf": "any"
    },
    {
        "_id": "b-03",
        "_parentId": "a-02",
        "_type": "block",
        "_requireCompletionOf": ["all"]
    },
    {
        "_id": "b-03",
        "_parentId": "a-02",
        "_type": "block",
        "_requireCompletionOf": "c-05"
    },
    {
        "_id": "b-03",
        "_parentId": "a-02",
        "_type": "block",
        "_requireCompletionOf": ["c-05","c-10"]
    },

### DIFF
--- a/src/core/js/models/blockModel.js
+++ b/src/core/js/models/blockModel.js
@@ -6,13 +6,147 @@
 
 define(function(require) {
 
-	var AdaptModel = require('coreModels/adaptModel');
+    var Backbone = require('backbone'),
+        Adapt = require('coreJS/adapt'),
+        AdaptModel = require('coreModels/adaptModel'),
+        BlockModel = AdaptModel.extend({
+            _parent:'articles',
+            _siblings:'blocks',
+            _children: 'components'
 
-    var BlockModel = AdaptModel.extend({
-        _parent:'articles',
-    	_siblings:'blocks',
-        _children: 'components'
+            checkCompletionStatus: function (model, isComplete) {
+                if (this.ignoreCompletionEvent) { return; }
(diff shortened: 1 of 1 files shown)


brian-learningpool:
Lukasz, there has been some discussion around the area of what counts towards completion in issue #490.  

I think that a better approach might be for components to say whether or not they require completion to be counted as part of the course completion.

LukaszGrela:
Maybe, but this way I found it less invasive, it will work with previous and future sources.

oliverfoster:
Note: Current isOptional proposals do not provide Lukasz 'any' functionality which could be quite useful.

taylortom:
Can see the use for the completion of *any* - will park this for when we tackle the `_isOptional` implementation

moloko:
@LukaszGrela - as we mentioned we have been working on implementing the _isOptional feature properly (which is now done). This clearly duplicates some of what you've done here - though not the 'any' feature. However, we really like this idea and so have logged [ticket ABU-673](https://adaptlearning.atlassian.net/browse/ABU-673) in Jira for it as a likely future development item

PR #555, opened 2015-03-18
Fix for leftover ie8 opacity filter

oliverfoster:
With a filter=opactiy(1) on the page, ie8 has to filter each video frame and so causes a 100% processor problem.


moloko:
+1

zenduo:
+1

dancgray:
+1

Issue #541, opened 2015-02-23
Change navigation view to use _globals._accessibility instead of _accessibility


oliverfoster:
Uniformity issue.

https://github.com/adaptlearning/adapt-contrib-vanilla/pull/65

https://github.com/adaptlearning/adapt-contrib-vanilla/issues/66

zenduo:
Moved to https://adaptlearning.atlassian.net/browse/ABU-487

PR #562, opened 2015-03-24
_globals.accessibility fix

zenduo:
To resolve: https://github.com/adaptlearning/adapt_framework/issues/541


oliverfoster:
Cool, can you revert notify and notifypush in this one? They work in a slightly different way to drawer and navigation.

PR #566, opened 2015-04-08
Resolves #565 - Implements 'navigation:homeButton'

### LINKED ISSUE #565: There is no 'navigation:homeButton' event
There is a handler for 'navigation:backButton' which redirects the user to the previous route but there is no such handler for 'navigation:homeButton'.

brian-learningpool:
It makes sense to implement a handler for this, since 'backButton' is
already supported.

### DIFF
--- a/src/core/js/router.js
+++ b/src/core/js/router.js
@@ -103,7 +103,9 @@ define(function(require) {
         },
 
         navigateToHomeRoute: function() {
-            this.navigate('#', {trigger: true});
+            if (Adapt.router.get('_canNavigate')) {
+                this.navigate('#', {trigger: true});                
+            }
         },
 
         navigateToParent: function() {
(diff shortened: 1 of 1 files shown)


src/core/js/router.js
```diff
@@ -101,6 +102,10 @@ define(function(require) {
             }
         },
 
+        navigateToHomeRoute: function() {
+            this.navigate('#', {trigger: true});
```
oliverfoster: Can we have if (Adapt.router.get('_canNavigate')) {  in there as well?

We've an extension called 'promptOnPageIncomplete'. This'd be quite usefull.

brian-learningpool: Good point.  I've updated the code.

oliverfoster:
+1

darylhedley:
+1

hrajotia:
+1

taylortom:
@brian-learningpool can we get this put into develop too?

PR #576, opened 2015-04-16
allowed navigation to any object via the location.hash

chris-steele:
e.g:

index.html#/id/co-05
index.html#/id/c-85
index.html#/id/b-30
index.html#/id/a-10

etc

Also as requested by Ollie the offset of the navigation is taken into account in Adapt.navigateToElement.

### DIFF
--- a/src/core/js/adapt.js
+++ b/src/core/js/adapt.js
@@ -35,7 +35,7 @@ define(function(require){
         }
 
         // Trigger scrollTo plugin
-        $.scrollTo(selector, settings, {top:settings.offset, left:0});
+        $.scrollTo(selector, _.extend(settings, { top: settings.offset, left: 0 }));
         // Trigger an event after animation
         // 300 milliseconds added to make sure queue has finished
         _.delay(function() {
--- a/src/core/js/adapt.js
+++ b/src/core/js/adapt.js
@@ -34,10 +34,11 @@ define(function(require){
             settings.duration = $.scrollTo.defaults.duration;
         }
         
-        var navigationOffset = -($('.navigation').height()+10);
-        if (!settings.offset) settings.offset = { top: navigationOffset, left: 0};
-        if (!settings.offset.top) settings.offset.top = navigationOffset;
-        if (!settings.offset.left) settings.offset.left = 0;
+        var navigationHeight = $(".navigation").outerHeight();
+
+        if (!settings.offset) settings.offset = { top: -navigationHeight, left: 0 };
+        if (settings.offset.top === undefined) settings.offset.top = -navigationHeight;
+        if (settings.offset.left === undefined) settings.offset.left = 0;
 
         // Trigger scrollTo plugin
         $.scrollTo(selector, settings);
--- a/src/core/js/adapt.js
+++ b/src/core/js/adapt.js
@@ -35,7 +35,7 @@ define(function(require){
         }
 
         // Trigger scrollTo plugin
-        $.scrollTo(selector, _.extend(settings, { top: settings.offset, left: 0 }));
+        $.scrollTo(selector, settings);
         // Trigger an event after animation
         // 300 milliseconds added to make sure queue has finished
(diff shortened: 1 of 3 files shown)


src/core/js/adapt.js
```diff
@@ -43,33 +46,34 @@ define(function(require){
...
 
+        if(offset===undefined) offset = -($('.navigation').height()+10);
+
         // Setup settings object
         var settings = (settings || {});
 
         // Removes . symbol from the selector to find the model
         var currentModelId = selector.replace(/\./g, '');
-        var currentModel = Adapt.findById(currentModelId);
+        var currentModel = Adapt[Adapt.mapById(currentModelId)].findWhere({_id: currentModelId});
```
oliverfoster: findById does Adapt[Adapt.mapById(currentModelId)].findWhere({_id: currentModelId}); so you can revert this line.

src/core/js/router.js
```diff
@@ -55,23 +55,30 @@ define(function(require) {
         
         handleId: function(id) {
             
-            this.removeViews();
-            this.showLoading();
+            var currentModel = Adapt[Adapt.mapById(id)].findWhere({_id: id});
```
oliverfoster: You can use findById here also

src/core/js/adapt.js
```diff
@@ -43,33 +46,34 @@ define(function(require){
         
     }
 
-    Adapt.navigateToElement = function(selector, settings) {
+    Adapt.navigateToElement = function(selector, settings, offset) {
         // Allows a selector to be passed in and Adapt will navigate to this element
 
+        if(offset===undefined) offset = -($('.navigation').height()+10);
```
oliverfoster: can we use settings.offset instead of a new parameter?

src/core/js/router.js
```diff
@@ -55,23 +55,30 @@ define(function(require) {
...
+
+            if (currentModel.get('_type') == 'page' || currentModel.get('_type') == 'menu')
```
oliverfoster: How about a switch?

switch (currentModel.get('_type)) {
case 'page': case 'menu': 
      /* menu + page loading here */
      break;
default:
      Adapt.navigateToElement('.' + id);
}

chris-steele:
Great thanks Ollie I'll make those changes now.

chris-steele:
Changes made - are these OK?

src/core/js/adapt.js
```diff
@@ -49,11 +50,13 @@ define(function(require){
         // Setup settings object
         var settings = (settings || {});
 
+        if(settings.offset===undefined) settings.offset = -($('.navigation').height()+10);
```
oliverfoster: if (settings.top === undefined) settings.top = -($('.navigation').height()+10);

src/core/js/adapt.js
```diff
@@ -33,8 +33,9 @@ define(function(require){
         if (!settings.duration) {
             settings.duration = $.scrollTo.defaults.duration;
         }
+
         // Trigger scrollTo plugin
-        $.scrollTo(selector, settings);
+        $.scrollTo(selector, settings, {top:settings.offset, left:0});
```
oliverfoster: $.scrollTo(selector, settings);

chris-steele:
I was just looking at this line and wondering what I was thinking. Thanks!

oliverfoster:
try passing settings.top instead... it might be a better idea?
the you can just use the old $.scrollTo(selector, settings);

oliverfoster:
+1

moloko:
+1

chris-steele:
Looks good Ollie, thanks for being a bit more thorough than I(!)

+1

hrajotia:
+1

oliverfoster:
+1

moloko:
+1
cheers everyone I think this will prove very useful to have. would it make sense to have things like bookmarking hook into this as well maybe?

PR #577, opened 2015-04-17
Accessibility fixes

oliverfoster:
Stop redirecting the tab key to the accessibility button if the user has scrolled or entered a native form element. 

@brian-learningpool This is a fix for stopping the tab key augmentation after a normal interaction.


brian-learningpool:
+1

moloko:
+1

dancgray:
+1

PR #578, opened 2015-04-17
Fixes #573 - Router struggles with nesting

### LINKED ISSUE #573: Bug in router navigateToParent() function when using nested pages
There seems to be an issue routing back when the page is nested inside one or more menus.  This can be replicated on a simple course structure as follows:
- Page 1
- Menu 1
    - Page 2

Using the Vanilla theme, hit the < (back) button when on Page 2.  The screen gets stuck as Loading.... due to what I think is a problem with this line:
https://github.com/adaptlearning/adapt_framework/blob/293eb9fdba452053f1bf67e5737511d5e7c05d92/src/core/js/router.js#L117

brian-learningpool:
- Slightly refactored router.js
- Setting model._isReady to false on AdaptView initialize()

### DIFF
--- a/src/core/js/router.js
+++ b/src/core/js/router.js
@@ -103,12 +103,9 @@ define(function(require) {
         },
 
         navigateToParent: function() {
-            var currentModel = Adapt.contentObjects.findWhere({_id:Adapt.location._currentId});
-            var parent = currentModel.getParent();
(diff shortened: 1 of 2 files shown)


oliverfoster:
+1

dancgray:
+1

taylortom:
+1

PR #580, opened 2015-04-17
BUGFIX: Load jquery after require to allow for other libraries to require jquery

oliverfoster:
@hrajotia it didn't work sometimes on a fast connection as it was before.

jquery would load before require and would not be available to scrollTo, velocity, backbone etc. 

it would works sometimes as jquery is a larger file than require and so require would load and execute before jquery, allowing jquery to 'define' itself and therefore allow jquery to be available to the other libraries.

i've rewritten so that it goes, json2, consoles, require (wait for load), jquery (wait for load), adapt

### DIFF
--- a/src/core/js/scriptLoader.js
+++ b/src/core/js/scriptLoader.js
@@ -13,17 +13,7 @@
         return false;
     })();
 
-    //2. Load requirejs
-    function loadRequireJS() {
-        Modernizr.load([
-            {
-                load: "libraries/require.js",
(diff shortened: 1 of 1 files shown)


hrajotia:
+1

brian-learningpool:
+1, does this affect master too?

oliverfoster:
@brian-learningpool  no, only to develop, this is part of using jquery.v2 for everything above ie8

moloko:
+1

PR #593, opened 2015-05-08
Adds a check for _children being undefined

### LINKED ISSUE #592: Issue with findDecendants()/getChildren() on component model
There's a slight issue around `findDecendants()` and `getChildren()` in that the `_children` property does not exist in components, but sometimes a call to `findDecendants()` can end up recursively calling `getChildren()`.  

We require a check for undefined here to avoid the "Cannot read property 'where' of undefined error".

brian-learningpool:
This resolves a potential issue with findDecendants() and getChildren() as
this property does not exist on components.  

Resolves #592.

### DIFF
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -122,7 +122,7 @@ define(function (require) {
 
                 returnedDescedants = new Backbone.Collection(flattenedDescendants);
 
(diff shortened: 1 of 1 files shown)


moloko:
+1

oliverfoster:
+1

PR #594, opened 2015-05-18
Unknown platform fix

oliverfoster:
Allows the devices platform to be identified as 'PlatformUnknown' when 'Windows', 'Mac' and 'Linux' aren't specified.

In the case of Windows Mobile 8.1, "ARM" is the platform name (which is rather non-specific, all things considered) https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/platform

### DIFF
--- a/src/core/js/device.js
+++ b/src/core/js/device.js
@@ -64,13 +64,14 @@ define(function(require) {
 
         if (platform.indexOf("Win") != -1) {
             return "Windows";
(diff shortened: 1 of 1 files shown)


brian-learningpool:
+1

zenduo:
+1

tomgreenfield:
+1

PR #598, opened 2015-05-21
Implemented _isOptional behaviour

oliverfoster:
PLP needs updating to display optional states.

We need some discussion around this.

https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/pull/42

### DIFF
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -39,6 +39,9 @@ define(function (require) {
                 this._parent = 'contentObjects';
             }
             if (this._children) {
+                //if parent is optional, apply to children
+                if (this.get('_isOptional')) this.setOptional(true);
+    
                 Adapt[this._children].on({
                     "change:_isReady": this.checkReadyStatus,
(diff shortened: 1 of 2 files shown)


moloko:
+1

zenduo:
+1

oliverfoster:
https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/pull/42 < jamie, any ideas for how plp should represent optional elements?

danielghost:
I would maybe get a designers view on this but I would suggest an optional component appears as normal, displaying its own progress, but is excluded from the overall progress for the extension. In cases where an optional component should be excluded from the view completely this could be done by disabling it in the JSON.

moloko:
I think generally one would not include optional components in the PLP

However, I'm sure there will be occasions when it is necessary to include them so perhaps we implement some method of styling them differently. I guess it would therefore make sense to add an 'optional' class to such items in the PLP to allow for that?

dancgray:
+1

Issue #544, opened 2015-03-04
Grunt: make file listing consistent


taylortom:
I think we need to have a look at how we specify the files arr; currently a bit inconsistent. In some places we do:
```
courseAssets: {
   files: [{
      expand: true, 
      src: ['**/*','!**/*.json'], 
      dest: 'build/course/', 
      cwd: 'src/course/'
   }]
},
```
And in others:
``` javascript
menuAssets: {
   files: grunt.file.expand(['src/menu/*/']).map(function(cwd) {
      return {
         expand: true,
         src: ['**'],
         dest: 'build/assets/',
         cwd: cwd + 'assets/',
         filter: "isFile"
      };
   })
},
```

taylortom:
Fixed in #613

PR #613, opened 2015-06-03
Grunt update

taylortom:
General tidying up of Gruntfile.js. Changes include:
- Fixed `watch` task
- Made sure src files are referenced consistently (fixes https://github.com/adaptlearning/adapt_framework/issues/544)
- Added getters for the dynamic variables for simplicity
- Removed `build/` after every reference to `<%= outputdir %>`, and instead moved this to getOutputDir (suggest altering slightly this so that the authoring tool supplies the `/build` prefix, and is not forcibly added to the end of the path)
- Src path is now dynamic to match theme
- Replaced all double quotes with single quotes for consistency
- Moved all package deps to devDeps
- Added some logs to various tasks (`build`, `dev`, `server`, `server-scorm`)


oliverfoster:
+1 works fine, have been using it since posted, tested build, dev and server

moloko:
+1

oliverfoster:
i think leaving tracking-insert as the build/dev task is the least damaging. (as far as i understand it)  courses can be patched after they have been distributed, so the ids have to remain consistent

oliverfoster:
+1

moloko:
+1

hrajotia:
+1

PR #614, opened 2015-06-03
Make sure focus goes to component state if both buttons are disabled

oliverfoster:
This is a problem when you get questions right in the assessment where there is no feedback button. The submit button becomes disabled and the focus cannot move to the next element in its parent.

Have forced focus to go to component-state element if available.


moloko:
+1

dancgray:
+1

taylortom:
+1

PR #616, opened 2015-06-05
ABU-640: Notify and Drawer

oliverfoster:
Scrollbars hidden for both drawer and notify to stop background content from scrolling whilst they are open.

Added: Wait for image loading in notify, fixed notify positioning.

Fixed animations for ie9+
No animations for ie8

NOTE: Must fade the page to white in order to hide the scrollbars and prevent the user from seeing a page jump, so have minimized the effect using fadeins.

TEST ON:
iphone4 ios7
ipad mini ios8
android chrome
winxp ie8
win8 ie11 + win8 ie11:ie9 mode
win8 ff
win8 chrome
blackberry

### DIFF
--- a/src/core/js/notify.js
+++ b/src/core/js/notify.js
@@ -87,12 +87,6 @@ define(function(require) {
 		var notify = new NotifyView({
 			model: new NotifyModel(notifyObject)
 		});
-
-		var element = notify.$(".notify-popup");
-
-		/*ALLOWS POPUP MANAGER TO CONTROL FOCUS*/
-		Adapt.trigger('popup:opened', element);
-		element.a11y_focus();
 		
 	};
 
--- a/src/core/js/views/drawerView.js
+++ b/src/core/js/views/drawerView.js
@@ -97,7 +97,6 @@ define(function(require) {
             //only trigger popup:opened if drawer is visible, pass popup manager drawer element
             if (!this._isVisible) {
                 Adapt.trigger('popup:opened', this.$el);
-
                 this._isVisible = true;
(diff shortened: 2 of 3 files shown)


moloko:
+1

dancgray:
+1

oliverfoster:
brought up to date with latest merged pr's

belal-almassri-spongeuk:
Perfect we need this feature, the only issue with it, is the page height is calculated when we open the popup only, and it is not updated when we resize the page, so if you add trickle to the article, trigger notify, and while the notify popup is open resize the page, now you can scroll!. close notify and you will notice either the some content at the end of the block is hidden or a white space between the content and the trickle button. I think we need to update the following line on page resize. 
_bodyHeight = $("body").css("height");

oliverfoster:
1. fixed the resize whilst popup open
2. added pr for plp to wait for drawer to close https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/pull/48

taylortom:
[PLP PR has now been merged]

taylortom:
Not a fan of the white flash. Is the alternative much worse?

oliverfoster:
the alternative is to keep the scrollbars as far as i can tell. unless we can find an implementation that doesn't cause the page to jump as the scrollbars are removed.

taylortom:
Kewl, I'll maybe create a separate low-priority issue

taylortom:
+1

moloko:
+1

dancgray:
+1

PR #617, opened 2015-06-05
Fix for screenreaders informing the user of page + menu loads

oliverfoster:
Changed focus order + timing and made sure model references were correct.


oliverfoster:
Not quite

PR #622, opened 2015-06-08
Stop all body text being selectable

oliverfoster:
fix to prevent accidental selection of text on touch devices.

### DIFF
--- a/src/core/less/reset.less
+++ b/src/core/less/reset.less
@@ -54,3 +54,11 @@ a, button {-webkit-tap-highlight-color: rgba(0,0,0,0.2);}
 	// Fix for iOS devices not displaying icons if hidden then shown
 	-webkit-transform: translateZ(0);
 }
+body {
+    -webkit-touch-callout: none;
+    -webkit-user-select: none;
+    -khtml-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
+}


moloko:
+1
which reminds me, should we not include some kind of no-prefix add-on for LESS in the framework?

oliverfoster:
Wise plan. Got on in mind?

moloko:
I know little of such things, might be best to check with a front end expert. http://lessprefixer.com/ looks like it might be OK though?

taylortom:
Excellent suggestion

taylortom:
+1

hrajotia:
+1

tomgreenfield:
I think there's a valid use case for being able to highlight and copy text out of Adapt courses. A learner may wish to paste content into another piece of software to take notes, copy terms into a Google search, or paste a hyperlink into a different browser, to name a few examples. It's also useful for reviewers who request amends to the course to be able to grab chunks of text from the page.

At the very least, could we consider re-enabling on desktop browsers? I'd be happy to hear your thoughts.

oliverfoster:
+1 @tomgreenfield

moloko:
ah right, I assumed this was for touch devices only - agree with @tomgreenfield

taylortom:
Thinking this over again, and in support of @tomgreenfield, I think that this can be valuable even for mobiles. Can we have an option to disable this, rather than turning it off by default? Can't speak for iPhone, but as an android user, I've come to expect *long-click-to-select-text* pretty much everywhere (web and apps), so not being able to do this seems odd.

dancgray:
I also agree with @tomgreenfield, if we are going to to turn it off can we have the option to disable.

oliverfoster:
How about this one > https://github.com/adaptlearning/adapt_framework/pull/634

PR #627, opened 2015-06-12
Fix issue with IOS (Add to home screen)

belal-almassri-spongeuk:
We noticed an issue on iPad and iPhone, open a course using Safari click add to home screen option, and then click the website icon that is created on the home screen, it will launch the website using Safari, but the course will not open!

It worked once I made this fix , we have the following line of code instead `<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">`.


moloko:
+1
Note that even with this fix this will likely only work for courses that require no login to be able to access...

brian-learningpool:
+1

taylortom:
+1

PR #634, opened 2015-06-17
ABU-695: Disable text selection configurable

oliverfoster:
1. Moved body {} styling to a .no-select {} class.
2. Added 'size-small',  'size-medium' and 'size-large' class for respective device sizes to html tag.
3. Added '_disableTextSelectOnClasses' to config.json in _accessibility: {}
4. Extended accessibility code to compare the classes in the above attribute with the classes on the html tag when the device size changes. The class 'no-select' is added / removed to the html tag when a match is found.
5. By default no text selection is permitted, do we want 'no-select' on just touch devices?

### DIFF
--- a/src/course/config.json
+++ b/src/course/config.json
@@ -5,7 +5,8 @@
     "_accessibility": {
         "_isEnabled": false,
         "_shouldSupportLegacyBrowsers": true,
-        "_isTextProcessorEnabled": true
+        "_isTextProcessorEnabled": true,
+        "_disableTextSelectOnClasses": "size-small size-medium size-large"
     },
     "_drawer": {
         "_showEasing": "easeOutQuart",
--- a/src/course/config.json
+++ b/src/course/config.json
@@ -6,7 +6,7 @@
         "_isEnabled": false,
         "_shouldSupportLegacyBrowsers": true,
         "_isTextProcessorEnabled": true,
-        "_disableTextSelectOnClasses": "size-small size-medium size-large"
+        "_disableTextSelectOnClasses": ".size-small .size-medium .size-large"
     },
     "_drawer": {
         "_showEasing": "easeOutQuart",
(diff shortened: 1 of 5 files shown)


src/course/config.json
```diff
@@ -5,7 +5,8 @@
     "_accessibility": {
         "_isEnabled": false,
         "_shouldSupportLegacyBrowsers": true,
-        "_isTextProcessorEnabled": true
+        "_isTextProcessorEnabled": true,
+        "_disableTextSelectOnClasses": "size-small size-medium size-large"
```
oliverfoster: This bit would become:

src/course/config.json
```diff
@@ -5,7 +5,8 @@
     "_accessibility": {
         "_isEnabled": false,
         "_shouldSupportLegacyBrowsers": true,
-        "_isTextProcessorEnabled": true
+        "_isTextProcessorEnabled": true,
+        "_disableTextSelectOnClasses": ".size-small .size-medium .size-large"
```
oliverfoster: This bit would become:

"_disableTextSelectOnClasses": ".touch.size-small"

for mobile touch devices

dancgray:
+1, needs some explanation in wiki or somewhere. Also how would this be selected in the authoring tool.

oliverfoster:
The authoring tool is a good question.

taylortom:
+1

brian-learningpool:
Regarding the question about the class combinations, won't this just be a feature of the selected theme?  In the authoring tool now it is also possible to input custom LESS or CSS markup on the Project Settings screen.

brian-learningpool:
+1

taylortom:
Going to merge this, and assume @dancgray doesn't object to the if-else amend.

dancgray:
+1

taylortom:
(Thanks :))

PR #647, opened 2015-07-03
Allow legacy courses to work with version 2.0

### LINKED ISSUE #642: Support content created in version 1.x of the Adapt Framework
I'm raising a ticket to keep track of this as I've had a look at the version 2.0 code and think it should be possible to support the JSON formats created in previous versions via some conditionals and some defensive programming.

This would really ease adoption of the new version among people who either have old content they want to maintain, or are in the middle of creating new content

brian-learningpool:
Fixes #642:
- Allows courses created in version 1.x to work with version 2.0 of the framework
- Allows accessibility to be toggled
- Adds an `_isActive` flag for the `_accessibility` attribute, for when accessibility can be toggled.


brian-learningpool:
@taylortom / @dancgray  - can you have a look a this please?  This has to go in before we start work on the 0.1.2 version of the authoring tool

dancgray:
Tabbing is kind of broken by default with this.

Should_accessibility._isEnabled be set to true by default in the config for v2.0.?

src/core/js/views/buttonsView.js
```diff
@@ -73,21 +73,23 @@ define(function() {
...
-                    .attr('aria-label', this.model.get('_buttons')["_" + changedAttribute].ariaLabel);
-                    break;
-                default:
-					//enabled button, make accessible and update aria labels and text.
-                    this.$('.buttons-action').a11y_cntrl_enabled(true).html(this.model.get('_buttons')["_" + changedAttribute].buttonText)
-                    .attr('aria-label', this.model.get('_buttons')["_" + changedAttribute].ariaLabel);
+            } else {        
+                // Backwords compatibility with v1.x
+                var ariaLabel = this.model.get('_buttons')["_" + changedAttribute].ariaLabel;
+                var buttonText = this.model.get('_buttons')["_" + changedAttribute].buttonText;
```
taylortom: These two are throwing errors when using a v1.x course due to a differing _buttons:
_buttons

oliverfoster: This stuff really breaks the buttons, i'm investigating

src/core/js/views/questionView.js
```diff
@@ -127,10 +127,34 @@ define(function(require) {
...
+                    
+                    // HACK - Append other missing values
+                    buttons['_showFeedback'] = {
+                        buttonText: 'Show feedback',
+                        ariaLabel: 'Show feedback'
+                    };
+                    
+                    this.model.set('_buttons', buttons);    
+                }              
             }
```
oliverfoster: this stuff is probably breaking the buttons....

src/core/js/accessibility.js
```diff
@@ -66,75 +101,140 @@ define(function(require) {
...
+                    isFocusControlEnabled: true,
+                    isFocusLimited: false,
+                    isRemoveNotAccessiblesEnabled: true,
+                    isAriaLabelFixEnabled: true,
+                    isFocusWrapEnabled: true,
+                    isScrollDisableEnabled: true,
+                    isScrollDisabledOnPopupEnabled: true,
+                    isSelectedAlertsEnabled: false,
+                    isAlertsEnabled: false
+                });
```
oliverfoster: this section shows all the features available from the jquery.a11y library

src/core/js/libraries/jquery.a11y.js
```diff
@@ -1,532 +1,861 @@
...
+            isFocusLimited: false,
+            isRemoveNotAccessiblesEnabled: true,
+            isAriaLabelFixEnabled: true,
+            isFocusWrapEnabled: true,
+            isScrollDisableEnabled: true,
+            isScrollDisabledOnPopupEnabled: true,
+            isSelectedAlertsEnabled: false,
+            isAlertsEnabled: false,
+            isDebug: false
+        };
```
oliverfoster: this shows what is on in the library by default, when accessibility._isEnabled: false

brian-learningpool:
+1 for @oliverfoster's changes.  I've tested this with the sample course from v1.1 and it behaves as expected.

 @dancgray, this reverts the tab behaviour to how it worked on 1.x when accessibility was not enabled.  I don't think accessibility should be enabled by default though.

oliverfoster:
+1 for all of Brian's stuffs :)

taylortom:
Tabbing *seems* to be working fine now, except that the focus outline isn't there for me.
Everything else looks great tho.

oliverfoster:
did you turn accessibility isEnabled:true in the config tom? outline works fine for me, failing that it'll be a theme issue. can we get your +1?

taylortom:
Yep, is enabled. Just to note that this is me testing a v1 course in the v2 framework.

moloko:
+1

taylortom:
+1

PR #655, opened 2015-07-13
ABU-946: Fix for newest backbone library

oliverfoster:
* updated backbone library to v1.2.1
* removed compatibility layer for v1.0.0
* fixed all uses of deprecated functionality

allows us to keep inline with backbone as released


moloko:
+1

brian-learningpool:
+1

zenduo:
+1

PR #675, opened 2015-07-29
Fix gruntfile JSON Check

oliverfoster:
Fix issue with broken json check on multiple course folders

* also added missing ids check (referenced parents which do not exist)


zenduo:
+1

belal-almassri-spongeuk:
+1

dancgray:
+1

PR #677, opened 2015-07-30
Fix for IE11, adding a class (ie) to html element.

belal-almassri-spongeuk:
On ie browsers, adapt should give a class name (ie) to the html element, but in ie 11 latest adapt gives a class(internet explorer) to the html element, This fix will catch (internet explorer) in the string and replace it with (ie).


zenduo:
+1

brian-learningpool:
+1

taylortom:
+1

Issue #587, opened 2015-04-23
Issue with router _canNavigate implementation


brian-learningpool:
There are some issues with the implementation of `_canNavigate` property.

This can be set to `false` by any component, at any time, and break routing to home or the back button  if it is not unset.  Furthermore, unless I'm mistaken it's essentially global, and one component "tidying up" after itself could disable the planned for behavior of another plugin.

Maybe we could open up some discussion on this?

oliverfoster:
i'm going for an Adapt.trigger("router:lock") && Adapt.trigger("router:unlock") scenario. Whereby each lock is a +1 and each unlock is a -1. When the lock is at 0 the router is allowed to navigate. Could make this named much like the old lockedAttributes, whereby Adapt.trigger('router:lock', 'nameOfPlugin') will +1 to the total, but Adapt.trigger('router:unlock', 'nameOfPlugin') would only unlock the instance created by its name. Mean that each extension using router:lock is forced to manage its lock.

oliverfoster:
https://adaptlearning.atlassian.net/browse/ABU-674 < fixed detailed here

oliverfoster:
This is in already.

PR #693, opened 2015-08-10
ABU-674: Locking as optional utility

oliverfoster:
* changes Adapt object from _.extend({}, Backbone.Events) to Backbone.Model to accommodate run-time variables in a consistent way
* extended Backbone.Model to include optional attribute locking, the locking mechanism is one lock per plugin name, to set or unset
* added _canScroll to Adapt model to block Adapt.scrollTo function from performing the scroll (useful for block sliders when they should manage their own scrolling)
* made Adapt.get("_canScroll") a locked attribute
* made Adapt.router.get("_canNavigate") a locked attribute (to overcome brians issue (https://github.com/adaptlearning/adapt_framework/issues/587)
* extended router to block all navigation events when using Adapt.router.set("_canNavigate", false, { pluginName: "plugin" });
* made router block navigation events when content is loading (to solve media element issue in ie8 on a navigate+back in quick succession)
* kept .set(attributeName, value, options) to old backboneModel.js implementation using {pluginName: ''}

This is the long awaited re-implementation of backboneModel.js. Rather than being a mediator, this is a locking mechanism. It is not enabled for every model, only for models where is it explicitly enabled. It is enabled on the Adapt and Adapt.router models at present, for _canScroll and _canNavigate respectively.

New backbone model functions can be found in https://github.com/adaptlearning/adapt_framework/blob/ABU-674/src/core/js/models/lockingModel.js

Description of new Backbone.Model functions:

```
set(attrName, attrVal, options)  : an extension to the original function

  Model.set("_canNavigate", false, { pluginName: "plugin" });
  Model.set("_canNavigate", true, { pluginName: "plugin" });


setLocking(attrName, defaultLockValue) : add locking to a model attribute

    Model.setLocking("_isVisible", false);


unsetLocking(attrName) : remove locking from a model attribute

    Model.unsetLocking("_isVisible");


isLocking(attrName) : returns t/f if the attribute is a locking attribute

    Model.isLocking("_isVisible");


isLocked(attributeName, options) : returns t/f/undefined is the attribute it locked, if it is locked for the supplied pluginName or undefined if it is not a locking attribute

    Model.isLocked("_isVisible");
    Model.isLocked("_isVisible", {pluginName: "plugin"});


getLockCount(attributeName, options) : returns an integer/undefined for the number of locks held in total for the attribute, or for the supplied pluginName or undefined if the attribute is not a locking attribute

    Model.getLockValue("_isVisible");
    Model.getLockValue("_isVisible", {pluginName: "plugin"});


setLockState(attributeName, value, options) : sets the locking value for the given pluginName to value (t/f)

    Model.setLockValue("_isVisible", true, {pluginName: "plugin"});


```

### DIFF
--- a/src/core/js/models/lockingModel.js
+++ b/src/core/js/models/lockingModel.js
@@ -21,7 +21,7 @@ define([
 				this.lockedAttributes[attrName] = !this.defaults[attrName];
 			}
 			var lockingValue = this.lockedAttributes[attrName];
-			var isAttemptingToLock = lockingValue === attrVal;
+			var isAttemptingToLock = (lockingValue === attrVal);
 
 			if (isAttemptingToLock) {
 
@@ -63,8 +63,8 @@ define([
 		},
 
 		isLocking: function(attrName) {
-			var isCheckingGeneralLockingState = attrName === undefined;
-			var isUsingLockedAttributes = this.lockedAttributes !== undefined;
+			var isCheckingGeneralLockingState = (attrName === undefined);
+			var isUsingLockedAttributes = (this.lockedAttributes !== undefined);
 
 			if (isCheckingGeneralLockingState) {
 				return isUsingLockedAttributes;
@@ -110,7 +110,7 @@ define([
 			}
 
 			var lockingAttributeValues = _.values(this._lockedAttributesValues[attrName]);
-			var lockingAttributeValuesSum = _.reduce(lockingAttributeValues, function(memo, value){ return memo + (value ? 1 : 0); }, 0);
+			var lockingAttributeValuesSum = _.reduce(lockingAttributeValues, function(sum, value){ return sum + (value ? 1 : 0); }, 0);
 			
 			return lockingAttributeValuesSum;
 		},
--- a/src/core/js/models/lockingModel.js
+++ b/src/core/js/models/lockingModel.js
@@ -1,0 +1,137 @@
+define([
+	'backbone'
+], function() {
+
+	var set = Backbone.Model.prototype.set;
+
+	_.extend(Backbone.Model.prototype, {
+
+		set: function(attrName, attrVal, options) {
+			var stopProcessing = !this.lockedAttributes || typeof attrName === "object" || typeof attrVal !== "boolean" || !this.isLocking(attrName);
+			if (stopProcessing) return set.apply(this, arguments);
+
+			var isSettingValueForSpecificPlugin = options && options.pluginName;
+			if (!isSettingValueForSpecificPlugin) {
+				console.error("Must supply a pluginName to change a locked attribute");
+				options.pluginName = "compatibility";
+			}
+
+			var pluginName  = options.pluginName;
+			var lockingValue = this.lockedAttributes[attrName] = !this.defaults[attrName];
+			var isAttemptingToLock = lockingValue === attrVal;
+
+			if (isAttemptingToLock) {
+
+				this.setLockValue(attrName, true, {pluginName:pluginName, skipcheck: true});
+
+				//console.log(options.pluginName, "locking", attrName, "on", this.get("_id"));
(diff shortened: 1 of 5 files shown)


src/core/js/models/lockingModel.js
```diff
@@ -0,0 +1,137 @@
...
+			delete this.lockedAttributes[attrName];
+			delete this._lockedAttributesValues[attrName];
+			if (_.keys(this.lockedAttributes).length === 0) {
+				delete this.lockedAttributes;
+				delete this._lockedAttributesValues;
+			}
+		},
+
+		isLocking: function(attrName) {
+			var isCheckingGeneralLockingState = attrName === undefined;
```
moloko: sorry I know this is a bit picky, but I think these sorts of statements are more obvious if you do:
```var isCheckingGeneralLockingState = (attrName === undefined);```

File: src/core/js/router.js [hunk 2]
```diff
@@ -17,12 +16,57 @@ define(function(require) {
...
+            "id/:id":"handleRoute",
+            ":pluginName(/*location)(/*action)": "handleRoute"
+        },
+
+        handleRoute: function() {
+            var args = [].slice.call(arguments, 0, arguments.length);
+            if (arguments[arguments.length-1] === null) args.pop();
+
+            
+            if (!this._isCircularNavigationInProgress) {
```
moloko: an explanatory comment for the `_isCircularNavigationInProgress` property would be useful

moloko:
+1

src/core/js/models/lockingModel.js
oliverfoster: cool, thanks man, i'll update it now

moloko:
+1

tomgreenfield:
+1

src/core/js/models/lockingModel.js
```diff
@@ -0,0 +1,143 @@
...
+	_.extend(Backbone.Model.prototype, {
+
+		set: function(attrName, attrVal, options) {
+			var stopProcessing = !this.lockedAttributes || typeof attrName === "object" || typeof attrVal !== "boolean" || !this.isLocking(attrName);
+			if (stopProcessing) return set.apply(this, arguments);
+
+			var isSettingValueForSpecificPlugin = options && options.pluginName;
+			if (!isSettingValueForSpecificPlugin) {
+				console.error("Must supply a pluginName to change a locked attribute");
+				options.pluginName = "compatibility";
```
dancgray: This is going to cause an error if set is called and options is undefined. Only a problem I guess if you have a combination of old and new plugins setting a locked value. e.g latest assessmentResults and trickle v2.0. This combination breaks the build. Would this be considered a breaking change?

oliverfoster: It'll function fine, you'd just get the error as a warning. We could change it to a console.log if that makes it clearer.

dancgray: console.error is fine, it was the following line (options.pluginName = "compatibility";) that causes the error
"Uncaught TypeError: Cannot set property 'pluginName' of undefined" with the plugins described. 

I'm being overly picky but it was more the awareness of a break in build that is caused with certain plugin version combinations that I was interested in. Just need to add a check for options before defining pluginName = compatibility, should then work with the v2.0 plugins that set the locked attribute :)

oliverfoster: Ahh I see. Yea just and options = options || {}; gotcha. Cool. Bug no1.

oliverfoster: Sorry, tired and skim reading. Not a good combination.

oliverfoster:
@dancgray does this solve it?

dancgray:
Yep +1

moloko:
+1

hrajotia:
+1

PR #705, opened 2015-08-24
Assessment, Trickle applied to block or article

belal-almassri-spongeuk:
In the assessments page, the trickle can be applied to block or article, not article only.
https://github.com/adaptlearning/adapt-contrib-trickle

### DIFF
--- a/src/course/en/components.json
+++ b/src/course/en/components.json
@@ -992,7 +992,7 @@
 			},
 			{
 				"text": "Block",
-				"_shouldBeSelected": false
+				"_shouldBeSelected": true
 			},
 			{
 				"text": "Component",


oliverfoster:
Use _onChildren: false for article article trickle.

belal-almassri-spongeuk:
Hi Ollie,

I thought we can add the trickle to a particular blocks in an article as well. So the question answer for the following question should be Article and block?!

The Trickle extension makes it possible to control how the learner scrolls down the page, but what element of a page is it applied to?

oliverfoster:
it can be applied to blocks or to the article, if on the article it can be applied to all the article's blocks or to the article itself.

belal-almassri-spongeuk:
Cool then add +1 :-)

oliverfoster:
+1 i got you.

zenduo:
+1

taylortom:
+1

PR #706, opened 2015-08-24
Assessment, Trickle applied to block or article

belal-almassri-spongeuk:
In the assessments page, the trickle can be applied to block or article, not article only.
https://github.com/adaptlearning/adapt-contrib-trickle


oliverfoster:
+1 i got you.

zenduo:
+1

taylortom:
+1

PR #708, opened 2015-08-24
Added SCORM interactions support. See also adapt-contrib-spoor.

chris-steele:
https://adaptlearning.atlassian.net/browse/ABU-690

### DIFF
--- a/src/core/js/views/questionView.js
+++ b/src/core/js/views/questionView.js
@@ -231,7 +231,9 @@ define(function(require) {
             // Triggers setCompletionStatus and adds class to widget
             this.checkQuestionCompletion();
 
-            // Used to setup the feedback by checking against
+            this.recordInteraction();
+            
+            // Used to setup the feedback by checking against 
             // question isCorrect or isPartlyCorrect
             this.setupFeedback();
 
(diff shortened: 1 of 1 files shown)


src/core/js/views/questionView.js
```diff
@@ -320,6 +322,16 @@ define(function(require) {
 
         },
 
+        recordInteraction:function() {
+            var config = Adapt.config.get('_spoor');
```
oliverfoster: can we remove the spoor specific stuff from here? the framework should be agnostic., it should preferably do this check after the recordInteraction event

oliverfoster:
did you still want the this.model.get("_recordInteraction") check from the model?

chris-steele:
Yep - there'll be times when interactions will only be needed for certain questions. That's been the case in Yorkshire Water (ken15092) so decided the granularity of choice would be good.

oliverfoster:
it'd make sense to have that bit here if you've put it in spoor.

chris-steele:
Ah see what you mean, yes, agreed.

oliverfoster:
+1 develop pr needed also

taylortom:
+1

hrajotia:
+1

PR #712, opened 2015-08-25
JS device width update

andrewhosgood:
`window.innerWidth` (if available) returns the viewport width of the window (including scrollbars) like the CSS media queries, unlike `$window.width()`, which excludes the scrollbars.
This ensures that the size classes applied to the `<html>` element match the CSS queries.


jasonbutler:
+1

zenduo:
-1 :-1:  this didn't work on mac chrome for me but i can see the issue that it's having, its detecting the width wrong.

zenduo:
I guess this isn't really an issue for mobile / tablet devices (which is where this issue kicks in) as those wont have scrollbars down the side of them, as they are typically placed over the top.

oliverfoster:
I don't think this is solvable with JavaScript and window inner and outer have the same value, plus calculating the scrollbar width is prone to problems. Could we perhaps identify the browsers and operating systems and do a CSS correction?

PR #715, opened 2015-08-26
PLP and menu globals moved to plugin specific fields

dancgray:
PLP and menu globals match respective handlebars templates.

Also remove some redundant aria labels.


oliverfoster:
Could you add in "visited": "Visited" to the globals bit? We seem to be using that a fair bit for accessibility. I think @guywillis wanted to include it at some point today.

oliverfoster:
+1

moloko:
+1

hrajotia:
+1

PR #722, opened 2015-09-09
Various updates and improvements

andrewhosgood:
I have tested the `window.innerWidth` fix, which produces different (correct) results on Windows machines where scrollbars are used. I logged the two variables, `window.innerWidth` and `$window.width()` and they produce different numbers, with the `window.innerWidth` producing the same number as the width used by CSS media queries:
code
window-width
This does not affect browsers that don't have a scrollbar, as they will function as before.

I have added support for detecting pixel density which allows background images to be changed based on the number of pixels per CSS pixel (such as Apple's "Retina" displays).

When using `document.title =`, any HTML entities (such as `&amp;`, `&mdash;`, `&pound;` etc.) in the value passed are escaped. Setting the value with jQuery's `.html()` means entities are preserved.

Prepending the drawer and notification element to the body allows for more dynamic CSS styles, such as `.notify ~ #wrapper` which would allow you to style elements in the wrapper when the notification modal is open.


belal-almassri-spongeuk:
+1

oliverfoster:
+1

hrajotia:
+1

belal-almassri-spongeuk:
@andrewhosgood, can we have PR for the master branch as well?

PR #728, opened 2015-09-17
Don't override template if already set

taylortom:
Allows devs to more easily extend existing components, by making available existing templates. Currently you're forced to create a new template with the name registered by your component.

This allows you to do such things as:
```
var TextDerivitive = Text.extend({
    template: "text",
    ...
});
```


oliverfoster:
Hell yes! +1

moloko:
+1

hrajotia:
+1

PR #744, opened 2015-09-25
re-added 'navigateToHomeRoute' function

### LINKED ISSUE #743: The router navigateToHomeRoute() function has been missed from v2
I'm not sure how this was missed but it needs to be added ASAP.

hrajotia:
Fixes #743

### DIFF
--- a/src/core/js/router.js
+++ b/src/core/js/router.js
@@ -17,6 +17,7 @@ define(function(require) {
                 document.title = Adapt.course.get('title');
             });
             this.listenTo(Adapt, 'navigation:backButton', this.navigateToPreviousRoute);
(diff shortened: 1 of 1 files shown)


brian-learningpool:
+1 -- thanks :+1:

oliverfoster:
+1 -- what's it for?

brian-learningpool:
It allows themes to have a 'home' button, which when clicked takes the user back to the home page.  The `backButton` handler only goes to the previous route, which can often mean multiple clicks before getting back to the menu.

oliverfoster:
ahh, ok, so there are plans to use it, coool

dancgray:
+1

PR #754, opened 2015-10-02
ABU-1075 - store uid on dom object

oliverfoster:
a jquery object is receiving a uid instead of the dom object, so saving and restoring dom object tabindexes is flawed.
need to change to assigning a uid to the dom object so that it's tabindex can be restored correctly.

change made.


moloko:
+1

tomgreenfield:
+1

hrajotia:
+1

PR #759, opened 2015-10-06
Modified code in setupPartlyCorrectFeedback

sarveshwar-gavhane:


moloko:
thanks for the update, just been testing and have found that if you:
* don't have partly correct feedback set
* have `_attempts` set to 2 (or more)
* have 'notFinal' incorrect feedback set
* get the question partly correct on the first attempt
Then the 'final' incorrect feedback is set rather than than the 'notFinal', which is what I would expect.

Would it not be better to call `setupIncorrectFeedback()` whenever there isn't appropriate `_partlyCorrectFeedback` set? i.e. instead of lines 415 - 418?

moloko:
Replacing lines 415 - 418 with
`this.setupIncorrectFeedback();`
Seems to solve the problem for me

sarveshwar-gavhane:
okay thats great.... :)

moloko:
+1

brian-learningpool:
+1

oliverfoster:
+1

PR #763, opened 2015-10-09
Removed base css and moved to buttons.less

guywillis:
Removed override as it was affected buttons that had not been converted from <a> tags. 

Removed content-box sizing as it was causing rendering errors with components such as the accordion item title.


oliverfoster:
+1

moloko:
+1

lc-thomasberger:
+1

PR #764, opened 2015-10-09
bugfix: ie8 document title fix

oliverfoster:
jquery.fn.html cannot be used to change head elements in ie8. will through an error.

### DIFF
--- a/src/core/js/router.js
+++ b/src/core/js/router.js
@@ -242,7 +242,11 @@ define([
             var documentTitle = courseTitle + pageTitle;
 
             Adapt.once("pageView:ready menuView:ready", function() {
-                $('title').html(documentTitle);
+                if ($("html").is(".ie8")) {
+                    document.title = documentTitle;
+                } else {
+                    $('title').html(documentTitle);
+                }
             });
 
         }
--- a/src/core/js/router.js
+++ b/src/core/js/router.js
@@ -239,14 +239,10 @@ define([
             }
 
             var courseTitle = Adapt.course.get("title");
-            var documentTitle = courseTitle + pageTitle;
+            var documentTitle = $("<div>" + courseTitle + pageTitle + "</div>").text();
 
             Adapt.once("pageView:ready menuView:ready", function() {
-                if ($("html").is(".ie8")) {
-                    document.title = documentTitle;
-                } else {
-                    $('title').html(documentTitle);
-                }
+                document.title = documentTitle;
             });
 
         }


oliverfoster:
@guywillis this should get it working. menu and pages now load. buttons in resources work nicely. course is actually quite snappy in winxp ie8. looks lovely.

moloko:
is the `if...else` really necessary? couldn't you just use:
```
document.title = courseTitle + pageTitle;
```

oliverfoster:
The previous or was to allow special characters to be parsed properly but it broke ie8.

https://github.com/adaptlearning/adapt_framework/commit/9e0d14497dd1bd2288dd68ca8703028221a8ef43

oliverfoster:
Can probably include HTML character parsing in ie8 also and make a more general PR if you'd like?

tomgreenfield:
I've just had a play around with using special characters in the course title via the JSON and I'm not entirely sure why one would need to use HTML entities when they already render out fine.

For example,
* ampersand `&`
* en dash `–`
* pound sign `£`

See [here](https://github.com/adaptlearning/adapt_framework/pull/722#issue-105558983) for original comment.

moloko:
maybe the authoring tools converts those characters to HTML entities automatically?

andrewhosgood:
I you use characters like `&` and `£` in the article title, they will be escaped properly within the document title but not in the article header (although they will still work in *most* browsers).
You should really be using `&amp;` and `&pound;` to properly encode entities within the page (and within the JSON files), but this causes issues with the document title where the entities used to be escaped, so `&amp;` became `&amp;amp;` and `&pound;` became `&amp;pound;`.
This is only to make sure the source files are entirely ASCII rather than relying on UTF-8 and also means the page should then be W3C compliant.

oliverfoster:
I think trying to impose the proper use of html entity escaping in JSON is quite an untenable position. No only is it wholly unnecessary - as handlebars performs escaping before Adapt outputs to the DOM - but in doing so causes issues in the document title as above. 
That being said, I do understand the want to handle characters of different formats appropriately. We can provide a solution for both camps which allows the document title text to be properly formatted by converting all the html entities before output, even if a change of practice could entirely mitigate the issue.

http://stackoverflow.com/questions/24816/escaping-html-strings-with-jquery

andrewhosgood:
Absolutely. If we can find a way to standardise the data on its way from the JSON to the Handlebar templates then this won't be an issue in the future.
As usual, modern browsers don't really care and don't have an issue rendering un-entitied characters and it's only IE8 that causes problems.
It was only my anal personality, adding all the named HTML entities for `&quot;` etc. into the JSON, that was causing an issue anyway.

oliverfoster:
The templates just need a triple curly bracket to make sure escaping is done properly http://handlebarsjs.com/#html-escaping

So any JSON which is passed into the handlebars templating engine can be escaped or not with double or curly braces respectively.

The only part which hasn't any escaping available or applied is the document.title line.

oliverfoster:
@andrewhosgood @moloko @tomgreenfield how's this?

andrewhosgood:
Looks good to me :+1:

moloko:
+1

tomgreenfield:
+1

Issue #770, opened 2015-10-13
develop, adapt.js, Adapt.register


LukaszGrela:
Recently there was a test added for existance of template property on the object passed, the problems caused by this can be unnoticed for a long time as it only occurs on components that already inherits (extends other) this property from other components

https://github.com/adaptlearning/adapt_framework/blob/develop/src/core/js/adapt.js#L100

the one that I've encountered this misbehaviour was on the GMCQ that extends MCQ and the object already has the template with the "mcq" value.

LukaszGrela:
Fixed with https://github.com/adaptlearning/adapt-contrib-gmcq/pull/87/files

PR #771, opened 2015-10-13
Don't override template if already set - fix to issue #770

LukaszGrela:
Fixes this change:
https://github.com/adaptlearning/adapt_framework/commit/a7af2e3f8713979f3b8933ed6c443f254ec6eb27#diff-dc2a9e6689317375d0da7294e71344c8
to avoid issue reported here:
https://github.com/adaptlearning/adapt_framework/issues/770

### DIFF
--- a/src/core/js/adapt.js
+++ b/src/core/js/adapt.js
@@ -97,7 +97,7 @@ define([
         // Store the component view
         if (Adapt.componentStore[name])
             throw Error('This component already exists in your project');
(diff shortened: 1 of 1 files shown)


oliverfoster:
This breaks the intention of the original line.

LukaszGrela:
Then this line would be removed altogether and all components should have the template set like @hrajotia proposed in the GMCQ modification

PR #785, opened 2015-10-26
ABU-1038: Accessibility reset checks

oliverfoster:
Check if component and page are ready before allowing focus to be set by question component reset functions.

Set focus to beginning of question on invalid submit click.


moloko:
+1

dancgray:
+1

hrajotia:
+1

PR #787, opened 2015-10-26
bugfix/accessibility

oliverfoster:
https://adaptlearning.atlassian.net/browse/ABU-1039:
* If the course is closed and then relaunched there is no way to switch on accessibility mode or tab to the yes/no buttons on the bookmarking pop-up. This also affects courses if there is an initial pop-up (eg a explanatory/help pop up).  
* fixed, framework now remember accessibility state

https://adaptlearning.atlassian.net/browse/ABU-1095:
* when the pop-up opens voiceover either reads 'index htlm link' or focus is on the first hyperlink and not the 'pop-up opened' aria. Close button is not in the tabbing sequence. When the close button is selected directly and the pop-up closes focus returns to the top of the page.  
* fixed, issue with notify having moved to the top of the document before the content and the focusguard at the bottom of the document wasn't capturing the last tab properly.
the focusguard now gets injected into the appropriate 'endpoint' in the document. either in the body or in the relevant popup.

https://adaptlearning.atlassian.net/browse/ABU-1091:
* If the user tabs through the resources/PLP drawer focus moves to the elementsbehind then to the top of the page.  
* part two, fixing the focusguard to inject into the bottom of popups, or in this case, the drawer

https://adaptlearning.atlassian.net/browse/ABU-1093:
* When feedback pop-up is opened the close button is not in the tabbing sequence.  
* fixed with injecting focusguard into the bottom of the relevant popup or in the document body

### DIFF
--- a/src/core/js/accessibility.js
+++ b/src/core/js/accessibility.js
@@ -10,7 +10,7 @@ define(function(require) {
         $accessibilityToggle: $("#accessibility-toggle"),
 
         _tabIndexElements: 'a, button, input, select, textarea, [tabindex]',
-        _hasTabPosition: false,
+        _isButtonRedirectionOn: true,
         _hasUsageInstructionRead: false,
         _isLoaded: false,
         _hasCourseLoaded: false,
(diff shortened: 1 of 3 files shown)


hrajotia:
+1

oliverfoster:
moving to master

Issue #792, opened 2015-10-30
adapt.js - missing underscore from currentLocation


LukaszGrela:
I think in the line https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/adapt.js#L22 there should be an underscore for `currentLocation`, it should read `Adapt.location._currentLocation` as it is used e.g. in router https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/router.js#L147

LukaszGrela:
Same in Develop and Master branch

LukaszGrela:
reopening as not yet in master

moloko:
should be in master now?

LukaszGrela:
I meant to keep it open until this fix from dev will be merged into master

moloko:
I merged dev into master earlier today...

PR #793, opened 2015-10-30
Fix to issue #792 - Added missing underscore

LukaszGrela:
It should be Adapt.location._currentLocation as it is used e.g. in router.js


oliverfoster:
develop and master are about to merge, you can probably get rid of this one.

LukaszGrela:
sure

Issue #804, opened 2015-11-05
"page", article, block, component - missing _classes adds 'undefined' to the class list


LukaszGrela:
I've noticed if I skip the `_classes` property then `undefined` class is added, could we add to the models that support `_classes` default empty value so we could then skip unused property?

oliverfoster:
Here are two examples:
https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/componentView.js#L12
https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/articleView.js#L11


You'll need to go through each view to check them:

https://github.com/adaptlearning/adapt_framework/tree/master/src/core/js/views

LukaszGrela:
wouldn't work modification of the models? e.g. for ArticleModel:

```
define(function(require) {

	var AdaptModel = require('coreModels/adaptModel');
	var Adapt = require('coreJS/adapt');

    var ArticleModel = AdaptModel.extend({
        _parent:'contentObjects',
    	_siblings:'articles',
        _children: 'blocks',
        _classes: ''
    });

    return ArticleModel;

});

```

oliverfoster:
you could try adding it here > https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/models/adaptModel.js#L18

LukaszGrela:
that's fine, I will do this on each "view" class

PR #806, opened 2015-11-09
Lukasz grela patch issue #804

LukaszGrela:
check on the existence of _classes property so there are not `undefined` in the class list, fix issue #804

### DIFF
--- a/src/core/js/views/articleView.js
+++ b/src/core/js/views/articleView.js
@@ -4,11 +4,11 @@ define(function(require) {
     var BlockView = require('coreViews/blockView');
 
     var ArticleView = AdaptView.extend({
-
+        var _classes = this.model.has('_classes') ? " " + this.model.get('_classes'):"";
         className: function() {
             return "article "
             + this.model.get('_id')
-            + " " + this.model.get('_classes')
+            + _classes
             + " " + this.setVisibility()
             + " nth-child-"
             + this.model.get("_nthChild");
--- a/src/core/js/views/blockView.js
+++ b/src/core/js/views/blockView.js
@@ -5,9 +5,10 @@ define(function(require) {
     var BlockView = AdaptView.extend({
 
         className: function() {
+            var _classes = this.model.has('_classes') ? " " + this.model.get('_classes'):"";
             return "block "
             + this.model.get('_id')
-            + " " + this.model.get('_classes')
+            + _classes
             + " " + this.setVisibility()
             + " nth-child-"
             + this.model.get("_nthChild");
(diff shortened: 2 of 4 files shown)


moloko:
+1

oliverfoster:
+1

brian-learningpool:
-1, sorry @LukaszGrela, I actually think that @oliverfoster's suggestion of moving the default to the AdaptModel.js is a more elegant solution.  As all these views inherit from AdaptView you also could've added the default there.

LukaszGrela:
@brian-learningpool I do agree that one place is best:) however (I have to check again) my attempt was unsuccessful with using model and also this will add `_classes` to all models that not necessarily are using it.

brian-learningpool:
On reflection, PageView, ArticleView, BlockView and ComponentView all inherit from AdaptView so I think this is probably a better place to put it.

LukaszGrela:
@brian-learningpool like so?

```
        initialize: function() {
            this.listenTo(Adapt, 'remove', this.remove);
            this.listenTo(this.model, 'change:_isVisible', this.toggleVisibility);
            this.model.set('_globals', Adapt.course.get('_globals'));
            this.model.set('_isReady', false);

            if (!this.model.has('_classes')) this.model.set('_classes', '');

            this._isRemoved = false;
            this.preRender();
            this.render();
        },
```

brian-learningpool:
Hi @LukaszGrela, yes that's what I was thinking.  Does that resolve it for you?

oliverfoster:
className() executes as the containing DOM object is created, which is done before initialize() is called. The DOM object will already contain class="undefined" by that point. You'll have to do it as a default on the model if this is to be a global fix.

LukaszGrela:
Adding `_classes` to the defaults of the `AdaptModel` works
```
defaults: {
            _canShowFeedback: true,
            _canReset: false,
            _isComplete: false,
            _isInteractionComplete: false,
            _isEnabled: true,
            _isResetOnRevisit: false,
            _isAvailable: true,
            _isOptional: false,
            _isTrackable: true,
            _isReady: false,
            _isVisible: true,
            _classes: ''
        }
```

however I feel it is wrong as this adds property that may not be used by all children of this model, could we just extend `defaults` in the proper models like so:

```
    var BlockModel = AdaptModel.extend({
        _parent: 'articles',
        _siblings: 'blocks',
        _children: 'components',
        defaults: _.extend({}, AdaptModel.prototype.defaults, {
            _classes: ''
        })
    });
```

oliverfoster:
The simple one works best and fits in with the way everything else works. If you add it to defaults now we can have a discussion about view and model separation when we do the next major version?

taylortom:
@LukaszGrela @oliverfoster what's the status on this? are we still aiming to get it in?

oliverfoster:
closed for alternate pr #1012

PR #807, opened 2015-11-09
bugfix/inview

oliverfoster:
Sometimes expando properties don't work well and so to uniquely identify each element it is necessary to give it a unique expando parameter.

Without this fix, bindings on multiple elements with the same callback function won't work. They'll callback once and only on the last element attached.

### DIFF
--- a/src/core/js/libraries/inview.js
+++ b/src/core/js/libraries/inview.js
@@ -2,6 +2,7 @@
 
 (function() {
 	var expando = $.expando;
+	var expandoIndex = 0;
 
 	//element + event handler storage
 	var onScreenObjs = {};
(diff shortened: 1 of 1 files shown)


moloko:
+1

brian-learningpool:
+1

andrewhosgood:
This doesn't break anything, but I didn't have any issues before: +1?

oliverfoster:
it's more a tech / backend thing allows:

```
Backbone.View.extend({

	setupItemEventListeners: function() {
		var $items = this.$(".item");
		var callback = _.bind(this.onInview, this);

		for (var i = 0, l = $items.length; i < l; i++) {
			$items.on("inview", callback);
		}
	},

	onInview: function() {

	}


});
```

rather than:

```
Backbone.View.extend({

	setupItemEventListeners: function() {
		var $items = this.$(".item");

		for (var i = 0, l = $items.length; i < l; i++) {
			$items.on("inview", _.bind(this.onInview, this));
		}
	},

	onInview: function() {

	}


});
```

as the former would call back only once on the last attached element if jquery was misbehaving.

very odd behaviour but fixed with this code.

PR #808, opened 2015-11-10
bugfix/accessibility: popdown fix

oliverfoster:
Cannot focus on disabled or removed elements when coming down from a popup. Fix finds the next focusable element and sets focus and scrolls.


moloko:
+1

brian-learningpool:
+1

hrajotia:
+1

Issue #805, opened 2015-11-06
Tight coupling between some components


brian-learningpool:
There are some components which either inherit from another component, e.g. GMCQ and MCQ, or are dependent on another component, e.g. Hot Graphic and Narrative.  Should we plan on de-coupling these, so each component is standalone?

oliverfoster:
was thinking about trying to kill two birds with one stone. https://adaptlearning.atlassian.net/browse/ABU-699 < size switcher?

moloko:
@brian-learningpool https://adaptlearning.atlassian.net/browse/ABU-691

moloko:
we discussed these at the start of the current sprint and the thinking was that maybe this should be part of Framework v3

brian-learningpool:
@moloko I think that it would be great to tidy this up as part of v3.  The new superclass approach mentioned in the JIRA ticket sounds like a potential solution.

moloko:
@brian-learningpool there's lots of things like this that I'd like to sort out as part of v3. Want to give the whole framework's structure a really good re-think and tidy-up - so if you spot anything else like this please do log a ticket in Jira.

chris-steele:
@moloko looks like this can be closed, would you say?

moloko:
yes, I think best dealt with under https://github.com/adaptlearning/adapt_framework/issues/835

PR #810, opened 2015-11-11
Fixes ABU-1128

taylortom:
Refactoring of the grunt process.
***
#### TODO:
- Fix up the `help` task


taylortom:
@brian-learningpool brought up a fairly serious issue regarding **including** plugins (this is the method that we're going to be using in the AT):

If you include a plugin which inherits for something else, but not the original plugin (i.e. in the case of MCQ and GMCQ), you run into a sticky situation.

We need to solve this before this refactor can go in really.

brian-learningpool:
For reference, this has been raised as #805.  In order to progress things though, a temporary/patch solution would be to support the core plugins which have dependencies on others.  Off the top of my head, the logic would be something like:
- ensure adapt-contrib-mcq is bundled if adapt-contrib-gmcq is bundled
- ensure adapt-contrib-narrative is bundled if adapt-contrib-hotgraphic is bundled
- ensure adapt-contrib-slider is bundled if adapt-contrib-confidenceSlider is bundled, 
etc.

Or we could add support for always bundling specific plugins (configurable) as part of the compilation?

oliverfoster:
can we not list dependencies in the bower jsons and have it check when it bundles the plugins?

taylortom:
That sounds sensible to me.

I think we also need to do something from the CLI end to pull in the relevant dependencies during an `adapt install`

taylortom:
@oliverfoster @brian-learningpool see https://github.com/adaptlearning/adapt_framework/pull/818 for a partial fix for the dependencies. 

I think we also still need something from the CLI-side.

PR #813, opened 2015-11-17
open default item in the drawer.

sarveshwar-gavhane:


src/core/js/drawer.js
```diff
@@ -25,7 +25,12 @@ define(function(require) {
 	Adapt.once('app:dataReady', function() {
 		init();
 	})
-
+	Adapt.on("drawer:opened",function(){
+		var itemsLength = $(".drawer-holder")[0].childElementCount;
```
oliverfoster: https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/childElementCount

^ childElementCount not supported in ie8. can you find a different way to do this?

src/core/js/drawer.js
```diff
@@ -25,7 +25,12 @@ define(function(require) {
 	Adapt.once('app:dataReady', function() {
 		init();
 	})
-
+	Adapt.on("drawer:opened",function(){
+		var itemsLength = $(".drawer-holder").children().length;
+		if(itemsLength == 1) { 
+		      Adapt.trigger("resources:showResources");
```
oliverfoster: Should be ``Adapt.trigger(DrawCollection.models[0].eventCallback);`` or similar.

sarveshwar-gavhane:
@oliverfoster I have re-created PR - https://github.com/adaptlearning/adapt_framework/pull/814, could you please review it, as it looks more simplistic.

PR #819, opened 2015-11-20
hotfix/router: missing break and navigateToElement fix

oliverfoster:
1. Add missing ``break;``
2. When using an article/block/component id to navigate, should replace route with content object id. This is to preserve the history of the back button.


moloko:
+1

brian-learningpool:
+1

hrajotia:
+1

PR #820, opened 2015-11-24
amend to slider in c-80

moloko:
Since the correct answer is a single number and not a range, it is confusing to specify it using the `_correctRange` property (even though it does work). Therefore am changing this to the more appropriate `_correctAnswer` property.


hrajotia:
+1

andrewhosgood:
+1

brian-learningpool:
+1

PR #826, opened 2015-12-01
feature/adaptModel: added getParents

oliverfoster:
getParents returns a Backbone.Collection array from [0] first parent to [x] last parent or null
getParents(true) returns a Backbone.Collection array from [0] current model to [x] last parent or null

### DIFF
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -212,6 +212,30 @@ define(function (require) {
             return parent;
         },
 
+        getParents: function() {
+            var context = this;
+            var parents = [];
+            while(true) {
+                if (!context.get("_parentId")) break;
+                context = context.getParent();
+                parents.push(context);
+            }
+            if (parents.length === 0) return null;
+            return new Backbone.Collection(parents);
+        },
+
+        getPath: function() {
+            var context = this;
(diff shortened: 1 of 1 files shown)


src/core/js/models/adaptModel.js
```diff
@@ -212,6 +212,30 @@ define(function (require) {
...
+            var parents = [context];
+            while(true) {
+                if (!context.get("_parentId")) break;
+                context = context.getParent();
+                parents.push(context);
+            }
+            if (parents.length === 0) return null;
+            return new Backbone.Collection(parents);
+        },
+
```
tomgreenfield: Suggest merging the two functions for readability producing something like this:
```js
getParents: function(shouldIncludeChild) {
    var parents = [];
    var context = this;

    if (shouldIncludeChild) parents.push(context);

    while (context.has("_parentId")) {
        context = context.getParent();
        parents.push(context);
    }

    return parents.length ? new Backbone.Collection(parents) : null;
}
```

oliverfoster: +1

hrajotia:
+1

moloko:
+1

brian-learningpool:
+1

PR #831, opened 2015-12-04
hotfix/accessibility

oliverfoster:
- removed ios select fix as it causes more issues that it fixes
- improved onFocusCapture to capture the next tabbable element if the current focus target is not tabbable, has no tabbable parents and is not a proxy for another element


moloko:
+1

brian-learningpool:
+1

hrajotia:
+1

PR #834, opened 2015-12-09
Allow ESC button to close the popup when accessibility is not active

belal-almassri-spongeuk:
Will allow ESC button to close the notification popup when the accessibility is not active rather than is not enabled!


belal-almassri-spongeuk:
Thanks @oliverfoster

oliverfoster:
+1

andrewhosgood:
+1

hrajotia:
+1

PR #889, opened 2016-01-07
Couple of small grunt amends

taylortom:
Finally fixed up help task, although it now only loads the tasks in grunt/tasks, which may not be suitable (better than before, at least!)


brian-learningpool:
+1

PR #920, opened 2016-01-27
Add jshint and jscs

### LINKED ISSUE #201: Create a JSLint file for checking compliance with coding standards

taylortom:
Fixes #201.

Also added jscs which enforces code style (e.g. spacing etc.). Have opened this as a separate PR to allow for discussion re. the actual options here.

See http://jscs.info/rules for more info.

p.s. I haven't integrated either of these into any of the existing tasks due tot the amount of time they take to execute. Depending on what we want to do, we can hook them into git to force devs to run them prior to committing (this could be a bit restrictive though...).

TODO: Look into hooking up to travis


grunt/config/jscs.js
```diff
@@ -0,0 +1,49 @@
...
+        disallowSpaceAfterObjectKeys: true,
+        disallowSpaceAfterPrefixUnaryOperators: true,
+        disallowSpacesInFunction: {
+            beforeOpeningRoundBrace: true
+        },
+        disallowTrailingWhitespace: true,
+        disallowTrailingComma: true,
+        requireCamelCaseOrUpperCaseIdentifiers: true,
+        requireCapitalizedConstructors: true,
+        requireCurlyBraces: true,
```
moloko: I think for `requireCurlyBraces` we'd need a few exceptions to prevent existing code from throwing errors and to allow for clean code e.g. 
` if (parameterNotSet) return;`
at the start of a function

Also for `case` and `default` in `switch` statements - I don't think we ever use curly braces with these

taylortom: Ah yeah, meant to take out the brace switch

Issue #925, opened 2016-01-29
Move towards true MVC


oliverfoster:
add questionModel and component models, move model behaviour from views into relevant models

oliverfoster:
#835

PR #926, opened 2016-01-29
enhancement/925: allow component specific & standard question models

oliverfoster:
#925 

* added questionModel
* moved questionModel code out of componentModel into questionModel
* added code to optionally allow components to register their own model through the ``Adapt.register`` interface.
```
Adapt.register("name", { view: View, model: Model }); //new style
Adapt.register("name", View); //old style - still works fine
```
* changed component model creation to use a component specified model (if available), the questionModel or the componentModel where required (which, as a plus, means any view-only question components will automatically use the questionModel and need no reworking for this PR to fit)
* reworked require statements to AMD style

should be entirely unobtrusive

task following this:
* move questionView model code into questionModel
* provide backwards compatibility layer in questionView to redirect older components to the moved questionModel code
* split all core components into a model and view where relevant

pros:
* all models represent the data and behaviour of their component model (rather than just the data as it is at the moment)
* views no longer contain model initialization or model only code
* score calculation functions are present on unrendered component models and allow extensions to act upon the models without having to first render them into a view
* smaller views = less complex code
* correct partitioning = higher clarity 
* closer to the MVC style = easier for programmers to grasp
* framework will continue to support View only style alongside View+Model style (until we decide to get rid of it)
* if we transition from backbone to another framework having a good mvc aligned implementation will help porting

cons:
* View+Model components will not work in earlier versions of the framework which don't support it, means potential issues with AT framework vs plugin versions

### DIFF
--- a/src/core/js/adapt.js
+++ b/src/core/js/adapt.js
@@ -110,6 +110,7 @@ define([
         
         Adapt.componentStore[name] = object;
 
+        return object;
     }
 
     // Used to map ids to collections
--- a/src/core/js/adapt.js
+++ b/src/core/js/adapt.js
@@ -99,7 +99,15 @@ define([
         // Store the component view
         if (Adapt.componentStore[name])
             throw Error('This component already exists in your project');
-        if(!object.template) object.template = name;
+
+        if (object.view) {
+            //use view+model object
+            if(!object.view.template) object.view.template = name;
+        } else {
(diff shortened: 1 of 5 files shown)


moloko:
+1

dancgray:
+1 nice

taylortom:
+1

Issue #927, opened 2016-01-31
complete cascade in wrong order


oliverfoster:
currently when a component triggers a block completion event, and a block an article, an article a page etc, the change events call back in the wrong order as the final parent relinquishes control of its change:_isComplete handler before the triggering element does.

So if a component completes, causing it's block to complete, causing it's article to complete and you're listening to the complete events, you'll get:

article completed
block completed
component completed

it should be:

component completed
block completed
article completed

oliverfoster:
https://github.com/adaptlearning/adapt_framework/issues/835 point 3

PR #928, opened 2016-01-31
bugfix/927: reverse order of completion events

oliverfoster:
* reworked event attachment from adapt.on to adapt.listento so that the model owns the event handler
* reversed the order of the complete event fires #927


moloko:
+1

LukaszGrela:
+1

taylortom:
+1

Issue #929, opened 2016-01-31
remove core require mappings "coreJS" "coreViews"


oliverfoster:
* abstraction is unneeded, no benefits and a possible hindrance to understanding the code

oliverfoster:
https://github.com/adaptlearning/adapt_framework/issues/835 point 2

PR #930, opened 2016-01-31
enhancement/929: remove core require mappings "coreJS" "coreViews"

oliverfoster:
* removed explicit conversion of module declarations to "coreJS" style from "core/js" (+models, views etc)
* remapped current require("coreJS/adapt) style to "core/js/adapt" in grunt
* remapped current require("coreJS/adapt) style to "core/js/adapt" in browser

#929

actions to follow:
* convert all old require declarations from "coreJS" style to new "core/js/" style
* new new grunt requirejs task needs to be compatible with this style

should be unobtrusive as the old style can be remapped to the new style until we've converted everything.

### DIFF
--- a/config.js
+++ b/config.js
@@ -10,11 +10,6 @@ require.config({
         inview: 'core/js/libraries/inview',
         a11y: 'core/js/libraries/jquery.a11y',
         scrollTo: 'core/js/libraries/scrollTo',
-        coreJS: 'core/js',
-        coreViews: 'core/js/views',
-        coreModels: 'core/js/models',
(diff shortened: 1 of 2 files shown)


moloko:
+1

taylortom:
+1 for concept (need to test for backwards-compatibility)

oliverfoster:
if you replace all the app.js "coreJS/adapt" with "core/js/adapt" you'll get the idea about backwards compatibility. it should just work with either for the time being

dancgray:
+1

taylortom:
+1

PR #931, opened 2016-01-31
ABU-680: new javascript compiler

oliverfoster:
* added compiler
* less fixes
* module updates & removes 
* tracking flatten

need to sort out javascript done and error output @taylortom 
~~need to test module updates work on node v0.10~~ done
~~need to test less sourcemaps come out correctly on windows~~ done

should work with multi user in the AT @brian-learningpool 

!notes are on commit!


taylortom:
I had a look at trying to get this work too, but couldn't find a decent solution.

Looks v.nice, but think this will still run into the same problem in the AT as the old implementation, as we still have a shared `src/plugins.js`.

taylortom:
Also, is there and standard way to specify shared options for targets? It seems silly that we're forced to have to specify most of them twice (in my test I hacked it in, as couldn't find anything useful in the docs).

oliverfoster:
the src/plugins.js file is not modified in the build process, it's only used as an endpoint in the build process to attach dependencies.

oliverfoster:
not that i know of in grunt, but i'm not a grunt expert. having the separate is clear enough for me, task + config is ok.

brian-learningpool:
+1

taylortom:
+1, works well

PR #941, opened 2016-02-15
Added a replace task to the end of grunt server-build

bell87:
Added a grunt config task called replace.

This makes use of [grunt-replace](https://www.npmjs.com/package/grunt-replace) npm library.

The task will replace any strings in xml files (prefixed with `@@`) with matching properties found in the course.json file.

### DIFF
--- a/grunt/config/replace.js
+++ b/grunt/config/replace.js
@@ -1,0 +1,33 @@
+module.exports = function (grunt, options) {
+
+  var courseDir = options.sourcedir + 'course/';
+  if (options.outputdir.indexOf('courses')) {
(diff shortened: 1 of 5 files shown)


brian-learningpool:
Can you add this task to the `build.js` and `dev.js` tasks too?

barry-learningpool:
+1

oliverfoster:
can we try and support globs for the language folders, rather than using a single language?

is this complex enough for your needs? are you going to need to map the course structure from the json? or is a simple string replacement good enough for a standard xml?

oliverfoster:
+1

PR #946, opened 2016-02-16
DISCUSSION: Add new and improved travis config

### LINKED ISSUE #933: Look into hooking up jscs/jshint to travis
Support for jshint/jscs has been added [here](https://github.com/adaptlearning/adapt_framework/pull/891), but compliance is currently at the individual developer's discretion.

Would be nice to get this enforced in someway -- either as a git hook to be called on commit, or at the travis ci level. 

I'm all for the consistency that this should bring, but think we need to be careful that we're not too strict with the rules, as will likely cause many :angry::anger::fist:

taylortom:
`script` process executes: jscs, jshint, build, and build of the changes
against core bundle v2.0.0. Note that `script` processes defined have
been defined in `.travis.yml` rather than `package.json`. Also added a
webhook for integrating with gitter.

Fixes #933

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+language: node_js
+node_js:
+  - 4.3
+branches:
+  only:
+    - master
+    - develop
+    - /^release\/.*$/
+sudo: required
+cache:
+  directories:
+    - node_modules
+before_script:
+  - npm install -g grunt-cli
+  - npm install -g adapt-cli
+  - adapt install
+script:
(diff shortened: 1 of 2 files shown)


moloko:
it's currently failing the Travis CI build...

taylortom:
Yeah, that's because of the code style stuff (jshint/jscs)

taylortom:
I think we could also do with setting up some more useful tests too now we're merging directly into master

Issue #910, opened 2016-01-18
"_requireCompletionOf": "any"


moloko:
Original thread #549 

Moved from [ABU-673](https://adaptlearning.atlassian.net/browse/ABU-673)

moloko:
done in https://github.com/adaptlearning/adapt_framework/commit/a9ab8c7555a5e1bd5ea550aa86effcbc582ea2f5

PR #951, opened 2016-02-16
issue/910: added requireCompletionOf: X

oliverfoster:
#910 

if you need all of the mandatory components to be complete use _requireCompletionOf: -1  (this is as default)
if you need to exclude some of the mandatory components from completion use _isOptional: true (making them non-mandatory)
if you need only X number of mandatory components to complete, use requireCompletionOf: X

### DIFF
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -10,6 +10,7 @@ define(function (require) {
             _canReset: false,
             _isComplete: false,
             _isInteractionComplete: false,
+            _requireCompletionOf: 'all',
             _isEnabled: true,
             _isResetOnRevisit: false,
             _isAvailable: true,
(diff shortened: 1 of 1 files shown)


LukaszGrela:
+1

moloko:
+1
This does all need documenting somewhere of course

One thing to note is that setting `_requireCompletionOf` to `0` triggers much the  the same behaviour as setting `_isOptional` on all components in the block - except that the PLP will not show those components as being optional. I don't think this is problem, so long as we're clear that it does this - perhaps we should recommend in the docs that `_isOptional` is the better method for making all components in a block optional?

oliverfoster:
might be better to make if "0" : false / off ? that way we remove the ambiguity? or introduce more?

taylortom:
+1
Don't think it particularly matters if it's the default? Otherwise, I'd maybe prefer a value like "all" to a number.

Issue #906, opened 2016-01-18
Single page courses + start pages


moloko:
see also #521 and #491

moved from [ABU-679](https://adaptlearning.atlassian.net/browse/ABU-679)

PR #952, opened 2016-02-16
issue/906: Added start controller

oliverfoster:
#906 

    /*example

        "_start": {
            "_isEnabled": true,
            "_startIds": [
                {
                    "_className": ".size-small.touch, .size-medium.touch",
                    "_id": "co-05",
                    "_skipIfComplete": false
                },
                {
                    "_className": ".size-small.no-touch, .size-medium.no-touch",
                    "_id": "co-10",
                    "_skipIfComplete": false
                }
            ],
            "_id": "course",
            "_force": true
        },
    */

### DIFF
--- a/src/core/js/startController.js
+++ b/src/core/js/startController.js
@@ -1,26 +1,7 @@
 define([
     'coreJS/adapt'
 ], function(Adapt) {
-
-    /*example
-
-        "_start": {
-            "_isEnabled": true,
-            "_screenSize": [
-                {
-                    "_className": ".size-small.touch, .size-medium.touch",
-                    "_id": "co-05"
-                },
-                {
-                    "_className": ".size-small.no-touch, .size-medium.no-touch",
-                    "_id": "co-10"
-                }
-            ],
-            "_id": "course",
-            "_force": true
-        },
-    */
-
+    
     var StartController = function() {
         this.initialize();
     };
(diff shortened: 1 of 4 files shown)


brian-learningpool:
This is a nice feature for core, I was just looking at the extension version of this the other day.  

It would be nice to get some documentation on this somewhere, aside from the JSON example in the code, as the question of how to do this is quite commonly asked in the forums.

taylortom:
+1, with the comments removed (these should stay in example.json I think)

oliverfoster:
agreed

lc-thomasberger:
+1 (update documentation)

tomgreenfield:
+1. Please also add `adapt:start` to the events [documentation](https://github.com/adaptlearning/adapt_framework/wiki/Core-Events#adapt-built-in-events).

oliverfoster:
@moloko  wants to extend with sting pages that hide once completed... might be worth considering how that would fit in with this before we merge

moloko:
Have a `_skipIfCompleted` property that can be added alongside any `_id` property?

oliverfoster:
Yup, then you could line them up in the screenSize array.. perhaps i should rename screensize?

moloko:
yeah, maybe `_startIds` instead?

oliverfoster:
@moloko  better?

moloko:
@oliverfoster looks good to me, I may actually give it a go on the project I'm working on at the moment

src/core/js/startController.js
```diff
@@ -0,0 +1,96 @@
...
+                    } catch(e) {
+                        console.log("startController: cannot find id", item._id);
+                        continue;
+                    }
+                    
+                    if (skipIfComplete) {
+                        if (model.get("_isComplete")) continue;
+                    }
+
+                    if (className || $("html").is(className)) {
```
moloko: shouldn't this be 
```
if (!className || $("html").is(className)) {
```
As it is, if you don't specify `_className` it won't select that startid
?
At the moment, if I don't specify a className is

oliverfoster: Yup, go for it!

moloko: If I make that change it seems to be working for not specifying `_className`; I've run out of time to test it with that in place though

oliverfoster: Yea, I can see any other reason for my hands writing that if+or unless it was meant to have a !

moloko:
+1

taylortom:
+1

moloko:
have updated the [events documentation](adaptlearning/adapt_framework/wiki/Core-Events), plse check what I've put is accurate

oliverfoster:
yup, perfect

oliverfoster:
do we have a section for framework configuration?

moloko:
I'm not sure we do... sounds like a job for @chucklorenz ;-)

also presumably need some way of adding this to the AT since the framework itself doesn't have a properties.schema... is that something you can pick up @taylortom or @brian-learningpool ?

tomgreenfield:
As discussed, `_skipIfComplete` doesn't seem to be working well with Spoor.

oliverfoster:
nope, it's because the completion handlers are now async, spoor restores, start does it's thing, the completion handlers call back and then the contentObject rolls over to complete. nahh... interesting one

oliverfoster:
that's a rather large change, but it basically waits until all the completion cascades and callbacks have fired before letting Adapt initialize. Could prove to be more handy than in this situation only. We should be able to use Adapt.deferUntilCompletionChecked anywhere else it may be needed.

Issue #953, opened 2016-02-16
make core resilient to theme


oliverfoster:
Move basic and essential parts of the theme into core to reduce the requirement on the theme and give the theme more "themeing" responsibility.

Makes the core and theme more stable to theme changes.

oliverfoster:
i had a thought that we could use grunt to take theme variables from the json and compile them into the css output through less variables. that way we could leverage the schema from the theme and the authoring tool editor to provide some quite complex customization - with comparably little... the themes would then be really easily to customize and we could have a less editor for the heavy stuffs?

lc-thomasberger:
would make sense for deveice size variables specified in the [json](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/master/theme.json) and used in [LESS](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/master/less/generic.less#L6-L8)

Would be also possible to use assests (Images) uploaded to the AT in the theme? E.g. select a Image and use it as a background/hero-image for a block/page

oliverfoster:
@sagarkachare @aniketdharia 
We're going to have to start again with this theme stuff.
Nobody is reviewing these PRs so we need to change course.
I've set up a new epic with the intention of starting from the beginning. [new epic](https://github.com/adaptlearning/adapt_framework/issues/1901)
Could I close these issue?
Can you let me know if you want to join in with the new attempt to sort out theming?

oliverfoster:
restarting on https://github.com/adaptlearning/adapt_framework/issues/1901

aniketdharia:
@oliverfoster assigned Sagar to the new Epic.. you can drop him an email and then he can get started.

Cheers!

oliverfoster:
cool, thanks 👍

PR #954, opened 2016-02-16
issue/953: make core resilient to theme

oliverfoster:
* added core templates to core from theme
* added foundation less to core from theme
* added icon fonts
* changed grunt to handle nested less folders in core
* changed grunt to handle fonts in core
* changed grunt to handle no theme folder
* added fix to accessibility-state to core

to test:
1. git clone into branch issue/953, or download the branch
1a. remember to do npm install and adapt install on the framework
2. compile and check working
3. delete src/theme folder
4. recompile and check working (see ghastly 1990's no theme version)

#953


src/core/less/theme/colors.less
```diff
@@ -0,0 +1,145 @@
...
+
+@white: #ffffff;
+@black: #4b4e4f;
+@blue: #7D7D7D;
+@light-grey: #78909c;
+@mid-grey: #546e7a;
+@dark-grey: #37474f;
+@red: #f26c4f;
+@green: #82ca9c;
+
```
tomgreenfield: Suggest dropping the arbitrary list of 'custom' colours in this non-theme and replacing the subsequent named colours in this file with hex codes?

oliverfoster: nice idea - remove a layer of abstract from the basic variables

lc-thomasberger:
@oliverfoster 
when I delete the src/theme folder, it throws an Error: template is not a function. app.js Line 37.
Looks like Handlebars.templates['loading'] is not defined.

oliverfoster:
let me just make sure i can replicate, but the code looks ok... give me an hour or so

oliverfoster:
holding this pr until we have a better understand of how to reduce / expand the core theme to suit everyone's needs.

PR #970, opened 2016-02-23
Resolves #969, corrects loop bug in Grunt getIncludes() helper

### LINKED ISSUE #969: Bug in Grunt helpers.js, in getIncludes()
The two `count` variables here can cause problems with the outer loop.

brian-learningpool:
Two 'count' variables were causing a problem with the loop.

### DIFF
--- a/grunt/helpers.js
+++ b/grunt/helpers.js
@@ -90,26 +90,32 @@ module.exports = function(grunt) {
 
     exports.getIncludes = function(buildIncludes, configData) {
         var dependencies = [];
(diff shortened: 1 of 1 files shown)


taylortom:
+1

bell87:
+1

dennis-learningpool:
+1

moloko:
+1

moloko:
Now we're requiring node 4.3 as a minimum we could presumably start using `let` instead of `var` as a way of avoiding issues like this

brian-learningpool:
Yes, we could.

Issue #939, opened 2016-02-11
Add a task for string substitution in XML files


brian-learningpool:
The `dev`, `build` and `server-build` tasks should allow substitution of strings.  For example, the Spoor plugin could set the imsmanifest.xml, the new TinCan extension could set its XML properties too.

moloko:
done in 11d87ab

PR #971, opened 2016-02-23
Added support for string substitution in XML files

brian-learningpool:
This is required to support upcoming work on the xAPI plugin but will also work with a slightly modified Spoor plugin.  The placeholder format is: @@property.

Right now properties from config.json and course.json are supported.

Squashed commit messages:
- Added replace task to build.js and dev.js
- Added support for passing both course and config to replace call
- Added 'description' to course.json.  This attribute is useful for the SCORM and xAPI extensions.
- Added shim for missing xAPI property. This is to cover edge cases where the xAPI extension may be defined but its configuration may be incomplete.
- Added a warning mesage for when xAPI activityID has not been set.  The course will not confirm to the spec if this has been left empty.

### DIFF
--- a/grunt/config/replace.js
+++ b/grunt/config/replace.js
@@ -1,0 +1,56 @@
+module.exports = function (grunt, options) {
+
+  var courseDir = options.sourcedir + 'course/';
+
+  if (options.outputdir.indexOf('build/') > -1) {
+    courseDir = options.outputdir + 'course/';
+  }
+
+  var configJson = grunt.file.readJSON(courseDir + 'config.json');
+  var defaultLanguage = configJson._defaultLanguage || 'en';
+  var courseJson = grunt.file.readJSON(courseDir + defaultLanguage + '/course.json');
+
+  // Backwards compatibility for courses missing 'description'
+  if (!courseJson.hasOwnProperty('description')) {
+    courseJson.description = '';
+  }
+
+  // A shim for edge cases where xAPI has not been configured.
+  if (!configJson.hasOwnProperty('_xapi')) {
+    configJson._xapi = {
+      'activityID': '',
+      '_isEnabled': false
+    };
+  } else {
+    // xAPI has been enabled, check if the activityID has been set.
+    if (configJson._xapi.hasOwnProperty('activityID') && configJson._xapi.activityID == '') {
+      grunt.log.writeln('WARNING: xAPI activityID has not been set');
+    }
+  }
+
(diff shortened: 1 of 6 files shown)


brian-learningpool:
This resolves #939.

bell87:
+1

grunt/config/replace.js
```diff
@@ -0,0 +1,56 @@
...
+      '_isEnabled': false
+    };
+  } else {
+    // xAPI has been enabled, check if the activityID has been set.
+    if (configJson._xapi.hasOwnProperty('activityID') && configJson._xapi.activityID == '') {
+      grunt.log.writeln('WARNING: xAPI activityID has not been set');
+    }
+  }
+
+  // Combine the course and config JSON so both can be passet to replace.
```
dennis-learningpool: passet -> passed

dennis-learningpool:
+1

moloko:
+1

grunt/config/replace.js
```diff
@@ -0,0 +1,56 @@
...
+  var defaultLanguage = configJson._defaultLanguage || 'en';
+  var courseJson = grunt.file.readJSON(courseDir + defaultLanguage + '/course.json');
+
+  // Backwards compatibility for courses missing 'description'
+  if (!courseJson.hasOwnProperty('description')) {
+    courseJson.description = '';
+  }
+
+  // A shim for edge cases where xAPI has not been configured.
+  if (!configJson.hasOwnProperty('_xapi')) {
```
ryangadams: I'm somewhat uncomfortable with having hacks for extensions in the core grunt tasks. Do we really need this section?

oliverfoster: Ditto. What happens when these variables aren't set?

brian-learningpool: I concur that the shim isn't ideal but it ensures that the XML files don't get created with `@@propertyName` in them.

ryangadams: Could we add a check for @@{someString}@@ in replaced files and complain if any are found?

brian-learningpool: Probably, but I think that would have to be handled in a separate task.

oliverfoster: Templating would have sorted this, string replacement is OK but it ain't got no logic. 
Cool, well, leave it in for the time being then. We can figure it out later. If everyone is OK with that?

ryangadams: Yes, I'm fine with that...

oliverfoster:
+1

Issue #967, opened 2016-02-22
mcq and gmcq model code > model from view


oliverfoster:

moloko:
+1

lc-thomasberger:
+1

oliverfoster:
wohoo!

PR #981, opened 2016-02-25
enhancement-967: questionView model code > questionModel

oliverfoster:
#967 

Work done:
* all model code removed from questionView to questionModel
* compatibility layer for question components which still use their view to define model behaviour

Outcome:
* all questions can now have a proper model with defined behaviour

Next:
* ~~i'll do a pr for mcq and gmcq in this style before the end of the day~~  
https://github.com/adaptlearning/adapt-contrib-gmcq/issues/95  
https://github.com/adaptlearning/adapt-contrib-mcq/issues/120  
* just need to check that this makes sense to everyone (aside from the compatibility layer - which could be a bit confusing)  
* needs a lot of testing with current question components

To test backwards compatibility:
* download enhancement-967 of the framework
* run ``adapt install`` and ``npm install``
* run ``grunt dev``

To test components in the new style:
* download enhancement-967 of the framework
* change ``adapt.json`` ``mcq`` and ``gmcq`` options to ``"enhancement-967"``
* run ``adapt install`` and ``npm install``
* run ``grunt dev``

### DIFF
--- a/src/core/js/models/questionModel.js
+++ b/src/core/js/models/questionModel.js
@@ -14,6 +14,7 @@ define([
             // Extend from the ComponentModel defaults
             return _.extend({
                 '_isQuestionType': true,
+                '_shouldDisplayAttempts': false,
                 '_canShowModelAnswer': true,
                 '_questionWeight': Adapt.config.get("_questionWeight"),
             }, ComponentModel.prototype.defaults);
--- a/src/core/js/models/questionModel.js
+++ b/src/core/js/models/questionModel.js
@@ -5,16 +5,278 @@ define([
 
     var QuestionModel = ComponentModel.extend({
 
+        //////
+        // Setup question types
+        ////
+
+        // Used to set model defaults
         defaults: function() {
+            // Extend from the ComponentModel defaults
             return _.extend({
-                '_isQuestionType': true
+                '_isQuestionType': true,
+                '_canShowModelAnswer': true,
+                '_questionWeight': Adapt.config.get("_questionWeight"),
             }, ComponentModel.prototype.defaults);
         },
 
+        init: function() {
+            this.setupDefaultSettings();
+            this.listenToOnce(Adapt, "adapt:initialize", this.onAdaptInitialize);
+        },
+
+        // Calls default methods to setup on questions
+        setupDefaultSettings: function() {
+            // Not sure this is needed anymore, keeping to maintain API
+            this.setupWeightSettings();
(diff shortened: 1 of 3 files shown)


src/core/js/models/questionModel.js
```diff
@@ -5,16 +5,278 @@ define([
...
+
+        // Used to set model defaults
         defaults: function() {
+            // Extend from the ComponentModel defaults
             return _.extend({
-                '_isQuestionType': true
+                '_isQuestionType': true,
+                '_canShowModelAnswer': true,
+                '_questionWeight': Adapt.config.get("_questionWeight"),
             }, ComponentModel.prototype.defaults);
```
tomgreenfield: Maybe add `_shouldDisplayAttempts` here?

tomgreenfield:
+1

chucklorenz:
Tutor uses Notify to display feedback. It displays the component title at the top. It would be nice to be able to replace this with text of the author's choosing.

If the feedback model could be changed to something like this:
```
"_feedback":{
        "_correct": {
            "title": "That's right! Nice work.",
            "final": "Erie, Superior, and Michigan are all Great Lakes."
        },
        "_incorrect": {
            "title": "Unfortunately that is incorrect.",
            "notFinal": "Why not try again.",
            "final": "Erie, Superior, and Michigan are all Great Lakes, but Ricky Lake is not."
        },
        "_partlyCorrect": {
            "title": "That's partly correct.",
            "notFinal": "At least one of your choices is incorrect. Why not try again.",
            "final": "Erie, Superior, and Michigan are all Great Lakes, but Ricky Lake is not."
        }
    },
```
Then each of the three functions, `SetupCorrectFeedback`, `SetupPartlyCorrectFeedback`, and  `SetupIncorrectFeedback`, could incorporate a line of code like:
`feedbackTitle: this.model.get("_feedback")._correct.title || this.model.get('title'),`

Granted having “final” as the primary “correct” value appears strange at first. But the parallelism with “_incorrect” and “_partlyCorrect” is probably better than _correct.correct. And maybe we go further and include _correct.notFinal. Who knows when someone might find a use case for it.

oliverfoster:
[2016-04-25-Moving_forward.pptx](https://github.com/adaptlearning/adapt_framework/files/234349/2016-04-25-Moving_forward.pptx)

moloko:
+1

lc-thomasberger:
+1

brian-learningpool:
+1

PR #987, opened 2016-03-01
Added layout sort override

guywillis:
Added a snippet of code which allows you to override the left/right sort order of components by applying '_sortComponents': false to the container block.

Ideally used for when you have an MCQ component left aligned and Graphic component right aligned on desktop but want the Graphic component to appear above the MCQ in mobile view. 

All component JSON is unchanged, only amendment is to define the right aligned Graphic component above the left aligned MCQ component in the components.json.

### DIFF
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -192,7 +192,7 @@ define(function (require) {
             }
 
             if (this.get('_type') == 'block' && childrenCollection.length == 2
-                && childrenCollection.models[0].get('_layout') !== 'left') {
+                && childrenCollection.models[0].get('_layout') !== 'left' && this.get('_sortComponents') !== false) {
                 // Components may have a 'left' or 'right' _layout,
                 // so ensure they appear in the correct order
                 // Re-order component models to correct it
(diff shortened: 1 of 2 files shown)


oliverfoster:
+1

oliverfoster:
dom order is very important for this usecase, the only way around this for both the authoring tool and the framework is to have a _layoutVertical: "top" / "bottom" as well as the _layout: "left, right, full"

oliverfoster:
How do we maintain block order inside articles in the authoring tool?

brian-learningpool:
There's a `_sortOrder` attribute on each block.

brian-learningpool:
OK, if there's no way around it.  I'd prefer not to complicate things with a `_layoutVertical` attribute.   

@guywillis, can you just add `_sortComponents: true` to the BlockModel defaults?

guywillis:
Thanks for adding `_sortComponents: true` to the BlockModel defaults, @oliverfoster.

tomgreenfield:
+1

hrajotia:
+1

brian-learningpool:
+1

PR #988, opened 2016-03-01
Resolves #984, removes dependency on 'build' folder existing

### LINKED ISSUE #984: Clean install grunt build cannot find config.json
Fresh install doesn't seem to be working:

```
$ grunt build
Loading "Gruntfile.js" tasks...ERROR
>> Error: Unable to read "/Users/Si1kIfY/dev/ucopea/build/course/config.json" file (Error code: ENOENT).
```

I'll try to remove / re-install and report back

brian-learningpool:
Also using path.join() now rather than string concatenation for folder names.

### DIFF
--- a/grunt/config/replace.js
+++ b/grunt/config/replace.js
@@ -1,19 +1,46 @@
+var path = require('path');
+var fs = require('fs');
+
 module.exports = function (grunt, options) {
 
-  var courseDir = options.sourcedir + 'course/';
+  var courseDir = path.join(options.sourcedir, 'course');
 
-  if (options.outputdir.indexOf('build/') > -1) {
-    courseDir = options.outputdir + 'course/';
+  if (options.outputdir.indexOf('build') > -1) {
+    courseDir = path.join(options.outputdir, 'course');
   }
 
-  var configJson = grunt.file.readJSON(courseDir + 'config.json');
+  var pathToConfig = path.join(courseDir, 'config.json')
+
+  try {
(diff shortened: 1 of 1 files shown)


oliverfoster:
+1

moloko:
+1 fixes the original issue (tested on OS X 10.11.3) - but I do get the following in the output, anything to be concerned about?
```
Running "replace:dist" (replace) task
>> Unable to match 1 pattern, remember for simple matches (String) we are using the prefix @@ for replacement lookup.
>> 0 replacements in 1 file.
```

grunt/config/replace.js
```diff
@@ -1,19 +1,46 @@
...
+            flatten: true,
+            src: [options.outputdir + '*.xml'],
+            dest: options.outputdir
+          }
+        ]
+      }
+    }
   }
+  
+  var configJson = grunt.file.readJSON(pathToConfig);
```
ryangadams: Would it be better to wrap this line in a try catch and set configJson to an empty object on error? Is is possible to have a circumstance where a replacement task should still be executed with the data in courseJson even if configJson is missing?

oliverfoster: or do files:[] ? i think the replace task still needs some configuration due to the way the tasks are queued.

brian-learningpool:
@moloko, nothing to be concerned about but I have added `silent: true` now to suppress the logging here.

oliverfoster:
+1

PR #998, opened 2016-03-14
updates to allow for better validation

moloko:


hrajotia:
+1

brian-learningpool:
+1

oliverfoster:
+1 (is you put a &nbsp; in the <title></title> tags you can hit that one also :D )

moloko:
I think I'll leave it - the title won't be blank when the course actually runs.

PR #1000, opened 2016-03-15
more segregation of copy & watch task

hrajotia:


oliverfoster:
+1

grunt/config/watch.js
```diff
@@ -23,7 +23,14 @@ module.exports = {
...
             '!<%= sourcedir %>theme/theme.js',
-            '!<%= sourcedir %>templates/templates.js'
+            '!<%= sourcedir %>templates/templates.js',
+            '!<%= sourcedir %>core/js/scriptLoader.js',
+            '!<%= sourcedir %>core/js/libraries/require.js',
+            '!<%= sourcedir %>core/js/libraries/modernizr.js',
+            '!<%= sourcedir %>core/js/libraries/json2.js',
+            '!<%= sourcedir %>core/js/libraries/consoles.js',
+            '!<%= sourcedir %>core/js/libraries/jquery.js',
+            '!<%= sourcedir %>core/js/libraries/jquery.v2.js'
```
taylortom: Do we want to use a glob here? (i.e. like `libraries/*`)

grunt/config/watch.js
```diff
@@ -62,5 +69,24 @@ module.exports = {
...
+        tasks: ['copy:scriptLoader']
+    },
+    libraries: {
+        files: [
+            '<%= sourcedir %>core/js/libraries/require.js',
+            '<%= sourcedir %>core/js/libraries/modernizr.js',
+            '<%= sourcedir %>core/js/libraries/json2.js',
+            '<%= sourcedir %>core/js/libraries/consoles.js',
+            '<%= sourcedir %>core/js/libraries/jquery.js',
+            '<%= sourcedir %>core/js/libraries/jquery.v2.js'
```
taylortom: glob?

grunt/config/watch.js
hrajotia: We need to make sure that after making change to these specific files should not trigger a `javascript:dev` task because we are just intended to copy them directly into the build folder.

I haven't used glob here because there are few other libraries available (e.g. `backbone.js`, etc..) which needs to be watched for there changes as they becomes a part of `adapt.min.js`.

grunt/config/watch.js
hrajotia: only these files could go to the `build\libraries` directory.

oliverfoster:
+1

moloko:
+1

chetan-hajare:
+1

PR #1013, opened 2016-04-07
Add _isLocked to AdaptModel

tomgreenfield:
Implementation as per #905.

PR #1018, opened 2016-04-14
Resolves #1017, checks _isLocked in router

### LINKED ISSUE #1017: Update the router to consider `_isLocked` attributes
With regards to the new `_isLocked` property, although the BoxMenu and the framework core code have been updated, it's still going to be possible to navigate to another page/menu via the URL.

I think it would be worthwhile to add a check in the router here.

brian-learningpool:
This prevents routing to a locked page/menu via the URL, or menu link which has not yet been updated to support _isLocked.

Also, Adapt.findById() has been updated to return a warning when the _id value passed in does not exist in the mapping.

### DIFF
--- a/src/core/js/router.js
+++ b/src/core/js/router.js
@@ -106,10 +106,11 @@ define([
             var type = '';
             
             if (!currentModel) {
+                Adapt.router.set('_canNavigate', true, {pluginName: "adapt"});
                 return;
-            } else {
-                type = currentModel.get('_type');
             }
+
+            type = currentModel.get('_type');
 
             switch (type) {
                 case 'page':
@@ -117,7 +118,11 @@ define([
                     if (currentModel.get('_isLocked') && Adapt.config.get('_forceRouteLocking')) {
                         console.log('Unable to navigate to locked id: ' + id);
                         Adapt.router.set('_canNavigate', true, {pluginName: "adapt"});
-                        return Backbone.history.history.back();
+                        if (Adapt.location._previousId === undefined) {
+                            return this.navigate("#/", {trigger:true, replace:true});
+                        } else {
+                            return Backbone.history.history.back();
+                        }
                     } else {
                         this.showLoading();
                         this.removeViews();
--- a/src/core/js/router.js
+++ b/src/core/js/router.js
@@ -94,7 +94,7 @@ define([
             this.setContentObjectToVisited(Adapt.course);
             this.updateLocation('course');
             Adapt.once('menuView:ready', function() {
-                //allow navigation
+                // Allow navigation
                 Adapt.router.set('_canNavigate', true, {pluginName: "adapt"});
             });
             Adapt.trigger('router:menu', Adapt.course);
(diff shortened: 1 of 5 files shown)


brian-learningpool:
This PR should go in once #1013 is merged.

moloko:
I feel very strongly that this is something that needs to be configurable (a bit like the `_force` config option for start pages) - with it switched off by default.

Being able to bypass the locking via URL is indispensable for testing/development purposes.

If this were merged in as-is it would add quite a lot of faff to development/testing time for us (having to constantly switch locking on/off) - whilst adding very little in the way of benefit since 99% of our courses are presented to the user in a popup window with no address bar, leaving them unable to navigate via the URL anyway.

brian-learningpool:
@moloko, I've added a `_forceRouteLocking` attribute to the config model, defaulted to `false`.

src/core/js/router.js
```diff
@@ -103,37 +103,51 @@ define([
         handleId: function(id) {
 
             var currentModel = Adapt.findById(id);
+            var type = '';
+            
+            if (!currentModel) {
+                return;
+            } else {
```
oliverfoster: Unnecessary else block

src/core/js/router.js
```diff
@@ -103,37 +103,51 @@ define([
...
-                        });
-                        Adapt.trigger('router:page', currentModel);
-                        this.$wrapper.append(new PageView({model:currentModel}).$el);
+            switch (type) {
+                case 'page':
+                case 'menu':
+                    if (currentModel.get('_isLocked') && Adapt.config.get('_forceRouteLocking')) {
+                        console.log('Unable to navigate to locked id: ' + id);
+                        Adapt.router.set('_canNavigate', true, {pluginName: "adapt"});
+                        return Backbone.history.history.back();
```
hrajotia: How about also checking for availability of previous route and if its not been there then navigate to the course?
```javascript
if(Adapt.location._previousId === undefined) {
    return this.navigate("#/", {trigger:true, replace:true});
} else {
    return Backbone.history.history.back();
}
```

hrajotia:
:+1:

moloko:
I've just tried this on the current version of the framework (together with @tomgreenfield's changes from #1013) and, when I run the course, it triggers the bookmarking dialog together with the following message in the console:
```
Adapt.findById() unable to find collection type for id: undefined
```
Clicking 'Yes' in the bookmarking dialog takes me to http://localhost:9001/#/id/undefined

If I disable spoor via config.json, it does not do this.

Equally it doesn't do this if spoor is enabled but I'm using the current (master) version of adapt.js - so it's likely to be the amend in there not playing nicely with the offline_API_wrapper.js in spoor - which is returning `undefined` for the lesson_location

I suspect a bit of defensive programming needs to be added to `Adapt.findById` so that it will return if if finds that the `id` argument is set to anything like `undefined`, `"undefined"`, `null` or `"null"`

brian-learningpool:
Hi @moloko, thanks for testing, however after looking at it this is more a problem with the Bookmarking extension itself storing the string "undefined" in the `restoredLocationID`, as opposed to an empty string or a null value.  I don't think we should be doing string comparisons to check for null/undefined values.

See here:
https://github.com/adaptlearning/adapt-contrib-bookmarking/blob/master/js/adapt-contrib-bookmarking.js#L39

Also, it appears somebody at least caught the error thrown by core here:
https://github.com/adaptlearning/adapt-contrib-bookmarking/blob/master/js/adapt-contrib-bookmarking.js#L53

A more correct solution would be to also check the model returned by `Adapt.findById()` is not null or undefined.

moloko:
fair point, checking for the string `"undefined"` is probably out of scope for this code (I'm just used to the days when IE would use `"undefined"` by default so got into the habit of checking for this as well) - but I still think it could do with exiting if `id` is `null` or `undefined`

brian-learningpool:
I think if we're to add parameter validation like this we should probably do it for all functions and create a separate ticket for that.  The function already returns early if the collection type can't be determined, as would happen with an invalid id/index.

moloko:
+1

hrajotia:
@brian-learningpool - Please take a look at https://github.com/adaptlearning/adapt_framework/blob/issue/1017/src/core/js/router.js#L176.

This check would become useless as `Adapt.findById` might not be throwing any error now but just returning `undefined` instead.

What do you think?

brian-learningpool:
Hi @hrajotia, I think the second link is fine as the `continue` is exiting the `for` loop.  I've updated the code you highlighted anyway as the empty `catch` in the first example was bugging me.

oliverfoster:
questionView L483

src/core/js/views/questionView.js
```diff
@@ -481,7 +481,7 @@ define(function(require) {
             try {
```
oliverfoster: can you remove the try catch also?

hrajotia:
+1

dennis-learningpool:
+1

moloko:
+1

PR #1024, opened 2016-04-19
default less-variables added

### LINKED ISSUE #1006: define @focus-outline in base.less
@focus outline is not defined in the core theme so it will break compilation when this variable is not in the theme.
to fix this add:
@focus-outline: 3px solid orange;
https://github.com/adaptlearning/adapt_framework/blob/master/src/core/less/base.less#L25

lc-thomasberger:
Fixes #1006


moloko:
+1

brian-learningpool:
+1

oliverfoster:
+1

PR #1027, opened 2016-04-21
remove one of the calls to checkLocking to avoid a warning appearing in the console

### LINKED ISSUE #1025: Warning being generated in console when using 'custom' locking
All still works fine, just getting a warning like:
> `AdaptModel.shouldLock: unknown _lockedBy ID "co-15" found on co-10`

Need to _.defer the call to `checkLocking` I suspect

moloko:
also refactor code so that we don't have to hide the call to `checkLocking` inside `checkCompletionStatus`

fixes #1025

### DIFF
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -41,9 +41,11 @@ define(function (require) {
                 this.setupChildListeners();
             }
 
-            this.checkLocking();
             this.init();
+            
             _.defer(_.bind(function() {
+                this.checkLocking();
+                
                 if (this._children) {
                     this.checkCompletionStatus();
                     this.checkInteractionCompletionStatus();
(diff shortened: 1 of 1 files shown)


moloko:
@hrajotia could you take a look at this? 

I understand from @tomgreenfield that this additional call to `checkLocking` was added at your request - but the problem is that if you do it before `init` then it's too early - the other models won't necessarily have been created yet...

So in fact having discussed it I think the call to `checkLocking` needs to be dropped from setupModel because if it's called before `init` it just generates an error; if it's called after (as I've set up in this PR) it's just a redundant extra call as it's already being called by `checkCompletionStatus`*

*which is not the ideal location for this, this bit needs refactoring

moloko:
as a reminder, here's your original comment on this https://github.com/adaptlearning/adapt_framework/pull/1013#issuecomment-209375523

hrajotia:
+1

tomgreenfield:
+1

lc-thomasberger:
+1

Issue #1079, opened 2016-05-26
core: router's navigateToHomeRoute function causes the page to scroll to the top


moloko:
...prior to navigating back to the home route

This is particularly noticeable if you're using an extension like [pageIncompletePrompt](https://github.com/cgkineo/adapt-pageIncompletePrompt) or [keepScrollPosition](https://github.com/cgkineo/adapt-keepScrollPosition) - it stops the latter from working since the `window.scrollY` will always be 0 if you've used the back button

PR #1080, opened 2016-05-26
set navigateToHomeRoute to navigate to '#/' (instead of '#') to stop …

moloko:
…the page from scrolling to the top immediately prior to navigation (see #1079)

If you want to test this you'll need to change the back button to a home button by changing theme/templates/navigation.hbs so that the data-event of the `button` with class `navigation-back-button` is `homeButton` (instead of `backButton`).


LukaszGrela:
+1

brian-learningpool:
+1

hrajotia:
+1

PR #1092, opened 2016-06-07
jquery.a11y tabindex fix

tomgreenfield:
Remove tabindex from hideable elements on popdown

### DIFF
--- a/src/core/js/libraries/jquery.a11y.js
+++ b/src/core/js/libraries/jquery.a11y.js
@@ -1147,10 +1147,14 @@
 
                 if (previousTabIndex === -1) {
                     //hide element from screen reader
-                    $item.attr('aria-hidden', true).addClass("aria-hidden");
-                } else {
-                    //show element to screen reader
-                    $item.removeAttr('aria-hidden').removeClass("aria-hidden");
+                    return $item.attr('aria-hidden', true).addClass("aria-hidden");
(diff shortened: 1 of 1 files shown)


oliverfoster:
+1

src/core/js/libraries/jquery.a11y.js
```diff
@@ -1147,10 +1147,14 @@
...
-                    //show element to screen reader
-                    $item.removeAttr('aria-hidden').removeClass("aria-hidden");
+                    return $item.attr('aria-hidden', true).addClass("aria-hidden");
+                }
+
+                //show element to screen reader
+                $item.removeAttr('aria-hidden').removeClass("aria-hidden");
+                
+                if ($item.is(domSelectors.hideableElements)) {
+                    $item.removeAttr("tabindex");
```
oliverfoster: CODE NOTES

```.a11y-hideable```
Some elements, which were never explicitly in the tab order, need to be removed from the tab order when a popup occurs. Label tags associated with input tags are an example of such elements. When the popdown occurs their tabindex="0" should be removed as it wouldn't have originally existed and will cause a double tab if left.

iFrame containers for youtube videos will also have this same problem.

moloko:
+1

brian-learningpool:
+1

PR #1098, opened 2016-06-13
Resolves #1096, device width for iOS devices

### LINKED ISSUE #1096: Consider using screen.width for iOS device calculations
While testing a course in an LMS which used frames, I noticed that the `Adapt.device.screenWidth` was being set incorrectly -- it was detected as large rather than small.

Android devices seem to work as expected, so I'm wondering if we should tweak the code to make an exception for iOS so that expected sizing class gets correctly added?

brian-learningpool:
When content is rendered inside frames, the wrong size classes were being applied for iOS devices.  Switching to use screen.width as part of the calculation resolves this.

Android devices already behave as expected.

### DIFF
--- a/src/core/js/device.js
+++ b/src/core/js/device.js
@@ -2,41 +2,44 @@ define(function(require) {
 
     var Adapt = require('coreJS/adapt');
     var Bowser = require('coreJS/libraries/bowser');
-
-    Adapt.device = {};
-
     var $window = $(window);
 
-    // Check whether device is touch enabled
-    Adapt.device.touch = Modernizr.touch;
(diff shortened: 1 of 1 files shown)


bell87:
+1

barry-learningpool:
+1

lc-thomasberger:
+1

moloko:
+1

tomgreenfield:
+1

taylortom:
+1

andrewhosgood:
https://github.com/adaptlearning/adapt_framework/issues/1096#issuecomment-235615130

PR #1102, opened 2016-06-16
Issue 1030

bell87:
Added a check for `_isAvailable` on each model within the `findDescendants()` function which fixes progress indicators / completion of course.

This will mean that if an article isn't "available" then the descendants of the article won't be available either.


barry-learningpool:
+1

taylortom:
+1

oliverfoster:
non of the other core model api traversal functions execute this way

brian-learningpool:
@bell87, if this resolves the issue I think this would be cleaner if the code you've changed was moved directly to the Page Level Progress extension.

oliverfoster:
seconded, filterAvailableChildren doesn't act upon the model. it's more like an extra utility function. well implemented though. 

a few notes:
array.map and array.indexOf aren't ie8 compatible, so you'll need to use the underscore versions. 
_.each is really inefficient, could you swap it for a for loop?

Issue #1114, opened 2016-06-23
Question components no longer show the marking on submit


RobertPeek:
When clicking the submit button on a question component the marking icons used to automatically show on each option. However after downloading and building the latest framework v2.0.10 this doesn't happen anymore. Was this functionality changed or is this a bug?

moloko:
You're quite right @RobertPeek this is a bug

Confirmed that this was still working properly with Framework v2.0.9

moloko:
done in https://github.com/adaptlearning/adapt_framework/commit/bb2396f7837b07c3bead4fa868aa263b938cec52

PR #1115, opened 2016-06-23
bugfix: #1114

oliverfoster:
standard question view (new version) functions were not being called by the compatibility layer.

the checking statement, line 484, was checking if the required function on the current view prototype was the same as the compatibility layer function - which it always will have been because the compatibility extension view was always an extension of the question view.

i have stopped the compatibility extension including the question view prototype functions when the comparison takes place so that line 484 will not be comparing "setQuestionAsSubmitted" and "checkQuestionCompletion" to itself.

### DIFF
--- a/src/core/js/views/questionView.js
+++ b/src/core/js/views/questionView.js
@@ -342,7 +342,7 @@ define([
     * Remove this section in when all components use questionModel and there is no need to have model behaviour in the questionView
     */
 
-    var viewOnlyCompatibleQuestionView = QuestionView.extend({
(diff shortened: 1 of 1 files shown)


brian-learningpool:
+1, I've tested this on the latest `master` branch and it resolves the issue for me.

RobertPeek:
+1, tested in the latest master branch too and all working as expected now

moloko:
+2 confirmed fixed for me too, tested in Firefox and IE11

Issue #1073, opened 2016-05-23
Assessment extension - Hide the feedback button?


hmaffin:
In the Authoring Tool even if in Assessment settings I have show feedback set to false all it does is grey out the button.
Can we have it so that the button is not shown, and the Submit/Reset button is shown full width?

Something to do with _canShowFeedback for the component?

LukaszGrela:
Yes, this flag could mark component with `no-feedback` class which then could be used to hide button.

```
.no-feedback .buttons-feedback {
    display:none;
}
```

Ofcourse you can do this right now with CSS only, if you want to hide the feedback for particular component just put a style with component id eg.

```
.c-005 .buttons-feedback {
    display:none;
}
```
in similar fashion the `button.buttons-action` needs to be modified to take full width.

hmaffin:
AH that's excellent Lukasz thank you for the in-depth answer!

hmaffin:
The CSS route is probably not ideal. It involves a lot of hand-adding classes to the theme-extras.less and has to be done for every component. I think perhaps more flexibility with the buttons might be useful. 

I am wondering if a new component might be a better idea - i'm trying to make one myself but not getting very far! - that offers a choice rather than a marked, feedbacked question.

paul-mediakitchen:
Hi Helen

I used the following to hide the feedback button and make the other button 100% 

`
.assessment .buttons button.buttons-feedback {
    display:none;
}

.assessment .buttons .buttons-marking-icon {
    position: absolute;
    left: 96%;
}

.assessment .buttons button {
    width: 100%;
    float: left;
}`

I also adjusted the position of the tick/cross

chetan-hajare:
PR - https://github.com/adaptlearning/adapt-contrib-vanilla/pull/193

chetan-hajare:
Raised 2 PRs: https://github.com/adaptlearning/adapt-contrib-vanilla/pull/194 and https://github.com/adaptlearning/adapt_framework/pull/1145

PR #1145, opened 2016-07-18
modified to resolve issue #1073

chetan-hajare:


moloko:
This is working great and I think we now have the right structure - my only remaining issue is with the name `disableFeedbackButton` - because that's not really what this code is doing - it's the theme that actually does that.

I'm thinking either `addNoFeedbackClasses` - or maybe better to have a more generic name `addClasses` so that this could be extended by plugins... @brian-learningpool any thoughts?

brian-learningpool:
I think naming the function `checkFeedbackState()`instead might work better.   This would be more in line with the existing `checkResetSubmittedState()`, which does something similar.

Also, mind the spacing.

moloko:
plus maybe an explanatory comment above line 115 explaining that the purpose is to allow the feedback button to be hidden and the submit button to stretch to fill the space - and that the CSS for this is in the theme

oliverfoster:
+1

brian-learningpool:
@chetan-hajare, are you ok to rename the function to `checkFeedbackState()`?  

It'd be great to get this into the next release.

oliverfoster:
@chetan-hajare this one went in on a separate pr just after you changed it this morning (https://github.com/adaptlearning/adapt_framework/pull/1171), sorry, thanks for doing the hard work. 
could you close this?

chetan-hajare:
Sure, and sorry I made the changes late.

oliverfoster:
nah don't worry, learningpool needed the change quicker than we'd normally expect. it's not your fault. 19 hours is a reasonable time between a request and an update. we've often have changes that take weeks to implement.

brian-learningpool:
Many thanks for your PR, @chetan-hajare.  This is a really good enhancement.

PR #1150, opened 2016-07-20
decode languagefiles on import

lc-thomasberger:
try to auto-detect language files encoding on import.
decode with detected character-encoding, fallback to utf-8

### DIFF
--- a/grunt/tasks/translate/loadLanguageFiles.js
+++ b/grunt/tasks/translate/loadLanguageFiles.js
@@ -1,6 +1,8 @@
 var csv = require("csv");
 var path = require("path");
 var async = require("async");
+var jschardet = require("jschardet");
+var iconv = require("iconv-lite");
 
 module.exports = function (grunt) {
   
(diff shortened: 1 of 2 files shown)


grunt/tasks/translate/loadLanguageFiles.js
```diff
@@ -54,8 +54,10 @@ module.exports = function (grunt) {
 
         if (iconv.encodingExists(detected.encoding)) {
           fileContent = iconv.decode(fileBuffer, detected.encoding);
+          grunt.log.debug(filename+' - encoding detected: '+detected.encoding);
         } else {
           fileContent = iconv.decode(fileBuffer, 'utf8');
+          grunt.log.debug(filename+' - encoding not detected, used utf-8 instead');
```
dancgray: I think this should be grunt.log.writeln so that it writes the log even when not in debug mode.

lc-thomasberger: This should just give useful information for debugging.

Issue #1152, opened 2016-07-21
Browser "Back" button issue with Start Controller


brian-learningpool:
If the Start Controller is being used, on initial load the browser is correctly routed to the specific page, however it's possible to hit the browser's "Back" button and be taken to the menu -- which the user probably shouldn't be able to see.

Perhaps if the `_forceRouteLocking` is set to `true` the user should be prevented from viewing the menu altogether?

brian-learningpool:
Raised by @paul-mediakitchen.

brian-learningpool:
PR has been merged into `master`.

PR #1153, opened 2016-07-21
Corrects #1152 with start controller functionality

brian-learningpool:
If the start controler is enabled, and '_force' is set to true, it should not be possible to route to the menu.

### DIFF
--- a/src/core/js/router.js
+++ b/src/core/js/router.js
@@ -92,7 +92,7 @@ define([
                 // Do not allow access to the menu when the start controller is enabled.
                 var startController = Adapt.course.get('_start');
                 
-                if (startController._isEnabled == true && startController._force == true) {
+                if (startController._isEnabled == true && startController._isMenuDisabled == true) {
                     return;
                 }
             }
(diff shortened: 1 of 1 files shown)


moloko:
can I just check this wouldn't cause problems in a scenario where you're using _start to introduce an 'intro/sting' page before the menu?

oliverfoster:
i think we need a isMenuDisabled property 

_force is meant to always start on the specified pages, it's not really meant to stop the menu.

brian-learningpool:
@moloko, if you set `_force` to `false` in that case it will be possible to route to the menu. 

I agree with @oliverfoster though in that it might be better served with a new property to hide the menu.

oliverfoster:
+1

moloko:
+1
`_isMenuDisabled` will also need adding to [the documentation](/adapt_framework/wiki/Content-starts-with-course.json)

brian-learningpool:
Documentation is now up to date.

dennis-learningpool:
+1

Issue #1163, opened 2016-07-25
move theme fonts from theme to core


oliverfoster:
- vanilla is standard and should be in core

PR #1165, opened 2016-07-25
issue-1163: moved vanilla font into core

oliverfoster:
https://github.com/adaptlearning/adapt_framework/issues/1163


lc-thomasberger:
+1

tomgreenfield:
There is probably now a good chance to drop a couple of the font formats before adding too much to core. According to [caniuse](http://caniuse.com/#feat=woff), WOFF is supported on:
* Chrome 5+
* Firefox 3.6+
* Internet Explorer 9+
* Edge
* Safari / iOS Safari 5.1+
* Chrome for Android (which runs on Android 4.1+)

We still need to keep the EOT version for Internet Explorer 8 but it looks safe to drop the **SVG** and **TTF** formats. In the future, we could discuss adding WOFF2 when it becomes more widely supported.

guywillis:
+1

tomgreenfield:
+1

moloko:
+1

Issue #1187, opened 2016-07-29
matching partial feedback not working


oliverfoster:
mcq partial feedback fine
core framework issue

PR #1188, opened 2016-07-29
issue-1187: matching partial feedback fix

oliverfoster:
Question View:

1. setupFeedback https://github.com/adaptlearning/adapt_framework/blob/cd2b27ba2c5df4e6f4c60ba15cacd4b769d9cfa2/src/core/js/views/questionView.js#L429-L447

2. markQuestion https://github.com/adaptlearning/adapt_framework/blob/cd2b27ba2c5df4e6f4c60ba15cacd4b769d9cfa2/src/core/js/views/questionView.js#L397-L410

Both functions should be called through the view always in compatibility mode as some of their sub functions can live on the view only

#1187


andrewhosgood:
+2 Tested and working

brian-learningpool:
+1

tomgreenfield:
+1

PR #1204, opened 2016-08-16
Rework to device.js

brian-learningpool:
- Abstracted code into a getScreenWidth() function
- Corrected issue with width reported for iOS devices in landscape by adding getAppleScreenWidth() function

### DIFF
--- a/src/core/js/device.js
+++ b/src/core/js/device.js
@@ -41,7 +41,7 @@ define(function(require) {
     function getScreenWidth() {
         return isAppleDevice()
             ? getAppleScreenWidth()
-            : window.innerWidth || $window.width()
+            : window.innerWidth || $window.width();
     }
 
     var onWindowResize = _.debounce(function onScreenSizeChanged() {
--- a/src/core/js/device.js
+++ b/src/core/js/device.js
@@ -6,7 +6,7 @@ define(function(require) {
 
     Adapt.device = {
         touch: Modernizr.touch,
-        screenWidth: isAppleDevice() ? screen.width : window.innerWidth || $window.width()
+        screenWidth: getScreenWidth()
     };
 
     Adapt.once('app:dataReady', function() {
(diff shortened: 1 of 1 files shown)


brian-learningpool:
This resolves the reporting of the width in iOS devices as discussed at #1096.  In iOS the orientation is important in calculating the `screenWidth`, as when it's in portrait `screen.width` works fine, but when it's in landscape `screen.height` refers to the 'width'.  This solution should work both on a course embedded within an `iframe` and without.

@andrewhosgood, can you please test this resolves your issue?

andrewhosgood:
+1 Tested but not within an iframe

oliverfoster:
need to see if the iframe will scale properly as well, otherwise we end up in a weirder situation where the content width doesn't match the javascript measurement.

oliverfoster:
+1 this needs to go in to fix the immediate issue, the iframe one still needs investigation

src/core/js/device.js
```diff
@@ -69,6 +73,10 @@ define(function(require) {
         return /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream;
     }
 
+    function getAppleScreenWidth() {
+        return (Math.abs(window.orientation) === 90) ? screen.height : screen.width;
```
oliverfoster: do we need 180 also?

andrewhosgood: You mean 270? It appeared to work in all orientations on the iPad I was testing on.

andrewhosgood: Actually ignore me, I missed the `Math.abs`!

oliverfoster: yea, that's exactly what i was meaning

oliverfoster: it goes -90 90 ?

oliverfoster: http://www.williammalone.com/articles/html5-javascript-ios-orientation/ yup, s'all good

src/core/js/device.js
```diff
@@ -38,11 +38,15 @@ define(function(require) {
         return screenSize;
     }
 
+    function getScreenWidth() {
+        return isAppleDevice()
+            ? getAppleScreenWidth()
+            : window.innerWidth || $window.width()
```
tomgreenfield: The ending semi-colon has gone astray here. Feel free to ignore if it doesn't bother anyone!

brian-learningpool: Ah, nuts! Of course bothers me now...

tomgreenfield:
+1

oliverfoster:
+1

Issue #1219, opened 2016-09-01
‘_createLookupTables’ Grunt task needs to support arrays of strings


tomgreenfield:
This [line](https://github.com/adaptlearning/adapt_framework/blob/feature/ml/grunt/tasks/translate/createLookupTable.js#L52) can cause an error when `description.items` is `undefined`. We need to handle when there are arrays in schemas with no nested items e.g. [`_lockedBy`](https://github.com/adaptlearning/adapt_authoring/blob/master/plugins/content/contentobject/model.schema#L67-L75).

A quick way would be to break early:
```javascript
case "array":
  var items = description.items;

  if (!items) break;
```
But we also need to handle _translatable_ arrays of strings.

lc-thomasberger:
also _answers array of textInput component is not exported 
[schema](https://github.com/adaptlearning/adapt-contrib-textInput/blob/master/properties.schema#L41-L53)
[json](https://github.com/adaptlearning/adapt-contrib-textInput/blob/master/example.json#L140-L144)

oliverfoster:
agreed to leave "List" unsupported as translatable at the moment

PR #1226, opened 2016-09-06
issue #1219 

lc-thomasberger:
Currently array's without items attribute are not supported. Example for this is `"inputType": "List"` used for lockedBy attribute.

Handle this case by skippinkg to next item.
If list of array needs to be supported, use array + items definition instead. E.g. textinput `_items._answers`

### DIFF
--- a/grunt/tasks/translate/createLookupTable.js
+++ b/grunt/tasks/translate/createLookupTable.js
@@ -49,7 +49,10 @@ module.exports = function (grunt) {
             break;
           
           case "array":
-            if (description.items.type === "object") {
+            if (!description.hasOwnProperty('items')) {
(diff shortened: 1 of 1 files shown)


grunt/tasks/translate/createLookupTable.js
```diff
@@ -49,7 +49,10 @@ module.exports = function (grunt) {
             break;
           
           case "array":
-            if (description.items.type === "object") {
+            if (!description.hasOwnProperty('items')) {
+              // handles "inputType": "List" edge-case
+              break;
+            } else if (description.items.type === "object") {
```
tomgreenfield: Doesn't require an `else` here since we have a `break`.

Issue #1228, opened 2016-09-06
slow javascript:compile and javascript:dev


oliverfoster:
for me:

dev takes 3.6seconds
compile takes 14.1seconds

oliverfoster:
missed bowser

oliverfoster:
^ moved libraries out of standard plugins - this is breaking as it'll require the framework to be at least with #1232, those three will not be backward compatible.

oliverfoster:
with these changes there is a vast improvement in compile time

oliverfoster:
all done here, @lc-thomasberger has taken care of matching, so hopefully... finger's crossed, we'll get an even faster compile time.... exciting!!! :D

PR #1229, opened 2016-09-06
review-issue-1228: speed up javascript task

oliverfoster:
https://github.com/adaptlearning/adapt_framework/issues/1228

* made all libraries in src/core/libraries load before adapt.min.js
  - they load on an http request each rather than being compiled (old school style) - seems like a reasonable trade-off now that network speeds are pretty good

* removed unnecessary code references

* removed build license comments in favour of dropping output compression (roughly the same byte size but reduces the compile time by 4 seconds)

* allowed all plugins to have a "libraries" folder

* added grunt build --newer
  - no clean
  - rebuild javascript only when source has changed

javascript:dev takes 2.2seconds - down from 3
javascript:compile takes 5 seconds - down from 14

### DIFF
--- a/config.js
+++ b/config.js
@@ -1,66 +1,17 @@
 require.config({
     paths: {
-        jquery: 'core/js/libraries/jquery',
-        underscore: 'core/js/libraries/underscore',
-        backbone: 'core/js/libraries/backbone',
-        modernizr: 'core/js/libraries/modernizr',
-        handlebars: 'core/js/libraries/handlebars',
-        velocity: 'core/js/libraries/velocity',
-        imageReady: 'core/js/libraries/imageReady',
-        inview: 'core/js/libraries/inview',
-        a11y: 'core/js/libraries/jquery.a11y',
-        scrollTo: 'core/js/libraries/scrollTo',
-        templates: 'templates/templates'
+        jquery: 'empty:',
+        underscore: 'empty:',
+        backbone: 'empty:',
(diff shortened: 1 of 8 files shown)


aniketdharia:
:+1: .. will test tomorrow and comment

aniketdharia:
Can we add a watch which will check if any JS is updated if not do not rebuild or copy these JS files?

taylortom:
@aniketdharia this is just a fix for the javascript compilation, we still have a watch at the end of the dev task.

oliverfoster:
@aniketdharia  i can only do that by removing the clean task on build, otherwise you've nothing to compare the srcfiles to because the clean task deletes the contents of the build folder

oliverfoster:
to explain a bit this pr is three ideas in one:

* move all libraries out of the compile process
* turn off unnecessary parts of the requirejs uglify2 process
* add a way of doing a dest-src comparative rebuild (could also be done on a dev-style task instead of the build task)

they'll have to be separate prs if we agree these things are a boon to productivity

aniketdharia:
+1 from me.. personally as it is improving the speed even of grunt dev if we make similar changes... the option of avoiding unnecessary javascript and less compilation reduces the build time drastically... as once the theme and components are fixed for a curriculum, the only updates that would happen are in JSON files. Basically publishing tasks.

oliverfoster:
need to split into separate prs

PR #1235, opened 2016-09-08
issue-1233: added grunt diff

oliverfoster:
This task is a dev rebuild, checking if files are newer and recompiling if necessary.
It doesn't do a watch or a clean.

Only performs a check on LESS, Handlebars and JavaScript

### DIFF
--- a/grunt/config/javascript.js
+++ b/grunt/config/javascript.js
@@ -39,6 +39,12 @@ module.exports = function (grunt, options) {
                 generateSourceMaps: true,
                 preserveLicenseComments:false,
                 optimize: 'none'
+            },
+            //newer configuration
+            files: {
+              '<%= outputdir %>adapt/js/adapt.min.js': [
+                '<%= sourcedir %>/**/*.js'
+              ]
             }
         },
         compile: {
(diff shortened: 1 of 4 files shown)


aniketdharia:
:+1:

aniketdharia:
Olli but I think for dev we should keep watch else the developer will need to run command everytime he updates a JSON.. what are your thoughts?

taylortom:
@aniketdharia this PR has no effect on the dev task?

aniketdharia:
Oops .

Got it..

lc-thomasberger:
Nice one! Would be good to have a Version of this that watches for json and js/less. If it listens to json changes, you get a really fast compilation time when adding content to the course.

oliverfoster:
+1 @thomas yea, should be pretty easy - following the same pattern.

aniketdharia:
+1

aniketdharia:
Should we merge?

oliverfoster:
@lc-thomasberger  it takes almost as much time to check if the json is newer as it does to rebuild it and the extra config needed it quite big. can we leave the json out for the moment?

sarveshwar-gavhane:
@oliverfoster  it would be the good enhancement for newer file checking and last modified source files, when will this changes are pushed into the framework.

oliverfoster:
@sarveshwar-gavhane i need to get the 3 +1s on it

sarveshwar-gavhane:
@oliverfoster can we close  https://github.com/adaptlearning/adapt_framework/issues/1445 with this PR or a separate one ?

oliverfoster:
a separate one would be good (y)

PR #1251, opened 2016-09-26
Issue/lp 7

lc-thomasberger:
handle languagePicker:languageChange Event:
- remove drawerItemView's
- reset DrawerCollection

### DIFF
--- a/src/core/js/drawer.js
+++ b/src/core/js/drawer.js
@@ -26,6 +26,10 @@ define(function(require) {
 		init();
 	})
 
+	Adapt.on('languagePicker:languageChange', function() {
+		DrawerCollection.reset();
+	})
+
 	Adapt.drawer = Drawer;
 
 });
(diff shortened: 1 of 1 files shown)


lc-thomasberger:
#1251

brian-learningpool:
The code looks fine in principle but it seems a bit off to me to see references to plugins in core and in other plugins.

Does anybody else think the name of the event should be changed to something like `app:changelanguage` or something similar?

moloko:
> Does anybody else think the name of the event should be changed to something like `app:changelanguage` or something similar?

I like the shorter name - but the way the code currently works, the event does originate from the languagePicker plugin so this does follow our standard convention.

A possible alternative would be to change it round a bit so that Adapt watches for changes to `config._defaultLanguage` - so all languagePickerModel has to do is set that and that will trigger Adapt to do all the reloading of .json etc. followed by it broadcasting the event `app:languageChange`.

It is getting a bit late in the day to be making big changes though...

moloko:
OK quick update - @tomgreenfield has been refactoring languagePicker and Framework so that now all the languagePicker plugin has to do is change `Adapt.config._defaultLanguage` - app.js now listens for changes to that property and will handle all the .json reloading by itself, then broadcast an `app:languageChanged` event.

I suggest we go ahead and merge this PR into feature/ml then @tomgreenfield can change drawer.js over to the new event

Any objections to this?

brian-learningpool:
Sounds good to me. +1

PR #1253, opened 2016-09-27
Move Language Picker loading into core

tomgreenfield:
Emits new event `app:languageChanged` with `language` as an argument.

### DIFF
--- a/src/core/js/app.js
+++ b/src/core/js/app.js
@@ -41,14 +41,22 @@ require([
 
     // This function is called anytime a course object is loaded
     // Once all course files are loaded trigger events and call Adapt.initialize
-    Adapt.checkDataIsLoaded = function () {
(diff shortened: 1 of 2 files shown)


src/core/js/app.js
```diff
@@ -103,7 +110,7 @@ require([
         console.error(e);
     }
 
-    Adapt.mapAdaptIdsToObjects = function() {
+    function mapAdaptIdsToObjects () {
```
brian-learningpool: Why change this?

moloko: I think @tomgreenfield was reverting [this change](https://github.com/tomgreenfield/adapt_framework/commit/cfa1f7acb171049a25bc46db9c01ed70224c43ab#diff-3eb546802b7afc77891e82198f76a794) since it's no longer necessary

brian-learningpool:
+1

PR #1260, opened 2016-09-29
startController fix for multilang courses

lc-thomasberger:
changed format in require paths
refactored startController for better code reuse

when a new language is loaded, either navigate to startHash created by startController or default location

### DIFF
--- a/src/core/js/app.js
+++ b/src/core/js/app.js
@@ -1,22 +1,23 @@
 require([
-    'coreJS/adapt',
-    'coreJS/router',
-    'coreJS/drawer',
-    'coreJS/device',
-    'coreJS/popupManager',
-    'coreJS/notify',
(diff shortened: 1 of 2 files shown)


sharpb:
I've tested this in a multi-language course we've developed and works great!

sharpb:
Though I have noticed that it makes the drawer icon appear if it is hidden?  When we initially launch the course there is no drawer icon (as it is empty we don't need it).  Change the language and it jumps straight to the correct start page but the drawer icon is shown.

moloko:
Looks good, need to test really but that's best done alongside all the other changes that are coming in so maybe just get this merged?

If you have a sec could you stick a comma on the end of line 35 of app.js and sort out the indentation on lines 165-167?

PR #1270, opened 2016-10-07
fix direction issues when switching from rtl to ltr

### LINKED ISSUE #1269: add dir-rtl | dir-ltr class to html element
These classes are required to properly style the content when switching from ltr to rtl and back again

lc-thomasberger:
resolve #1269

### DIFF
--- a/src/core/js/app.js
+++ b/src/core/js/app.js
@@ -183,7 +183,11 @@ require([
     }
 
     function onDirectionChange(model, direction) {
-        $('html').toggleClass('dir-rtl', direction === 'rtl');
+        if (direction === 'rtl') {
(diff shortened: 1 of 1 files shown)


tomgreenfield:
This doesn't fix the rendering issues in https://github.com/adaptlearning/adapt_framework/issues/1277 for me I'm afraid.

The existing code already adds/removes the `dir-rtl` class on the `html` element. The newer code just does this in a more verbose way, with the added inclusion of a `dir-ltr` class with seemingly no associated styling. Am I missing something more?

I think the issue is more down to the floated elements, in particular, 'sticking' and we need to somehow force the DOM to render the new styles. I can replicate in Chrome but not in Firefox.

PR #1287, opened 2016-10-11
add startText to trickle button

### LINKED ISSUE #1282: Language selector: Assessment trickle
The label on the first trickle button ('Begin') is always in English, but subsequent buttons are translated.

lc-thomasberger:
add startText label to trickle buttons so that the export task can get the value and add to the language files

fixes #1282


moloko:
I think you can just go ahead and merge this, it's not a change to the code and isn't going into the master branch (yet)

Issue #1294, opened 2016-10-21
Switching languages displays the resources icon and empty drawer on the new course


brian-learningpool:
On a multi-language course, if you switch the language, the resources/drawer icon is displayed when the new language loads, even if the courses has no resources.

moloko:
I can replicate this

PR #1295, opened 2016-10-21
Corrects issue with phantom drawer displaying on a language change

brian-learningpool:
This appears to be because the drawer listens to the "adapt:initialize"
before checking that it has no items and should be hidden.

Since "adapt:initialize" should not be triggered more than once, I've added
a 'forceCheckVisibility' flag which forcees this check when the language
changes.

### DIFF
--- a/src/core/js/app.js
+++ b/src/core/js/app.js
@@ -47,7 +47,6 @@ require([
             mapAdaptIdsToObjects();
 
             if (isLanguageChange) {
-                Adapt.trigger('app:languageChanged', language);
 
                 _.defer(function() {
                     var startController = new StartController();
@@ -90,6 +89,8 @@ require([
             }
             
             Adapt.setupMapping();
+
+            isLanguageChange && Adapt.trigger('app:languageChanged', language);
             
             try {
                 Adapt.trigger('app:dataReady');
@@ -111,6 +112,9 @@ require([
     }
 
     function mapAdaptIdsToObjects () {
+        // Clear any existing mappings, as they are no longer valid.
+        Adapt.mappedIds = {};
+
         Adapt.contentObjects._byAdaptID = Adapt.contentObjects.groupBy("_id");
         Adapt.articles._byAdaptID = Adapt.articles.groupBy("_id");
         Adapt.blocks._byAdaptID = Adapt.blocks.groupBy("_id");
--- a/src/core/js/app.js
+++ b/src/core/js/app.js
@@ -91,7 +91,7 @@ require([
                     Backbone.history.navigate(hash, { trigger: true, replace: true });
                 });
             }
-            
+
             try {
                 Adapt.trigger('app:dataReady');
             } catch(e) {
@@ -112,9 +112,6 @@ require([
     }
 
     function mapAdaptIdsToObjects () {
-        // Clear any existing mappings, as they are no longer valid.
-        Adapt.mappedIds = {};
-
         Adapt.contentObjects._byAdaptID = Adapt.contentObjects.groupBy("_id");
         Adapt.articles._byAdaptID = Adapt.articles.groupBy("_id");
         Adapt.blocks._byAdaptID = Adapt.blocks.groupBy("_id");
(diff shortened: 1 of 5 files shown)


brian-learningpool:
Resolves #1294.

src/core/js/views/drawerView.js
```diff
@@ -8,8 +8,10 @@ define(function(require) {
         className: 'drawer display-none',
         disableAnimation: false,
         escapeKeyAttached: false,
+        forceCheckVisibility: false,
```
moloko: could just use a comment explaining why this is necessary

moloko: also `forceCheckAvailability` would work better as a name given that its purpose is to trigger a call to `checkIfDrawerIsAvailable`

src/core/js/views/drawerView.js
```diff
@@ -78,6 +80,10 @@ define(function(require) {
         // Set tabindex for select elements
         postRender: function() {
             this.$('a, button, input, select, textarea').attr('tabindex', -1);
+
+            if (this.forceCheckVisibility) {
+                this.checkIfDrawerIsAvailable();
```
moloko: out of interest, what happens if you always run `this.checkIfDrawerIsAvailable()` here? i.e. ditch doing it on `adapt:initialize`?

src/core/js/views/drawerView.js
```diff
@@ -27,7 +27,7 @@ define(function(require) {
         setupEventListeners: function() {
             this.listenTo(Adapt, 'navigation:toggleDrawer', this.toggleDrawer);
             this.listenTo(Adapt, 'drawer:triggerCustomView', this.openCustomView);
-            this.listenToOnce(Adapt, 'adapt:initialize', this.checkIfDrawerIsAvailable);
+            this.listenTo(Adapt, 'app:dataReady', this.checkIfDrawerIsAvailable);
```
moloko: now that `checkIfDrawerIsAvailable` has been added to `postRender` it actually works fine for me without this line

moloko:
Annoyingly there is a slight flash of the resources icon when you change languages in Firefox. seems fine in IE11, Chrome and iOS10 though.
I suppose the only way to fix that would be to reverse the logic i.e. the icon should be hidden unless there are resources rather than shown unless there aren't.
It doesn't seem that serious if it's just FF though.

brian-learningpool:
@moloko, I've updated the code based on your feedback.  When testing with Vanilla though, as the navigation template is overridden it requires a `display-none` added to the Drawer button to get rid of the flash here:
https://github.com/adaptlearning/adapt-contrib-vanilla/blob/master/templates/navigation.hbs#L5

I'm happy to make this change also, but if the Vanilla theme is released prior to the framework there will be an issue with the drawer not displaying.

src/core/js/views/drawerView.js
```diff
@@ -9,7 +9,7 @@ define(function(require) {
         disableAnimation: false,
         escapeKeyAttached: false,
 
-        initialize: function() {
+        initialize: function(options) {
```
moloko: the `options` parameter is now redundant I think?

src/core/js/app.js
```diff
@@ -111,6 +112,9 @@ require([
     }
 
     function mapAdaptIdsToObjects () {
+        // Clear any existing mappings, as they are no longer valid.
+        Adapt.mappedIds = {};
```
oliverfoster: mapAdaptIdsToObjects  and mappedIds are unrelated. this should go into Adapt.setupMapping

mapAdaptIdsToObjects : is for indexing items (components, blocks, articles) in their collections
mappedIds: is for indexing the collections

brian-learningpool: Good catch. I've just moved it.

moloko:
all works for me, get rid of that `options` parameter from `drawerView.initialize` and I'll plus-one

oliverfoster:
i'll let you delete the branch

Issue #1303, opened 2016-11-08
jquery v3


oliverfoster:
http://jquery.com/download/

moloko:
done in https://github.com/adaptlearning/adapt_framework/commit/401a250219ca485d508ee6444e3cb7be49e3a504

moloko:
Had to revert back to JQuery v2 as v3 isn't compatible with the version of rangeslider we're currently using in adapt-contrib-slider. 

Should be able to come back to this when we finally update to the latest version of rangeslider

PR #1304, opened 2016-11-08
issue-1303: updated to jquery 3.1.1

oliverfoster:
https://github.com/adaptlearning/adapt_framework/issues/1303


brian-learningpool:
JQuery 3+ doesn't seem to support IE 8.
https://jquery.com/upgrade-guide/3.0/#browser-support

moloko:
@brian-learningpool different jquery library is loaded for IE8

brian-learningpool:
Great, thanks for clarifying.

oliverfoster:
i thoroughly recommend you test this with all your commonly used plugins and extensions. it may break some of the more outlandish ones.

brian-learningpool:
I guess we need a full regression test on a course with the core plugins before this is merged?

oliverfoster:
I'm not so much worried about the core plugins. It's more 3rd party plugins I'm concerned will break. We need a way to mitigate that. I'm a bit lost in the jQuery migration docs. Perhaps you could have a read and suggest a decent course of action?

brian-learningpool:
I will read when I get a chance.  Is there a pressing reason to upgrade to JQuery v3 at this moment or could it wait until the next major release of the framework?

oliverfoster:
It can wait until we're all happy.

moloko:
@brian-learningpool it's really just to take advantage of v3 (supposedly) being faster than v2

moloko:
I just used JQuery v3 (and the updated inview) in a build where I was testing some updates made to confidenceSlider. Didn't notice any problems but it wasn't the most exhaustive of tests by any means. Only checked in Firefox for a start.

moloko:
should we stick a post on the forums asking people to test this (and maybe inview as well)?

brian-learningpool:
Hi @moloko, to answer your question (apologies for the delay), I think yes, we we should ask forum users to test this.

moloko:
done: https://community.adaptlearning.org/mod/forum/discuss.php?d=1845

moloko:
no feedback from forum users; agreed with @brian-learningpool to put this on hold until 02/02/17

PR #1313, opened 2016-11-11
Resolves #1312, performance improvement for Grunt copyMain task

### LINKED ISSUE #1312: Performance issue with globbing pattern in copy.js
The globbing pattern on the re-worked copy.js to pull in the `required` and `libraries` folders is a little too lose.  While the theme and menu can be defined and passed to Grunt -- and are passed in by the authoring tool -- they are not used to restrict the pattern.

In situations where there are lots of menus and themes in their respective folders this causes a significant performance degradation.  

I'm proposing that we modify the copyMain to remove the leading **/* and pass in the theme and menu folder name (if available).

brian-learningpool:
Replacing the **/* wildcard with the specific folder names to search in for the 'required' and 'libraries' folders.

### DIFF
--- a/grunt/config/copy.js
+++ b/grunt/config/copy.js
@@ -181,7 +181,7 @@ module.exports = function (grunt, options) {
                 },
                 {
                     expand: true,
-                    src: ['**/libraries/**/*'],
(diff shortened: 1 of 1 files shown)


oliverfoster:
seems to work with grunt dev, so not changing terminal behaviour
needs AT testing

+1

Issue #1227, opened 2016-09-06
grunt server not automatically launching course on Windows


moloko:
When you run the `$ grunt server` command the course does not automatically open in a browser on Windows (tested using Windows 10 with Firefox and Chrome as default browser). You can obviously open the course manually - but a novice user wouldn't know to do that, particularly as it's not clear what correct URL is.

Seems fine on OS X and Ubuntu

moloko:
seems to be sporadic... not working for me or @lc-thomasberger but working for @tomgreenfield (Win 10, Firefox)

finlayb:
Running Windows 10 and Chrome and this happens to me sometimes. Usually just run it a second time and it works.

sarveshwar-gavhane:
tested on windows 10 with Firefox and Chrome as default browser on two different machine's with 2.0.12 version it works fine on one machine and but not on another.

NayanK444:
I made following changes in connect.js file.
```
module.exports = function (grunt, options) {

  var port = grunt.option('port') || 9001;

  return {
    server: {
      options: {
        port: port,
        base: '<%= outputdir %>',
        keepalive:true,
        open: true
      }
    },
    spoorOffline: {
      options: {
        port: port,
        base: '<%= outputdir %>',
        keepalive:true,
        open: true
      }
    }
  }
}
```

moloko:
@NayanKhedkar I just went to test this - only to find that it has started working again without me having changed anything, so I'm currently unable to verify whether these changes fix the problem or not!

I'll try again tomorrow

NayanKhedkar:
@moloko I tested it on windows 10 with IE11, Firefox and Chrome as default browser on various machines.
It worked well even on those machines on which it was not working properly previously.

moloko:
Well I keep trying but still it is working normally for me.

However, I did notice that if I modified connect.js as detailed above I'd end up with two instances of the course being opened...

sarveshwar-gavhane:
https://github.com/adaptlearning/adapt_framework/pull/1317

moloko:
done in https://github.com/adaptlearning/adapt_framework/commit/39d2b8aa8cd1891adbb84581489985f72719140c

PR #1317, opened 2016-11-15
Fix for grunt server not always opening the course on Windows

sarveshwar-gavhane:
removed open : server from concurrent.js since it doesn't assure for launching course every time .
added open : true in connect.js that launching course every time.


moloko:
typically the existing version is still working for me so I can't test whether this fixes the problem or not :-(

chetan-hajare:
The issue https://github.com/adaptlearning/adapt_framework/issues/1227 was replicating at my end but this PR had resolved the issue atleast at my end.

Issue #1302, opened 2016-11-04
inview - update to latest version


oliverfoster:
new inview library being released:

https://github.com/adaptlearning/jquery.onscreen/issues/6

please replace your ```core/js/libraries/inview.js``` with this version:

https://raw.githubusercontent.com/adaptlearning/jquery.onscreen/issue-6/jquery.onscreen.js

then leave bugs, comments etc on this pr:

https://github.com/adaptlearning/jquery.onscreen/pull/7

PR #1321, opened 2016-11-17
 issue-1306: allowScrollOver + first-measure fixes

oliverfoster:
* added allowScrollOver configuration in config.json
* trigger inview check after pageView:ready and menuView:ready
* fix inview force check on first measurement
* improved scrolled over calculations

config.json:
```json
   "_inview": {
        "_allowScrollOver": false
    }
```

### DIFF
--- a/src/core/js/app.js
+++ b/src/core/js/app.js
@@ -115,7 +115,7 @@ require([
 
     function configureInview() {
         
-        var adaptConfig = Adapt.course.get("_inview");
+        var adaptConfig = Adapt.config.get("_inview");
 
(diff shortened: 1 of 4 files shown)


oliverfoster:
https://github.com/adaptlearning/adapt_framework/issues/1302

src/core/js/app.js
```diff
@@ -111,6 +113,18 @@ require([
         console.error(e);
     }
 
+    function configureInview() {
+        
+        var adaptConfig = Adapt.course.get("_inview");
```
moloko: do we want this setting in course.json? seems to me like it would fit in config.json better

PR #1332, opened 2016-11-22
fixes issue/1331

lc-thomasberger:
add more screen related properties to Adapt.device object

### DIFF
--- a/src/core/js/device.js
+++ b/src/core/js/device.js
@@ -56,12 +56,16 @@ define(function(require) {
         return (Adapt.device.screenWidth >= Adapt.device.screenHeight) ? 'landscape' : 'portrait';
     }
 
+    function getScreenAspectRatio() {
+        return Adapt.device.screenWidth / Adapt.device.screenHeight;
+    }
+
     var onWindowResize = _.debounce(function onScreenSizeChanged() {
         // Calculate the screen properties:
         Adapt.device.screenWidth = getScreenWidth();
         Adapt.device.screenHeight = getScreenHeight();
         Adapt.device.orientation = getScreenOrientation();
-        Adapt.device.aspectRatio = Adapt.device.screenWidth / Adapt.device.screenHeight;
+        Adapt.device.aspectRatio = getScreenAspectRatio();
 
         var newScreenSize = checkScreenSize();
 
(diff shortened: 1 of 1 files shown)


src/core/js/device.js
```diff
@@ -44,9 +44,22 @@ define(function(require) {
...
     }
 
+    function getScreenHeight() {
+        return isAppleDevice()
+            ? getAppleScreenHeight()
+            : window.innerHeight || $window.height();
+    }
+
+    function getScreenOrientation() {
+        return (Adapt.device.screenWidth >= Adapt.device.screenHeight) ? 'landscape' : 'portrait';
```
oliverfoster: I seem to remember the screen height doesn't change when the orientation changes on iOS. So you'd always have to use the window.innerWidth to calculate the orientation. I'll see if that applies here a bit later and find some references if applicable.

src/core/js/device.js
```diff
@@ -77,6 +90,10 @@ define(function(require) {
         return (Math.abs(window.orientation) === 90) ? screen.height : screen.width;
     }
 
+    function getAppleScreenHeight() {
+        return (Math.abs(window.orientation) === 90) ? screen.width : screen.height;
```
oliverfoster: Cool. Ignore my last comment. :+1:

src/core/js/device.js
```diff
@@ -4,10 +4,12 @@ define(function(require) {
...
-    Adapt.device = {
-        touch: Modernizr.touch,
-        screenWidth: getScreenWidth()
-    };
+    Adapt.device = {};
+    Adapt.device['touch'] = Modernizr.touch;
+    Adapt.device['screenWidth'] = getScreenWidth();
+    Adapt.device['screenHeight'] = getScreenHeight();
+    Adapt.device['orientation'] = getScreenOrientation();
+    Adapt.device['aspectRatio'] = getScreenAspectRatio();
```
brian-learningpool: Was there a reason for switching to bracket notation here?

lc-thomasberger: Yes, to be able to reuse the functions. When called from the object, Adapt.device is not defined.

oliverfoster: You can do ```Adapt.device.touch = ``` is what he means.

Issue #1095, opened 2016-06-09
There is no loading feedback until after the CSS/JS is loaded


LukaszGrela:
There is noticeable blank screen before loading animation is displayed this is caused by the delay needed to load CSS and JS files as the loading is created via code. Can this be separated and be part of index page so loading feedback is given to the user from the earliest possible?

LukaszGrela:
I have sorted this out with following changes (I was trying to avoid any code changes so this is a duplicate)

`index.html`
```
<!--[if IE]><![endif]-->
<!doctype html>
<!--[if IE 7 ]><html id="adapt" class="ie ie7 no-js" lang="en"><![endif]-->
<!--[if IE 8 ]><html id="adapt" class="ie ie8 no-js" lang="en"><![endif]-->
<!--[if IE 9 ]><html id="adapt" class="ie ie9 no-js" lang="en"><![endif]-->
<!--[if gt IE 9]><!-->
<html id="adapt" class="no-js" lang="en">
<!--<![endif]-->
<head>
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta content="text/html;charset=utf-8" http-equiv="Content-Type">
    <meta content="utf-8" http-equiv="encoding">
    <title></title>
    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
    <style>
        .loading-init {
            width: 100%;
            height: 100%;
            position: fixed;
            top: 0;
            left: 0;
            text-align: center;
            z-index: 5000;
        }

        .dir-rtl .loading-init {
            left: auto;
            right: 0;
        }

        .loading-init .loader-gif {
            background-image: url('adapt/css/assets/ajax-loader.gif');
            background-repeat: no-repeat;
            background-position: center;
            width: 128px;
            height: 128px;
            position: absolute;
            top: 50%;
            left: 50%;
            margin-top: -64px;
            margin-left: -64px;
            font-size: 20px;
            font-weight: 300;
            line-height: 120%;
        }
        .pixel-density-low .loading-init,
        .pixel-density-high .loading-init,
        .size-large .loading-init,
        .size-medium .loading-init,
        .size-small .loading-init
         {
            display: none;
        }
    </style>
    <link href="assets/certificate-print.css" type="text/css" media="print" rel="stylesheet" />
    <link href="adapt/css/adapt.css" type="text/css" rel="stylesheet" />
[...]

PR #1334, opened 2016-11-23
Issue/1095

lc-thomasberger:
add initial loading spinner to index.html
will be removed once Adapt is loaded

Original issue: #1095.

### DIFF
--- a/src/core/js/app.js
+++ b/src/core/js/app.js
@@ -24,6 +24,7 @@ require([
     // Append loading template and show
     window.Handlebars = _.extend(require("handlebars"), window.Handlebars);
 
+    removeInitialLoadingMessage();
     var template = Handlebars.templates['loading'];
     $('#wrapper').append(template());
     
@@ -201,6 +202,10 @@ require([
         } else {
             Adapt.config.set('_activeLanguage', Adapt.config.get('_defaultLanguage'));
         }
+    }
+
+    function removeInitialLoadingMessage() {
+        $('#initial-loading').remove();
     }
 
     // Events that are triggered by the main Adapt content collections and models
--- a/src/core/js/app.js
+++ b/src/core/js/app.js
@@ -23,10 +23,6 @@ require([
 
     // Append loading template and show
     window.Handlebars = _.extend(require("handlebars"), window.Handlebars);
-
-    removeInitialLoadingMessage();
-    var template = Handlebars.templates['loading'];
-    $('#wrapper').append(template());
     
     Adapt.config = new ConfigModel(null, {url: "course/config.json", reset:true});
     Adapt.config.on({
@@ -202,10 +198,6 @@ require([
         } else {
             Adapt.config.set('_activeLanguage', Adapt.config.get('_defaultLanguage'));
         }
-    }
-
-    function removeInitialLoadingMessage() {
-        $('#initial-loading').remove();
     }
 
     // Events that are triggered by the main Adapt content collections and models
--- a/src/core/js/scriptLoader.js
+++ b/src/core/js/scriptLoader.js
@@ -8,6 +8,23 @@
         return false;
     })();
 
+    // Inject loading spinner into wrapper
+    domReady(function() {
+        var spinner = document.createElement("div");
+        var attr = document.createAttribute("class");
+        attr.value = "loading";
+        spinner.attributes.setNamedItem(attr);
+        spinner.innerHTML = 
+            '<style>' +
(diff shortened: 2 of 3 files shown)


lc-thomasberger:
will also need a PR for the index files in spoor

src/core/js/app.js
```diff
@@ -203,6 +204,10 @@ require([
         }
     }
 
+    function removeInitialLoadingMessage() {
```
moloko: given that this is only called once I'm not sure it's worth making it into a function?

moloko:
Works well (on Firefox anyway).

My only other comment is that there's a bit of a jump in the "Loading..." text when the font loads, so it goes from this:
2016-11-24 17_01_26-mozilla firefox
To this:
2016-11-24 17_01_01-mozilla firefox

Do other people think this is a problem?

lc-thomasberger:
@moloko valid point. I tried to keep the styling in the html as light as possible. Using the same font as in the content will require to use google font directly from the html style rules.

I thin it should be enough to just give some visual feedback. A developer would then probably customize it further to fit the requirements. 
We might also only show the spinner without the text as the initial loading spinner.

src/core/js/app.js
moloko: `$('#initial-loading').remove();` seems clear enough to me but I'd be interested to hear what others thing

tomgreenfield:
Does the browser render anything before it's grabbed `adapt.css`?

Suggest we instead simply add the existing loading divs to the wrapper:
```html
<div id="wrapper">
    <div class="loading">
        <div class="loader-gif">
            <div role="heading" tabindex="0" class="h3" aria-level="1">Loading...</div>
        </div>
    </div>
</div>
```
And remove these [lines](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/app.js#L27-L28) and associated [template](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/templates/loading.hbs).

lc-thomasberger:
@tomgreenfield yes Chrome waits until adapt.css is loaded. Even if I change the order of the style tag and the call to the stylesheet, the first render is only once the css file is loaded. Therefore we should get rid of the style tag and move it to core.less. 

What do others think about moving loading markup into the wrapper?

tomgreenfield:
Cool. With the suggested approach we don't need any additional styling; it just uses the existing Less.

oliverfoster:
as a maintainability thing: we're gonna have to put all this in every .html file in scorm etc. did you not want to inject it from the script loader / such? one place to maintain etc.

src/core/js/scriptLoader.js
```diff
@@ -8,6 +8,23 @@
...
+    domReady(function() {
+        var spinner = document.createElement("div");
+        var attr = document.createAttribute("class");
+        attr.value = "loading";
+        spinner.attributes.setNamedItem(attr);
+        spinner.innerHTML = 
+            '<style>' +
+            '.loading {position: fixed!important;background-color:transparent;text-align: center;}'+
+            '.loading .loader-gif {background-image: url(adapt/css/assets/ajax-loader.gif);background-repeat: no-repeat;background-position: center;width: 128px;height: 128px;position: absolute;top: 50%;left: 50%;margin-top: -64px;margin-left: -64px;}' +
+            '</style>' +
```
tomgreenfield: The inline `<style>` shouldn't be necessary here as styling already exists in the Less.

lc-thomasberger: yes that is correct. I added it as an inline style, so that browsers can render it before the main.css is loaded. Looks like Chrome waits until all css files are loaded.

oliverfoster: my fault, sorry

src/core/js/scriptLoader.js
```diff
@@ -8,6 +8,18 @@
         return false;
     })();
 
+    // Inject loading spinner into wrapper
+    domReady(function() {
+        var spinner = document.createElement("div");
+        var attr = document.createAttribute("class");
+        attr.value = "loading";
+        spinner.attributes.setNamedItem(attr);
```
tomgreenfield: In Firefox I get a warning that `document.createAttribute()` is deprecated. Would it also just be simpler to condense the three lines into one?
```javascript
spinner.className = "loading";
```

src/core/js/app.js
```diff
@@ -23,9 +23,6 @@ require([
 
     // Append loading template and show
```
tomgreenfield: We'll want to remove this comment as well.

PR #1345, opened 2016-11-29
Added new orientation classes - orientation-portrait and orientation-…

NayanKhedkar:
…landscape

### DIFF
--- a/src/core/js/device.js
+++ b/src/core/js/device.js
@@ -14,8 +14,9 @@ define(function(require) {
 
         $('html').addClass("size-" + Adapt.device.screenSize);
 
-        // As Adapt.config is available it's ok to bind the 'resize'.
+        // As Adapt.config is available it's ok to bind the 'resize' and 'orientationchange'.
         $window.on('resize', onWindowResize);
+        $window.on('orientationchange', onOrientationChange);
     });
 
     /**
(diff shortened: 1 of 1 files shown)


src/core/js/device.js
```diff
@@ -61,7 +62,28 @@ define(function(require) {
...
+                windowOrientationName = 'landscape';
+                break;
+            default:
+                $('html').removeClass("orientation-landscape");
+                windowOrientationName = 'portrait';
+                break;
+        }
+        Adapt.device.orientationName = windowOrientationName;
+        $('html').addClass("orientation-" + windowOrientationName);
+        Adapt.trigger('device:orientationchange', windowOrientation, windowOrientationName);
```
lc-thomasberger: Not sure if we should have orientationchange as a single event. Orientation is stored in the public Adpat.device object, so it can be accessed in Eventhandlers bound to `device:resize` or `device:changed`.

oliverfoster: i agree, we should certainly do ``$window.on('resize orientationchange', onWindowResize);`` though

brian-learningpool: Also, is it necessary to store `orientationName` as well as `orientation`?

NayanKhedkar:
Any one of them will do. I will update the code using one of them.

NayanKhedkar:
should I close this PR

brian-learningpool:
Yes, I think so, but it would be useful if you could check #1349 works for you.

PR #1346, opened 2016-11-29
Added ES5 shim for IE 8

brian-learningpool:
I've added the shim from https://github.com/inexorabletash/polyfill/blob/master/es5.js.

This provides implementations for:
- Object.getPrototypeOf
- Object getOwnPropertyNames
- Object.create
- Object.defineProperty
- Object.defineProperties
- Object.keys
- Function.prototype.bind
- Array.isArray
- Array.prototype.indexOf
- Array.prototype.lastIndexOf
- Array.prototype.every
- Array.prototype.some
- Array.prototype.forEach
- Array.prototype.map
- Array.prototype.filter
- Array.prototype.reduce
- Array.prototype.reduceRight
- String.prototype.trim
- Date.now
- Date.prototype.toISOString

### DIFF
--- a/src/core/js/libraries/es5-shim.min.js
+++ b/src/core/js/libraries/es5-shim.min.js
@@ -1,0 +1,78 @@
+/**
+https://github.com/inexorabletash/polyfill/blob/master/es5.js, 24th November 2016
+
+https://github.com/inexorabletash/polyfill/blob/master/LICENSE.md
+*/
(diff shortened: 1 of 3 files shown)


brian-learningpool:
Good point, @oliverfoster.  This is now done, tested locally in IE 8.

oliverfoster:
ta dude

PR #1349, opened 2016-12-01
Re-worked device.js to make use of Bowser 1.5 features

brian-learningpool:
- Added support for orientation-landscape and orientation-portrait classes
- Added new 'renderingEngine' property
- Added more advanced support for detecting the operating system
- Renamed Adapt.device.OS to Adapt.device.os

### DIFF
--- a/src/core/js/accessibility.js
+++ b/src/core/js/accessibility.js
@@ -173,7 +173,7 @@ define(function(require) {
             var bottomoffset = 0;
             $.a11y.options.focusOffsetTop = topOffset;
             $.a11y.options.focusOffsetBottom = bottomoffset;
-            $.a11y.options.OS = Adapt.device.os.toLowerCase();
+            $.a11y.options.OS = Adapt.device.OS.toLowerCase();
             $.a11y.options.isTouchDevice = Modernizr.touch;
 
             if (this.isActive()) {
--- a/src/core/js/accessibility.js
+++ b/src/core/js/accessibility.js
@@ -173,7 +173,7 @@ define(function(require) {
             var bottomoffset = 0;
             $.a11y.options.focusOffsetTop = topOffset;
             $.a11y.options.focusOffsetBottom = bottomoffset;
-            $.a11y.options.OS = Adapt.device.OS.toLowerCase();
+            $.a11y.options.OS = Adapt.device.os.toLowerCase();
             $.a11y.options.isTouchDevice = Modernizr.touch;
 
             if (this.isActive()) {
--- a/src/core/js/device.js
+++ b/src/core/js/device.js
@@ -10,7 +10,7 @@ define(function(require) {
         screenHeight: getScreenHeight(),
         browser: Bowser.name,
         version: Bowser.version,
-        os: getOperatingSystem(),
+        OS: getOperatingSystem(),
         osVersion: Bowser.osversion || '',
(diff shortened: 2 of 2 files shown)


brian-learningpool:
In Internet Explorer 8 this requires the ES 5 shim as detailed here: https://github.com/adaptlearning/adapt_framework/pull/1346

That must go in before this code is merged.

brian-learningpool:
Both `orientation` and `aspectRatio` are derived from the `screenWidth` and `screenHeight` properties, which are still updated on the window's `resize`/`orientationchange` handler.  So when the `screenWidth` or `screenHeight` changes, the orientation and aspect ratio are automatically re-evaluated, since they are only getters now, and don't have to be explicitly set via another function, which is how they were before.

oliverfoster:
Yea, I get that, but it still doesn't make sense to me using a shim in a situation where it clearly isn't needed. Those values are calculable at the point the device size changes. At not point between the device resize events will the derived value be different than it is on the event. Having a live getter offers no advantage and could be more expensive as the value is recalculated every time despite it not changing.

brian-learningpool:
The shim at #1346 isn't exclusively for this though.  Its purpose was to address gaps in ES 5 support in IE 8 and allow us to produce more structured and readable code -- which I think this PR is -- while we're still supporting obsolete browsers.

I think having a getter for both these properties is a more correct approach in terms of OO design, as both properties _should_ be read-only.

oliverfoster:
Should we make all the other device properties read-only as well, otherwise it's two completely nixed styles here? The device object could become a class with static properties if we're going full OO.

brian-learningpool:
It's an idea but I don't think we can right now without introducing breaking changes in IE 8.

brian-learningpool:
This is dependent on adaptlearning/adapt_framework#1346 for IE 8 compatibility.

moloko:
@lc-thomasberger now you mention it I noticed just yesterday that on the [ES5 compatibility table](https://kangax.github.io/compat-table/es5/) `object.defineProperty` is not listed as being supported via ES5-shim..

brian-learningpool:
@lc-thomasberger, `Adapt.device.orientation` never worked in IE 8.  

The purpose of the shim wasn't to get this function working in IE 8, it was to ensure the ES 5 functions wouldn't break.

brian-learningpool:
Actually, hold off on merging this.  I noticed that the previous version of Bower actually output the iOS device type as a class too, i.e. `iphone`, `ipad`, etc.  I'd like to add that back in.

brian-learningpool:
OK, now it can be merged. 👍

tomgreenfield:
Is there a particular reason why `Adapt.device.OS` has been renamed to `Adapt.device.os`?

This will be a breaking change for plugins that reference this property.

brian-learningpool:
Fair point, @tomgreenfield.  I was trying to standardise the case of the property names in Adapt.device to camelCase.  I had updated the references in the core framework code but you're right, there is a chance that non-core plugins out there might reference this.  I've switched it back.

oliverfoster:
i'm still not happy with the ```Object.defineProperty```. It's a waste of cpu cycles in circumstances where i'll need to repeatedly call those two properties. They are derivable at a single point in time and can be calculated once and once only on the window size change.

PR #1356, opened 2016-12-07
Reverts #1095

brian-learningpool:
This allows loading.hbs in the theme to set the loading content.


andrewhosgood:
+1

oliverfoster:
+1
did we still want an early loading screen?
one that at least gives an indication of something before the styled loading screen appears?

moloko:
@oliverfoster yes definitely

brian-learningpool:
I agree, and I've re-opened https://github.com/adaptlearning/adapt_framework/issues/1095.

PR #1359, opened 2016-12-09
Allow plugins to pause initialisation

chris-steele:
Allow plugins to pause initialisation between the app:dataReady and adapt:start/adapt:initialisation events. This complements the earlier addition that allowed plugins to pause initialisation between app:dataLoaded and app:dataReady.


src/core/js/app.js
```diff
@@ -112,6 +112,13 @@ require([
             outputError(e);
         }
 
+        if (!Adapt.isWaitingForPlugins()) triggerInitialisation();
+        else Adapt.once('plugins:ready', function() {
+            triggerInitialisation();
```
tomgreenfield: Any reason we couldn't condense down to just the named function?
```js
Adapt.once('plugins:ready', triggerInitialisation);
```

chris-steele: @tomgreenfield something odd going on with this PR - most of the commits have been merged. The line you've quoted is out of date (https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/app.js#L87)

chris-steele: We obviously don't want this change, so propose this PR just be closed

Issue #1291, opened 2016-10-19
adapt-contrib-assessment: Suppress marking/feedback until after final attempt


brian-learningpool:
It would be a good enhancement if we had a new option to suppress feedback (marking, model answers) until after the last assessment attempt was completed.  Obviously this would only work in cases where there is a finite number of attempts.

This was raised in the forum here:
https://community.adaptlearning.org/mod/forum/discuss.php?d=1554

sarveshwar-gavhane:
https://github.com/adaptlearning/adapt-contrib-assessment/pull/78

moloko:
done via https://github.com/adaptlearning/adapt_framework/commit/516a4c154e25fd91fd20fad32099eb8ed75c6bb0 and https://github.com/adaptlearning/adapt-contrib-assessment/commit/4a58f24c25770be3cdc72ba00605bbb4abaa047b

PR #1360, opened 2016-12-09
Issue #1291 - Suppress marking/feedback until after final attempt

barry-learningpool:
This is required by the changes to the Assessment extension, related PR here: https://github.com/adaptlearning/adapt-contrib-assessment/pull/78

Added the ability to 'refresh' a question view at any time i.e. ensure its display is accurate, given the component's current state variables.


src/core/js/views/buttonsView.js
```diff
@@ -76,48 +72,51 @@ define(function() {
...
-                this.$('.buttons-action').addClass('buttons-action-fullwidth');
-                this.$('.buttons-feedback').addClass('no-feedback');
-            }
+          // If feedback should be hidden, the 'Submit' button should stretch
+          // to fill the space and the 'Feedback' button should be hidden.
+          // (These classes must be implemented in the theme.)
+          this.$('.buttons-action').toggleClass('buttons-action-fullwidth', !this.model.get('_canShowFeedback'));
+          this.$('.buttons-action').toggleClass('buttons-action-enlarge', !this.model.get('_canShowFeedback'));
+          this.$('.buttons-feedback').toggleClass('no-feedback', !this.model.get('_canShowFeedback'));
+          this.$('.buttons-marking-icon').toggleClass('no-feedback', !this.model.get('_canShowFeedback'));
```
tomgreenfield: Could we reuse variables here rather than doing multiple lookups?

For example:
```js
var canShowFeedback = this.model.get('_canShowFeedback');

this.$('.buttons-action').toggleClass('buttons-action-fullwidth buttons-action-enlarge', !canShowFeedback);
this.$('.buttons-feedback').toggleClass('no-feedback', !canShowFeedback);
this.$('.buttons-marking-icon').toggleClass('no-feedback', !canShowFeedback);
```

src/core/js/views/buttonsView.js
```diff
@@ -76,48 +72,51 @@ define(function() {
...
+            var ariaLabel = this.model.get('_buttons')["_" + changedAttribute].ariaLabel;
+            var buttonText = this.model.get('_buttons')["_" + changedAttribute].buttonText;
+
+            switch (changedAttribute) {
+              case "submit":
+              case "showCorrectAnswer":
+              case "hideCorrectAnswer":
+                //make model answer button inaccessible but enabled for visual users
+                //	due to inability to represent selected incorrect/correct answers to a screen reader, may need revisiting
+                this.$('.buttons-action').a11y_cntrl_enabled(true).html(buttonText).attr('aria-label', ariaLabel);
```
tomgreenfield: Has `a11y_cntrl(false)` been amended intentionally to `a11y_cntrl_enabled(true)`?

The existing code exists specifically to make the *Correct/My answer* buttons inaccessible to screenreaders whilst still remaining enabled.

src/core/js/views/questionView.js
```diff
@@ -279,6 +282,23 @@ define([
...
+          this._runModelCompatibleFunction("updateButtons");
+
+          if (this.model.get('_canShowMarking') && this.model.get('_isInteractionComplete') && this.model.get('_isSubmitted')) {
+            this.showMarking();
+          }
+
+          if (this.buttonsView) {
+            _.defer(_.bind(function() {
+                this.buttonsView.refresh();
+              }, this)
```
tomgreenfield: I'm pretty sure this may be simplified to:
```js
_.defer(_.bind(this.buttonsView.refresh, this));
```

barry-learningpool:
@tomgreenfield requested changes made

src/core/js/views/buttonsView.js
```diff
@@ -96,28 +92,29 @@ define(function() {
                 var buttonText = this.model.get('_buttons')["_" + changedAttribute].buttonText;
 
                 switch (changedAttribute) {
-                    case "showCorrectAnswer": case "hideCorrectAnswer":
+                    case "submit":
+                    case "showCorrectAnswer":
+                    case "hideCorrectAnswer":
```
tomgreenfield: Has `"submit"` been intentionally added here as a case?

src/core/js/views/buttonsView.js
```diff
@@ -96,28 +92,29 @@ define(function() {
...
     				    //make model answer button inaccessible but enabled for visual users
     				    //	due to inability to represent selected incorrect/correct answers to a screen reader, may need revisiting
-                        this.$('.buttons-action').a11y_cntrl(false).html(buttonText).attr('aria-label', ariaLabel);
+                        this.$('.buttons-action').a11y_cntrl_enabled(false).html(buttonText).attr('aria-label', ariaLabel);
```
tomgreenfield: This still should be left as `ally_cntrl` as per https://github.com/adaptlearning/adapt_framework/pull/1360#discussion_r97984171.

src/core/js/views/buttonsView.js
oliverfoster: It does look like it.

src/core/js/views/buttonsView.js
oliverfoster: Agreed.

src/core/js/views/questionView.js
tomgreenfield: May I retract this comment – realise now you're needing to actually call the function at this point! Apologies.

tomgreenfield: Or this works:
```js
_.defer(_.bind(this.buttonsView.refresh, this.buttonsView));
```
(without the anonymous function)

src/core/js/views/buttonsView.js
barry-learningpool: Sorry, this was my misunderstanding. I'll make the change.

I've noticed an issue around this, the buttons aren't being enabled. Do I need to call
this.$('.buttons-action').a11y_cntrl_enabled(true).a11y_cntrl(false) to ensure the button is 'enabled' but is then 'aria-hidden'?

oliverfoster: just reverting your changes to these lines should do it: https://github.com/barry-learningpool/adapt_framework/blob/135dddd29245bf9456bb3f9581c7c460b43b508e/src/core/js/views/buttonsView.js#L94-L106

barry-learningpool: I tried that @oliverfoster but I was seeing the aforementioned issue. I think it's due to the fact that the buttons now go to an intermediate 'submitted and disabled' state before the display of the 'model answer' buttons. This means I need to explicitly enable the button first. I've pushed a change for this.

oliverfoster: i'll have to investigate further. i still don't understand how allowing a buttons to refresh helps with the issue specified.

moloko:
there's some conflicts in buttonsView.js that need resolving...

src/core/js/views/buttonsView.js
tomgreenfield: When I revert the `onButtonStateChanged()` function as @oliverfoster suggests, it fixes it for me.

barry-learningpool: I've reverted the function back and I'm still able to see the issue I mentioned.

I have a simple assessment with two MCQ components with 2 allowed attempts and suppress marking enabled. On submission of the second attempt and completion of the assessment (having answered incorrectly on both questions), the 'Show correct answer' button is disabled on the first MCQ, second is enabled fine.

It seemed to be down to the fact that the submit button is initially disabled, then when 'a11y_cntrl_enabled(false)' occurs, it disables the aria stuff, but doesn't explicitly enable the button. Hopefully that makes sense.

I've resolved the issue in this commit by ensuring the button is enabled first:
https://github.com/adaptlearning/adapt_framework/pull/1360/commits/7f0e4ab43fa5906ddd8ce7ee0ae652e72bf22fb9

oliverfoster: So.. Tom and I have had a look together. There are a whole series of issues here.

* updateButtons is a "click event handler" it was really intended to switch the button states from one mode to another on the buttons click. you are using it as part of your refresh functionality to maintain the correct state for the changed attributes.. so it probably shouldn't be used for that at all and the multiple calls are causing havock with states
* updateButtons is being called twice from the questionModel.refresh function. once in the function and then again from the questionView event handler for the ```question:refrseh``` event
* we don't have a state for "incorrect + don't show model answer button" but we should, instead it reverts to "correct" (as this doesn't show the "show model answer" button) even though the question was incorrectly answered
* the assessment doesn't have a handle to the questionView so instead it calls refresh on the questionModel, the questionModel then triggers a refresh event to the listening questionViews... this would work fine if the questionView wasn't then changing the questionModel. It's a horrid call stack, the refresh on the questionModel should make sure the questionModel state is consistent then trigger for the view to update. period.
* you were right about the buttons-action button not being made enabled properly when the state changes, but i don't think changing the current functions is going to help, especially with the above outstanding issues

i'm sure there are a few other bits i've missed... as a result going to have to remove my +1 from the assessment pr because i don't want to complicate an already bad situation and that shouldn't be merged as it needs the questionModel.refresh function.

barry-learningpool: Thanks for the detailed response @oliverfoster, very helpful. I've pushed a commit there which hopefully addresses everything you mentioned.

oliverfoster:
very nice. does it all work as expected now?

barry-learningpool:
Yeah I tested it out, it's working well

brian-learningpool:
Nice job.  That was a bit of a baptism of fire for a framework a pull request.  

Almost there, though your changes will be impacted by the enum changes here:
https://github.com/adaptlearning/adapt_framework/pull/1442

oliverfoster:
depending on which pr goes in first you might get a conflict with one i'm making for the same code areas https://github.com/adaptlearning/adapt_framework/pull/1442
we'll cross that bridge when we get to it.

@tomgreenfield doesn't like your nested if statement, and i've just noticed a little weird bit of logic in it. so i'll write a comment. thanks for updating so quick.

src/core/js/views/questionView.js
```diff
@@ -279,6 +282,38 @@ define([
         // be a full reset
         resetQuestion: function() {},
 
+        determineButtonState: function() {
```
oliverfoster: this function should be in the model really. it does only model calculations.

```javascript
            var state = "submit";

            if (this.get('_isInteractionComplete')) {

                if (this.get('_isCorrect')) {
                    state = 'correct';
                } else if (!this.get('_canShowModelAnswer')) {
                    state = 'incorrect';
                } else {
                    state = 'showCorrectAnswer';
                }

            } else if (this.get('_isSubmitted')) {
                // use _isSubmitted instead as it reads better?
                state = 'reset';

            }

            // write the set code once only to reduce code density
            this.set('_buttonState', state);
```

oliverfoster: you don't have to redo the if statement if you can't see any way to make it not nested. i couldn't see any way to make it non-nested.

tomgreenfield: Just my two cents...the easiest way I find reading the logic is with with a separate getter, then you can return early:
```js
getButtonState: function() {
    if (this.get('_isCorrect')) return 'correct';

    if (this.get('_attemptsLeft') === 0) {
        return this.get('_canShowModelAnswer') ? 'showCorrectAnswer' : 'incorrect';
    }

    return this.get('_isSubmitted') ? 'reset' : 'submit';
}
```
P.S. **not tested** – just the theory.

src/core/js/views/questionView.js
```diff
@@ -279,6 +282,38 @@ define([
...
+                } else {
+                    this.model.set('_buttonState', 'reset');
+                }
+            }
+        },
+
+        refresh: function() {
+            this.renderState();
+              
+            this.determineButtonState();
```
oliverfoster: with:

```javascript
this.model.set("_buttonState", this.model.getButtonState());
```

instead of:
```javascript
this.determineButtonState();
```

src/core/js/views/questionView.js
oliverfoster: with:
```javascript
         refresh: function() {
             this.renderState();
               
             this.model.set("_buttonState", this.model.getButtonState() );
 
             if (this.model.get('_canShowMarking') && this.model.get('_isInteractionComplete') && this.model.get('_isSubmitted')) {
                 this.showMarking();
             }
 
             if (this.buttonsView) {
                 _.defer(_.bind(this.buttonsView.refresh, this.buttonsView));
             }
         },
```

oliverfoster: wow, i'm really sorry about that... x) too many lines

Issue #1370, opened 2017-01-05
accessibility - add option to disable automatic switching-on of accessibility on touch devices


moloko:
Quite often it's only a requirement to have accessibility on desktop browsers (usually only IE) so it would be useful to be able to disable the 'auto-on' feature for touch devices as accessibility on iOS can trigger bugs that don't appear on the desktop version.

For example, I had an iOS accessibility bug with [articleBlockSlider](https://github.com/cgkineo/adapt-articleBlockSlider) on a recent project where focus would always jump to the first component in the ABS on page load - but only on iOS. Obviously in an ideal world I'd fix the underlying issue with ABS, but since there was no iOS accessibility requirement for this project it was far easier (and cheaper!) just to disable accessibility for touch devices ;-)

moloko:
this functionality can be removed should we ever manage to do #1182

brian-learningpool:
@moloko, I noticed this issue was still open.  Are we ok to close?

PR #1371, opened 2017-01-05
adds ability to disable accessibility for touch devices

moloko:
as per #1370


moloko:
@tomgreenfield yeah I suppose it should be

Issue #1380, opened 2017-01-16
add article/block/component id as a data attribute


moloko:
The article/block/component id is already being added to the page as a class which is useful for styling but little else.

Adding this as a data attribute would make it easy to access. Something like `data-adapt-id` would be suitable.

PR #1381, opened 2017-01-17
add a data-adapt-id attribute to the HTML for pages, menus, articles, blocks + components

moloko:
see #1380

also switch to AMD style module definitions, stop using shortcut paths

move concatenation operators from beginning of lines to end of lines as JSHint's recommendation

### DIFF
--- a/src/core/js/views/blockView.js
+++ b/src/core/js/views/blockView.js
@@ -4,12 +4,6 @@ define([
 
     var BlockView = AdaptView.extend({
 
-        attributes: function() {
-            return {
-                "data-adapt-id": this.model.get('_id')
-            };
-        },
-
         className: function() {
             return "block " + 
             this.model.get('_id') +
--- a/src/core/js/views/blockView.js
+++ b/src/core/js/views/blockView.js
@@ -1,16 +1,22 @@
 define([
     'core/js/views/adaptView'
-    ], function(AdaptView) {
+], function(AdaptView) {
 
     var BlockView = AdaptView.extend({
 
+        attributes: function() {
+            return {
+                "data-adapt-id": this.model.get('_id')
+            };
+        },
+
         className: function() {
-            return "block "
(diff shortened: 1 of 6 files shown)


src/core/js/views/blockView.js
```diff
@@ -1,16 +1,22 @@
 define([
     'core/js/views/adaptView'
-    ], function(AdaptView) {
+], function(AdaptView) {
 
     var BlockView = AdaptView.extend({
 
+        attributes: function() {
```
oliverfoster: this should not been needed as view extends adaptView

oliverfoster: Same in other occurrences.

moloko: @oliverfoster i didn't change adaptView so it wouldn't inherit this from there. or are you saying you think i 'should' put this in adaptView and not in its subclasses?

oliverfoster: I thought you had. But alas it's the articleView. Perhaps it would be better in one place if you're able and it works?

moloko: I'll give this a try tomorrow

oliverfoster: Aight. It is late.

moloko: You're quite right, better to have in adaptView. For some reason I thought adaptView was inherited more widely than it actually is.

oliverfoster: yeah, i always think that for some reason. silly really, it's such a very specific class definition.

moloko: I don't think it's ever actually instantiated, so AbstractView would have been a better name for it

oliverfoster: In my rewrites it always ends up being a generic container view in place of page, article and block. I then tend to have a distinct menu view and a component view. So... 3 classes instead of 5 and an abstract.

Issue #1384, opened 2017-01-19
Adding a class when Accessibility - Enable text reader support is set to true


ciaranlp:
When 'Enable accessibility?’ is set to true a class of .accessibility is added to html this is fine but when 'Enable text reader support?’ is set to true I’m getting a peculiar behaviour on a menu item which has a hover effect for the body text, can a class be added to html for example .textreader when this is set to true?

moloko:
done in https://github.com/adaptlearning/adapt_framework/commit/186056f92120bd60b7491f1a49bc8b3e3438cb60

PR #1385, opened 2017-01-19
Adds a 'text-to-speech' class when _isTextProcessorEnabled is true

brian-learningpool:


brian-learningpool:
This resolves https://github.com/adaptlearning/adapt_framework/issues/1384

PR #1390, opened 2017-01-23
Issue 1388

ciaranlp:
add the ability to skip to menu or page main content bypassing the navigation links

https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-skip.html

### DIFF
--- a/src/core/js/router.js
+++ b/src/core/js/router.js
@@ -18,6 +18,7 @@ define([
             });
             this.listenTo(Adapt, 'navigation:backButton', this.navigateToPreviousRoute);
             this.listenTo(Adapt, 'navigation:homeButton', this.navigateToHomeRoute);
+            this.listenTo(Adapt, 'navigation:skipNavigation', this.skipNavigation);
             this.listenTo(Adapt, 'navigation:parentButton', this.navigateToParent);
             this.listenTo(Adapt, "router:navigateTo", this.navigateToArguments);
         },
@@ -204,6 +205,20 @@ define([
                 break;
             }
             this.handleRoute.apply(this, args);
+        },
+
+        skipNavigation: function() {
+
+            var pageCount = $(".page").length;
+            var pageContent = $(".page-inner");
+            var menuContent = $(".menu");
+
+            if (pageCount == 1) {
+                pageContent.a11y_focus();
+            } else {
+                menuContent.a11y_focus();
+            }
+
         },
 
         navigateToPreviousRoute: function(force) {
--- a/src/core/js/router.js
+++ b/src/core/js/router.js
@@ -41,7 +41,7 @@ define([
             }
 
             if (Adapt.router.get('_canNavigate')) {
-                
+
                 //disable navigation whilst rendering
                 Adapt.router.set('_canNavigate', false, {pluginName: "adapt"});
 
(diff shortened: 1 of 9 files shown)


src/core/js/router.js
```diff
@@ -206,6 +207,20 @@ define([
...
+            var pageCount = $(".page").length;
+            var pageContent = $(".page-inner");
+            var menuContent = $(".menu");
+
+            if (pageCount == 1) {
+                pageContent.a11y_focus();
+            } else {
+                menuContent.a11y_focus();
+            }
+
```
oliverfoster: this can be replaced with:

```
$("."+Adapt.location._contentType).a11y_focus();
```

oliverfoster: core template also needs updating as per pr to vanilla https://github.com/adaptlearning/adapt-contrib-vanilla/pull/201

src/core/templates/navigation.hbs
```diff
@@ -1,5 +1,8 @@
 {{! All links in the nav bar should have an attritbue of data-event, navigationView uses this to fire a global event}}
 <div class="navigation-inner clearfix" role="navigation" aria-label="{{_globals._accessibility._ariaLabels.navigation}}">
+    {{#if _accessibility._isSkipNavigationEnabled}}
+        <button class="skip-nav-link" aria-label="Skip navigation" data-event="skipNavigation" role="button">Skip Navigation</button>
```
brian-learningpool: It doesn't seem right that "Skip Navigation" is hard-coded twice here.  

We need to add a language string to the `_globals._accesibility` object, like `_accessibilityToggleTextOn` and another for the aria-label.

src/core/schema/course.model.schema
```diff
@@ -171,6 +171,17 @@
               ],
               "translatable": true
             },
+            "_skipNavigationText": {
```
tomgreenfield: Please remove underscore from `_skipNavigationText`.

src/core/templates/navigation.hbs
```diff
@@ -1,5 +1,8 @@
 {{! All links in the nav bar should have an attritbue of data-event, navigationView uses this to fire a global event}}
 <div class="navigation-inner clearfix" role="navigation" aria-label="{{_globals._accessibility._ariaLabels.navigation}}">
+    {{#if _accessibility._isSkipNavigationEnabled}}
+        <button class="skip-nav-link" data-event="skipNavigation" aria-label="{{_globals._accessibility._ariaLabels.skipNavigation}}">{{_globals._accessibility._skipNavigationText}}</button>
```
tomgreenfield: Please remove underscore from `_skipNavigationText`.

tomgreenfield: Sorry I missed this the first time around but we usually allow HTML with triple braces `{{{` `}}}` on the actual rendered text e.g. on [Box Menu buttons](https://github.com/adaptlearning/adapt-contrib-boxmenu/blob/master/templates/boxmenu-item.hbs#L15).

moloko:
sorry all, I've only just spotted this:

If we're going to have `_isSkipNavigationEnabled` enabled by default (which it currently is in both src/course/config.json and config.model.schema) then we also need to make sure there's a `skipNavigation` ARIA label in the default course.json.

Or we don't have it enabled by default in which case it doesn't matter so much if the ARIA label isn't present.

I don't mind which we go with.

ciaranlp:
hey guys a PR for @moloko last comment can be found here https://github.com/adaptlearning/adapt_framework/pull/1414

Issue #1399, opened 2017-01-27
Add logging system


ryanrolds:
It's popular to use `console.log()` in web development. The problem is that unless developers are careful to remove to remove all their `console.log()` calls we can quickly end up with console spam. In many cases simply removing the console.log calls is desired, but a consistent way for the framework and plugins to log a fatal problems, errors, warnings, and debug information would be helpful to course authors, developers, and operations.

A logging model could have functions for reporting `fatal`, `error`, `warn`, `info`, and `debug` events and details about the event. If these events are then emitted as a traditional event (`log:fatal`, `log:warn`, etc...) plugins that wish to receive these events and relay them to a central logging system can. Developers can use `debug` and `info` level events to add debugging and publish their plugins without having to worry about log spam. Course developers can disable or adjust the log level so that only high severity issues are written to their consoles and/or reported to log forwarder plugins. In the event that a course author experiences a problem they can lower the log level and see debug or info logs which will help improve bug reports.

I've done a little work on this and when I have the code in working state I will post a link.

moloko:
funnily enough @chris-steele was talking to me just the other day about whether we should have something like this. we certainly would be interested to see this.

ryanrolds:
I'm running in to issues with the load order. It's desirable to get logging up as early as possible (minimize the parts of the system that can't use it). One problem I'm hitting is that plugins are starting to load before ConfigModel has finished loading the configs. So, if we want the main config to control what the logging module relays to forwarders or is written to console log, it really needs to load before the plugins. I have a working logging module and it uses reasonable defaults so that it can be available before the config loading completes and while the plugins are doing their thing.

A lot of this could be avoided if app.js (or some other init/bootstrap code) loaded the configs followed by setup of the logging module and Adapt. Once Adapt is in a sold state it fires off plugin loading. I've been able to delay the plugin loading (and still have them included in adapt.min.js), but I'm seeing a lot of plugins trying to hook in to earlier events and missing them because of the delayed loading. Ensuring the configs are loaded and Adapt is ready before the plugins are let loose would probably simplify plugin bootstrapping among other things.

oliverfoster:
Put it in the core alongside offlineStorage. We've got to change the fundamental architecture of adapt + bower to get loader orders installation etc.

ryanrolds:
[...]

PR #1405, opened 2017-01-30
Logging module

ryanrolds:
Adds logging module discussed in [#1390](https://github.com/adaptlearning/adapt_framework/issues/1399).

### DIFF
--- a/src/core/js/logging.js
+++ b/src/core/js/logging.js
@@ -1,78 +1,92 @@
 define(['core/js/adapt'], function(Adapt) {
-  var Logging = Backbone.Model.extend({
-    _config: {
-      _isEnabled: true,
-      _level: 'info',
-      _console: true,
-    },
-    _override: false,
-    _levels: {
-      debug: 0,
-      info: 1,
-      warn: 2,
-      error: 3,
-      fatal: 4
-    },
-    initialize: function() {
-      // Override default log level with level present in query string
-      var matches = window.location.search.match(/[?&]loglevel=([a-z]*)/i);
-      if (matches && matches.length >= 2 && this._levels[matches[1]] >= 0) {
-        this._config._level = matches[1];
-        this._override = this._config._level;
-        this.debug('Loglevel override in query string:', this._config._level);
-      }
+    var Logging = Backbone.Model.extend({
+        _config: {
+            _isEnabled: true,
+            _level: 'info', // Default log level
+            _console: true, // Log to console
+        },      
 
-      Adapt.once('configModel:loadCourseData', this.onLoadCourseData.bind(this));
-    },
-    onLoadCourseData: function() {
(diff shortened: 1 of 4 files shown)


src/core/js/logging.js
```diff
@@ -0,0 +1,78 @@
+define(['core/js/adapt'], function(Adapt) {
```
oliverfoster: could you put some whitespace into this file?

should be spaces 4 indentation, line breaks between function definitions etc

src/core/js/logging.js
```diff
@@ -0,0 +1,78 @@
...
+
+      // Query string log level wins over the config
+      if (this._override) {
+        this._config._level = this._override;
+      }
+
+      this.debug('Logging config loaded');
+      this.trigger('log:ready');
+    },
+    debug: function(msg, data) {
```
oliverfoster: these functions need their parameters to behaviour more like the ``console.log`` function.
essentially each argument forms part of the output string.

```javascript
    fatal: function() {
      this._log('fatal', [].array.slice.call(arguments, 0)));
    },
    _log: function(level, datas) {
      if (!this._config._isEnabled) {
        return;
      }

      if (this._levels[level] < this._levels[this._config._level]) {
        return;
      }

      if (this._config._console) {
        var log = [level.toUpperCase() + ':'];
        data && log.push.apply(log, datas);

        console.log.apply(console, log);
      }

      // Allow error reporting plugins to hook and report to logging systems
      this.trigger('log:' + level, datas);
    }
```

ryanrolds: Will do.

src/core/js/logging.js
ryanrolds: The 4 for indentation conflicts with: https://github.com/adaptlearning/documentation/blob/master/01_cross_workstream/style_guide.md which is linked in https://github.com/adaptlearning/adapt_framework/wiki/Contributing-code

oliverfoster: Ignore the referenced/copied standard, it hasn't been formalized yet. It should say "spaces 4" indentation, no tabs, etc.

ryanrolds:
* Fixed the indentation
* Adjusted log function arguments
* Query string override is now case-insensitive
* Added some blank lines
* Comments

Should be ready to review again.

ryanrolds:
IE8 has been EOL for a while now and as far as I'm aware is not supported on current MS OSs.

ryanrolds:
For ethical reasons I don't make changes to specifically support IE8, the browser is EOL and no longer receiving security updates, it's usage is negligent.

oliverfoster:
Fair enough. I'll see if someone else can take over.

oliverfoster:
functionality added on #1425 

thanks for doing this @ryanrolds 
nice bit of code

PR #1411, opened 2017-02-03
Updated content based on internal user testing.

danielghost:


src/course/en/articles.json
```diff
@@ -47,8 +47,14 @@
 			"_attempts": "infinite"
 		},
 		"_trickle": {
-			"_isEnabled": true
-		}
+			"_isEnabled": true,
+            "_button": {
```
moloko: indentation issues here

src/course/en/components.json
```diff
@@ -572,11 +563,15 @@
...
 		"displayTitle": "Slider",
-		"body": "Working memory is thought to be responsible for our ability to temporarily hold and manipulating information but according to a paper by Miller (1956) what ‘magic number’ describes its capacity?",
+		"body": "Working memory is thought to be responsible for our ability to temporarily hold and manipulate information, but according to a paper by Miller (1956) what ‘magic number’ describes its capacity?",
 		"instruction": "Drag the slider to make your choice and select Submit.",
 		"labelStart": "",
 		"labelEnd": "",
-		"_correctAnswer": "7",
+		"_correctAnswer": "",
+		"_correctRange": {
+			"_bottom": 7,
```
moloko: shouldn't this be 5? Feedback says 'According to Miller’s paper 7 +/- 2 chunks of information was the limited capacity of working memory'

src/course/en/components.json
```diff
@@ -572,11 +563,15 @@
...
+			"_top": 7
```
moloko: shouldn't this be 9?

src/course/en/course.json
```diff
@@ -89,14 +89,17 @@
...
-                "ipad": "Usage instructions for touchscreens. Use swipe right for next. Use swipe left for previous. Use a double tap to select. Use a two finger slide up to go to the top of the page. Usage instructions for keyboard access. Use right for next. Use left for previous. Use up and down together to select."
+                "ipad": "Usage instructions for touchscreens. Use swipe left for next. Use swipe right for previous. Use a double tap to select. Use a two finger slide up to go to the top of the page. Usage instructions for keyboard access. Use left for next. Use right for previous. Use up and down together to select."
             },
             "_ariaLabels": {
                 "navigation": "Course navigation bar",
                 "menuLoaded": "Menu Loaded.",
                 "menu": "Menu",
+                "menuItem": "Menu item.",
+                "menuViewButton": "Select here to enter.",
+                "menuEnd": "You have reached the end of the menu.",
```
tomgreenfield: `menuItem` and `menuEnd` already appear in the Box Menu globals (lines 84 & 85), which are in turn pulled in automatically from the [properties schema](https://github.com/adaptlearning/adapt-contrib-boxmenu/blob/master/properties.schema).

Unsure where `menuViewButton` has cropped up from?

oliverfoster: that's a weird one ``menuViewButton`` I can't find any reference to it in any of the menu's I've got. but it is in the course.model.schema

moloko: Can confirm it's definitely the `_boxmenu` versions of `menuItem` and `menuEnd` that are used in the output - and that `_ariaLabels.menuViewButton` isn't being used.

moloko:
also looking at the code (resources.hbs) I think `_ariaLabels.resourcesEnd` should actually be `_extensions._resources.resourcesEnd`

Issue #1429, opened 2017-02-21
Add enumeration support to the environment


oliverfoster:
Good for use in places like:

buttonsView states: https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/buttonsView.js#L99

logging levels: https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/logging.js#L10

moloko:
death to magic strings!

PR #1430, opened 2017-02-21
issue-1429: added enumeration support

oliverfoster:
fixes issue: https://github.com/adaptlearning/adapt_framework/issues/1429

library from: https://github.com/cgkineo/enum

### DIFF
--- a/src/core/js/scriptLoader.js
+++ b/src/core/js/scriptLoader.js
@@ -31,7 +31,8 @@
                 inview: 'libraries/inview',
                 a11y: 'libraries/jquery.a11y',
                 scrollTo: 'libraries/scrollTo',
-                bowser: 'libraries/bowser'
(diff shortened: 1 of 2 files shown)


src/core/js/scriptLoader.js
```diff
@@ -31,7 +31,8 @@
                 inview: 'libraries/inview',
                 a11y: 'libraries/jquery.a11y',
                 scrollTo: 'libraries/scrollTo',
-                bowser: 'libraries/bowser'
+                bowser: 'libraries/bowser',
+                'enum': 'libraries/enum'
```
moloko: should 'enum' have quotes? none of the previous lines do

oliverfoster: It's an ie8 reserved keyword

Issue #1417, opened 2017-02-09
adapt-contrib-languagePicker: Missing aria-label in navigation button


brian-learningpool:
The language picker button in the header navigation is missing an `aria-label`.

moloko:
done in https://github.com/adaptlearning/adapt-contrib-languagePicker/commit/61890ec3415f1c3271963af37804710a85e97b14

PR #1431, opened 2017-02-21
add ariaLabel for language picker to course.json and course.model.schema

moloko:
see #1417


moloko:
@oliverfoster ah OK that answers [this question] then... yes, more than happy to change it as it'll negate the need for this PR and a further one to amend these schemas in the AT.

PR #1460, opened 2017-03-09
Resolves #1458, corrects various issues with router.js

### LINKED ISSUE #1458: Miscellaneous issues with router.js
It's been some time since I've looked at `router.js` but while trying to develop a plugin which would make use of the plugin routing I noticed a few problems:
- `arguments` being parsed as `args` but raw `arguments` are still passed to functions
- The code assumes that only the last argument is may be `null`, this is not necessarily the case
- Plugin routing via the `handlePluginRouter` function does not appear to reset the `_canNavigate` property in the same way the `handleId` function does
- `_isCircularNavigationInProgress` is undefined until it's set, the code is going into the block on line 48-51 when it should not be

brian-learningpool:
- arguments were being parsed as args but raw arguments are still passed to functions
- The code assumed that only the last argument may be null, this is not necessarily the case
- Plugin routing via the handlePluginRouter function did not appear to reset the _canNavigate property in the same way the handleId function does
- Tidied the check on _isCircularNavigationInProgress

### DIFF
--- a/src/core/js/router.js
+++ b/src/core/js/router.js
@@ -31,11 +31,21 @@ define([
 
         handleRoute: function() {
             var args = [].slice.call(arguments, 0, arguments.length);
-            if (arguments[arguments.length-1] === null) args.pop();
+
+            if (args.length) {
+                // Remove any null arguments, starting from the last.
+                for (var i = (args.length - 1); i >= 0; i--) {
+                    if (args[i] === null) {
+                        args.pop();
+                    } else {
+                        break;
+                    }
+                }
+            }
 
             //check if the current page is in the progress of navigating to itself
             //it will redirect to itself if the url was changed and _canNavigate is false
-            if (!this._isCircularNavigationInProgress) {
+            if (this._isCircularNavigationInProgress === false) {
                 //trigger an event pre 'router:location' to allow extensions to stop routing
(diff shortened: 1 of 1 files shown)


src/core/js/router.js
```diff
@@ -31,11 +31,21 @@ define([
 
         handleRoute: function() {
             var args = [].slice.call(arguments, 0, arguments.length);
-            if (arguments[arguments.length-1] === null) args.pop();
+
+            if (args.length) {
+                // Remove any null arguments, starting from the last.
+                for (var i = (args.length - 1); i >= 0; i--) {
```
moloko: as mentioned you could probably use `_.without` here to exclude null arguments in just one line of code - but if you want to leave it as-is I'm fine with that

src/core/js/router.js
```diff
@@ -31,11 +31,21 @@ define([
 
         handleRoute: function() {
             var args = [].slice.call(arguments, 0, arguments.length);
```
moloko: I know you didn't touch this line, but wondered if you thought 
```
var args = _.toArray(arguments);
```
would look neater?

brian-learningpool: It looks much neater.  I'll update.

Issue #1463, opened 2017-03-10
grunt copy tasks: assets folders are flattened when copied


LukaszGrela:
When in the extension/component or menu assets folder I will put the resources in the folder tree then this structure is not kept but flattened. Resources and all folders are copied into the build/assets folder. It should keep the folder structure when copying.

moloko:
done in 32c51d8c8ed1c80a9e0b69f696a4b4e18c82ec46

PR #1464, opened 2017-03-10
Resolution to the #1463

LukaszGrela:
Extending the solution used for 'required' and 'libraries' I've modified the copy routines for assets in components, extensions and menus. The regex used prevents the folder assets itself to be copied into the build assets.

### DIFF
--- a/grunt/config/copy.js
+++ b/grunt/config/copy.js
@@ -1,4 +1,22 @@
 module.exports = function (grunt, options) {
+    
+    
+    var renameAssets = function (destFolder, srcFileName) {
+        var collateAtName = "assets";
+        var collateAtFolder = collateAtName + "/";
+        var startOfCollatePath = srcFileName.indexOf(collateAtFolder) + collateAtFolder.length;
+        var collatedFilePath = destFolder + srcFileName.substr(startOfCollatePath);
+        //ignore the folder alone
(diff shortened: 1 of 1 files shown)


grunt/config/copy.js
```diff
@@ -53,7 +53,20 @@ module.exports = function (grunt, options) {
                     filter: function(filepath) {
                         return grunt.config('helpers').includedFilter(filepath);
                     },
-                    flatten: true
+                    
+                    rename: function(destFolder, srcFileName) {
+                        var name = "assets",
```
oliverfoster: Can you:

Move the function to the top of the file and then reference it where needed, rather than write it multiple times.
``var`` for each line rather than commas.
Rename ``name`` to ``collateAtName``  
Rename ``folder`` to ``collateAtFolder``  
Rename ``endOfFolder`` to ``startOfCollatePath``  
Rename ``re`` to ``testEndsWithCollateName``  
Rename ``output`` to ``collatedFilePath``  
(Or something more descriptive for each one than we've currently got.)

LukaszGrela: ok, I'll do it in my nearest free time:)

oliverfoster: thanks man :)

PR #1476, opened 2017-03-20
Turn browserstring into valid css classes

lc-thomasberger:
- replace "." with "-"
- convert to lowercase

Old syntax kept in to maintain backwards compatibility.


oliverfoster:
can you keep the old syntax as well so it's not breaking?

lc-thomasberger:
sure @oliverfoster

brian-learningpool:
Is the old syntax actually required?  As far as I'm aware CSS is case insensitive so `os-windows` should work fine.  For version-56.0 to actually have ever worked it would have required some backslash escaping in the CSS.

tomgreenfield:
+1 for backwards compatibility @oliverfoster, I have been using (escaped) version classes in CSS for various plugins.

oliverfoster:
I'm not sure we can remove supported syntax for selectors without notifying the community.

Issue #1484, opened 2017-03-23
start controller adds to history


oliverfoster:
it should replace the current location rather than navigation forward

PR #1485, opened 2017-03-23
issue-1484: start controller replace history

oliverfoster:
#1484

### DIFF
--- a/src/core/js/startController.js
+++ b/src/core/js/startController.js
@@ -13,7 +13,16 @@ define([
         setStartLocation: function() {
             if (!this.isEnabled()) return;
 
-            window.location.hash = this.getStartHash();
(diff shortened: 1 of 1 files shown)


src/core/js/startController.js
```diff
@@ -13,7 +13,16 @@ define([
...
             if (!this.isEnabled()) return;
 
-            window.location.hash = this.getStartHash();
+            var hash = this.getStartHash();
+
+            if ('replaceState' in window.history) {
+                window.history.replaceState('', '', hash);
+            } else {
+                // < ie9
+                if (window.location.hash !== hash) window.history.back();
```
moloko: my only comment would be that this line needs a bit more explanation than just '< IE9'

Issue #1487, opened 2017-03-24
_isOptional completion


danielghost:
If a parent model is set to _isOptional AND its immediate children are also set to be _isOptional, simply launching a model with the same parent will mark the other optional models as completed.

This issue is dependant on the course structure and is less likely since https://github.com/adaptlearning/adapt_framework/commit/61ebb6bfdbd5a0a1c2670f099e0a1be4ed33378e, but it is still very possible.

moloko:
done in framework 2.1.0

moloko:
I think now we have the option to do component-level tracking ids this may no longer be relevant?

olivermartinfoster:
What is the outstanding issue with this one? It looks/sounds to be fixed.

I recollect two old problems:
1. That `_isOptional` would mark `_isComplete` on children when completion calculations were performed for any child, which I don't think it does that now. 
2. Block level completion restoration used to mark block children complete when the block was complete - which I'm not sure that had any interplay with `_isOptional`.

Dan seems to have sorted issue 1 a while back and the state restoration in spoor shouldn't have an impact on completion now except at the component level. Each entry is always stored at component level now, regardless of whether tracking ids are on the article, block or component. It uses the tracking id + an nth-descendant index, so that each component entry starts with a small header container the component's relative descendant offset to a specified tracking id.

moloko:
> What is the outstanding issue with this one? It looks/sounds to be fixed.

I agree, I'm just checking before closing

moloko:
@danielghost feel free to re-open if you think there's still an issue...

danielghost:
Seems the required amend for this never got merged. Reopening as it has been raised for a recent project where models are set to `_isOptional` via role selection.

Issue #1486, opened 2017-03-24
Locking doesn't account for optional models


danielghost:
This essentially means they can become mandatory.

moloko:
@danielghost could you give a bit more background/detail on this one?

danielghost:
Previously the locking condition was just skipping over any unavailable models, as these don't need to be included when working out if it they are included as a requirement for a particular model being locked. We should also skip over any of the models in the list which are optional, otherwise these become mandatory in regards to unlocking the model being checked.

moloko:
@danielghost is this still relevant or can it be closed now?

Issue #695, opened 2015-08-12
adaptModel::setupChildListeners listening to ALL child types


chris-steele:
This function should presumably set up listeners on its own children? At the moment each page model listens to changes on ALL article models, each article listens to changes on ALL block models etc.

danielghost:
This was addressed in our fork a while ago, but I guess it didn't make its way back into adaptlearning during the v2 amends:
https://github.com/cgkineo/adapt-framework-kineoFork/blob/master/js/adapt-framework-kineoFork.js#L101-L109

moloko:
is this still relevant?

chris-steele:
Yeah, still the same issue in v2. The handlers for the change events all check the actual children (using getChildren), but this still seems like it should be fixed even if just for code readability. I'm not aware of it ever causing issues, but it would boost performance if nothing else. Happy to leave as is though.

PR #1488, opened 2017-03-24
Suggested amend for child listeners. Thought this had been implemente…

danielghost:
…d a long time ago, but obviously not. #695.

Was added and tested alongside #1486 and #1487 on a large complex project.

### DIFF
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -55,15 +55,17 @@ define(function (require) {
         },
 
         setupChildListeners: function() {
+            var children = this.getChildren();
 
-            if (!this.getChildren()) return;
+            if (!children) return;
 
-            this.listenTo(Adapt[this._children], {
-                "change:_isReady": this.checkReadyStatus,
-                "change:_isComplete": this.onIsComplete,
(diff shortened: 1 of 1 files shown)


src/core/js/models/adaptModel.js
```diff
@@ -55,15 +55,17 @@ define(function (require) {
...
-            if (!this.getChildren()) return;
-
-            this.listenTo(Adapt[this._children], {
-                "change:_isReady": this.checkReadyStatus,
-                "change:_isComplete": this.onIsComplete,
-                "change:_isInteractionComplete": this.checkInteractionCompletionStatus
-            });
+            if (!children) return;
 
+            children.each(function(child) {
```
oliverfoster: Can you add this in? 
```javascript
                this.stopListening(child, {
                    "change:_isReady": this.checkReadyStatus,
                    "change:_isComplete": this.onIsComplete,
                    "change:_isInteractionComplete": this.checkInteractionCompletionStatus
                });
```
So that if we change the children we can get the model to re-listen properly?

oliverfoster: @danielghost ?

Issue #1040, opened 2016-05-03
Feature request: pull data from SCORM


hmaffin:
Desperate to have my courses pull usernames, emails. names, whatever it can get, from SCORM, such as Moodle, to both personalise and validate within the course.

Uses could include in course certificates (Moodle's are rubbish!), emailing out (see second feature request), personalising learner experience and more!

moloko:
Unfortunately it's generally just the user's name that SCORM makes available (weirdly always in military/private school-style format i.e. "Lastname, Firstname") but I think this is a great idea, it would allow for some nice personalisation of the course content to be able to refer to the user by name.

Ideally this would be done in a way that allowed it to be specified in different formats and from anywhere in the JSON e.g. 
```
body: "Hi there {{student_firstname}} and welcome to this course on data protection"
```

Big +1 from me for this feature request.

moloko:
First stage of this is complete in that the spoor extension is now able to pull in what information SCORM makes available about the user - next stage is to make it easy to use that information within Adapt.

I think we need to get this data into the `_globals` object (with some sensible default for when testing without an LMS) then set up some handlebars helpers to allow that data to be inserted easily into the .json

hmaffin:
Amazing! You really can just use {{firstname}} in the AT's text boxes to display the info (after that funky code you gave me went in the right place). 

Great stuff Matt!

Deathjam:
Hi i am too looking to include this data, what is this funky code i need?

brian-learningpool:
@moloko, should we add a `_globals.user` object, or what had you in mind?

oliverfoster:
can we have an adapt api for this? something like:

```javascript
Adapt.global("propertyName", "value"); //setter
Adapt.global("propertyName"); //getter
```
```handlebars
{{global "propertyName"}}
```

moloko:
I think it needs to be something in `_globals` so that we can set default values for when there's no LMS available

oliverfoster:
Sounds fiine to me, as long as we're happy using the current globals object, we just need to expose it in the handlebars and make a quicker reference to it in on the Adapt object.

```javascript
Handlebars.registerHelper("include_globals", function(data) {
  data.root._globals = Adapt.course.get("_globals");
  return "";
}) ;
```
and
```javascript
Adapt.globals = Adapt.course.get("_globals");
```

?

moloko:
[...]

PR #1491, opened 2017-03-27
updates to allow json to include handlebars style expressions

moloko:
This allows for course authors to use these in their JSON to display data from the models that are available to the template, specifically it's to allow for the functionality described in #1040 whereby users can pull in data from SCORM by using expressions like `{{_globals._learnerInfo.firstname}}` in their JSON, giving you the option to do things like this:

2017-03-27 18_14_59-adapt version 2 0 demonstration _ presentation components

It also introduces a new way of getting the `_globals` object into handlebars templates - via an `{{import_globals}}` expression at the top of each template - meaning we can get rid of [this line in adaptView.js](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/adaptView.js#L16)... only question is whether we do that immediately or leave it in place for a bit to cover people who use the latest version of the framework with a old version of a theme..?

Note that I will also be updating adapt-contrib-spoor to get it to replace the default values in `_globals._learnerInfo` with the actual data from the LMS.

I will also need to amend the copies of the handlebars templates that are stored in adapt-contrib-vanilla.


src/core/js/views/adaptView.js
```diff
@@ -13,7 +13,6 @@ define([
         initialize: function() {
             this.listenTo(Adapt, 'remove', this.remove);
             this.listenTo(this.model, 'change:_isVisible', this.toggleVisibility);
-            this.model.set('_globals', Adapt.course.get('_globals'));
```
oliverfoster: removing this line is going to break all existing themes

moloko: hey, it was your suggestion ;-)

oliverfoster: :P if we leave it in place for the moment, it can come out in v3

oliverfoster:
The ``{{import_globals}}`` can be used in any handlebars, not just those which belong to a view that extend ``AdaptView``, which is a nice change. It doesn't matter so much that the ``_globals`` object will be included in the ``handlebars`` data context twice - once in ``AdaptView`` and once from the global helper. Not breaking the themes I would say is more important for the moment.

oliverfoster:
I would like to remove the templates from the theme.

PR #1499, opened 2017-03-31
proposed new icons for vanilla

ciaranlp:
A selection of proposed new icons to add to vanilla, illustrated in image attached.

proposed-new-vanilla-icons


oliverfoster:
could you do a pr for the core instead? i'm trying to get the standard icon font out of the theme

ciaranlp:
@oliverfoster consider thee forgiven, it's Friday!!!

mevelevesque:
Hi, I try to figure it out how to add icons in my text. For exemple, if I want to add icon e910. How can I add this one in my paragraph?

moloko:
@mevelevesque this is a good question - would you mind asking it [on the forums](https://community.adaptlearning.org/mod/forum/view.php?id=4) so that others can benefit from seeing the answer?

mevelevesque:
Sure! Actually, I found a way, but maybe there is a better option. And I would really like to know if the Vanilla library contains an icon representing the progression bar.

What I found it to add a span within my paragraph with the inline display style. Ex.: <span class="icon icon-list" style="display:inline;">&nbsp;</span>

**And the span cannot be empty, so I added a space.

moloko:
@mevelevesque if you email Adapt.Learning@kineo.com with details of the name you used to register on the forums I'll set you up with forum post access (we disabled it by default to stop forum spam)

PR #1502, opened 2017-04-04
Added new ItemsModel to core

lc-thomasberger:
add core itemsModel

`_isActive` attribute is added to `_items` Array. This attributes are managed by funcitons defined in the itemsModel.
The setter functions trigger a change event `change_items:_isActive`. Views listen to this event and hande Dom updates. Plugins can use the functions on the itemsModel to manipulate the state of the Component.

also fixes:
#1505

### DIFF
--- a/src/core/js/models/itemsModel.js
+++ b/src/core/js/models/itemsModel.js
@@ -1,0 +1,37 @@
+define([
+    'core/js/adapt',
+    'core/js/models/componentModel'
+], function(Adapt, ComponentModel) {
+
+    var ItemsModel = ComponentModel.extend({
+
+        resetItems: function() {
+            _.each(this.model.get('_items'), function(item) {
+                item._isVisited = false;
+            });
+        },
+        
+        getItemCount: function() {
+            return this.get('_items').length;
+        },
+
+        getCurrentItem: function(index) {
+            return this.get('_items')[index];
+        },
+
+        getVisitedItems: function() {
+            return _.filter(this.get('_items'), function(item) {
+                return item._isVisited;
+            });
+        },
+
+        setItemAsVisited: function(index) {
+            var item = this.get('_items')[index];
+            item._isVisited = true;
+        }
+
+    });
+
+    return ItemsModel;
+
+});
--- a/src/core/js/models/itemsModel.js
+++ b/src/core/js/models/itemsModel.js
@@ -65,7 +65,7 @@ define([
         },
         
         getFirstActiveItem: function() {
-            return this.model.getActiveItems()[0];
+            return this.getActiveItems()[0];
         },
         
         getActiveItemsCount: function() {
--- a/src/core/js/models/itemsModel.js
+++ b/src/core/js/models/itemsModel.js
@@ -21,7 +21,7 @@ define([
             return this.get('_items').length;
         },
 
-        getItemAtIndex: function(index) {
+        getItem: function(index) {
             return this.get('_items')[index];
         },
 
@@ -31,7 +31,7 @@ define([
             });
         },
 
-        setItemAtIndexAsVisited: function(index) {
+        setItemVisited: function(index) {
             var item = this.get('_items')[index];
             if (item) {
                 item._isVisited = true;
@@ -91,7 +91,7 @@ define([
             return this.getActiveItemsIndexes()[0];
         },
 
-        setItemAtIndexAsActive: function(index, trigger) {
+        setItemActive: function(index, trigger) {
             var items = this.get('_items');
             var item = items[index];
             
@@ -105,7 +105,7 @@ define([
             return item;
         },
         
-        setItemAtIndexAsInactive: function(index, trigger) {
+        setItemInactive: function(index, trigger) {
             var items = this.get('_items');
             var item = items[index];
             
--- a/src/core/js/models/itemsModel.js
+++ b/src/core/js/models/itemsModel.js
@@ -5,9 +5,9 @@ define([
 
     var ItemsModel = ComponentModel.extend({
 
-        reset: function(type) {
+        reset: function(type, force) {
             this.resetItems();
-            ComponentModel.prototype.reset.call(this, type);
+            ComponentModel.prototype.reset.call(this, type, force);
         },
 
         resetItems: function() {
--- a/src/core/js/models/itemsModel.js
+++ b/src/core/js/models/itemsModel.js
@@ -39,7 +39,7 @@ define([
         },
 
         areAllIItemsCompleted: function() {
-            return (this.getVisitedItems().length == this.getItemCount());
+            return (this.getVisitedItems().length === this.getItemCount());
         },
 
         checkCompletionStatus: function() {
--- a/src/core/js/models/itemsModel.js
+++ b/src/core/js/models/itemsModel.js
@@ -20,7 +20,7 @@ define([
             return this.get('_items').length;
         },
 
-        getCurrentItem: function(index) {
+        getItemAtIndex: function(index) {
             return this.get('_items')[index];
         },
 
--- a/src/core/js/models/itemsModel.js
+++ b/src/core/js/models/itemsModel.js
@@ -47,14 +47,13 @@ define([
         },
 
         resetActiveItems: function(trigger) {
-            var _trigger = (trigger === undefined) ? true : trigger;
             var items = this.get('_items');
 
             _.each(items, function(item) {
                 item._isActive = false;
             });
 
-            if (_trigger) this.trigger('change:_items:_isActive', this, items);
+            if (trigger !== false) this.trigger('change:_items:_isActive', this, items);
         },
 
         getActiveItems: function() {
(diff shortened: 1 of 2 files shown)


oliverfoster:
that looks really succinct! i'll check it out but off the top of my head, i love it already :D

src/core/js/models/itemsModel.js
```diff
@@ -5,8 +5,13 @@ define([
 
     var ItemsModel = ComponentModel.extend({
 
+        reset: function(type) {
+            this.resetItems();
+            ComponentModel.prototype.reset.call(this, type);
+        },
```
oliverfoster:
```
        reset: function(type, force) {
            this.resetItems();
            ComponentModel.prototype.reset.call(this, type, force);
        },
```
Sorry I missed out the second property last night.
This should also be added to the ItemsModel reset function.

src/core/js/models/itemsModel.js
```diff
@@ -0,0 +1,52 @@
...
+            _.each(this.get('_items'), function(item) {
+                item._isVisited = false;
+            });
+        },
+        
+        getItemCount: function() {
+            return this.get('_items').length;
+        },
+
+        getCurrentItem: function(index) {
```
brian-learningpool: Since `index` is a parameter, would this be better called `getItemAtIndex`?

To me something called `getCurrentItem` should return the item using an internal indexer, not something that must be passed in.

lc-thomasberger: Will update this @brian-learningpool

src/core/js/models/itemsModel.js
```diff
@@ -43,6 +44,61 @@ define([
...
+        
+        setItemAtIndexAsInactive: function(index, trigger) {
+            var _trigger = (trigger === undefined) ? true : trigger;
+            var items = this.get('_items');
+            var item = items[index];
+            
+            if (item === undefined) return false;
+
+            item._isActive = false;
+            if (_trigger) this.trigger('change:_items:_isActive', this, items);
```
oliverfoster: ```if (trigger)``` will work as undefined equals to false in an if statement, otherwise known as a "falsy" http://adripofjavascript.com/blog/drips/truthy-and-falsy-values-in-javascript.html;

src/core/js/models/itemsModel.js
```diff
@@ -0,0 +1,105 @@
...
+        },
+
+        resetActiveItems: function(trigger) {
+            var items = this.get('_items');
+
+            _.each(items, function(item) {
+                item._isActive = false;
+            });
+
+            if (trigger !== false) this.trigger('change:_items:_isActive', this, items);
```
brian-learningpool: Can you add the braces here for readability?

src/core/js/models/itemsModel.js
```diff
@@ -0,0 +1,105 @@
...
+        
+        getActiveItemsCount: function() {
+            return this.getActiveItems().length;
+        },
+        
+        getActiveItemsIndexes: function() {
+            var indexes = [];
+            var items = this.get('_items');
+            
+            for (var i = 0, l = items.length; i < l; i++)
```
brian-learningpool: It'd be more readable with braces on the `for` and the `if`.

src/core/js/models/itemsModel.js
```diff
@@ -0,0 +1,105 @@
...
+        },
+
+        setItemAtIndexAsActive: function(index, trigger) {
+            var items = this.get('_items');
+            var item = items[index];
+            
+            if (item === undefined) return false;
+
+            item._isActive = true;
+            if (trigger !== false) this.trigger('change:_items:_isActive', this, items);
```
brian-learningpool: Curly braces?

src/core/js/models/itemsModel.js
```diff
@@ -0,0 +1,105 @@
...
+        },
+        
+        setItemAtIndexAsInactive: function(index, trigger) {
+            var items = this.get('_items');
+            var item = items[index];
+            
+            if (item === undefined) return false;
+
+            item._isActive = false;
+            if (trigger !== false) this.trigger('change:_items:_isActive', this, items);
```
brian-learningpool: Braces missing?

oliverfoster:
ok. now we've got the outline of the API for ItemsModel, is there any way we can combine some of these functions to make it smaller? or rename/reorder them so that it is easier to group them? rename parameters so the are easy to understand? or are we all happy this set is what we need?

In Order:
```
reset: function(type, force);
resetItems: function();
getItemCount: function();
getItemCount: function();
getVisitedItems: function()
setItemAtIndexAsVisited: function(index);
getCompletionStatus: function();
checkCompletionStatus: function();
resetActiveItems: function(trigger);
getActiveItems: function();
getFirstActiveItem: function();
getActiveItemsCount: function();
getActiveItemsIndexes: function();
getFirstActiveItemIndex: function();
setItemAtIndexAsActive: function(index, trigger);
setItemAtIndexAsInactive: function(index, trigger);
```

src/core/js/models/itemsModel.js
```diff
@@ -0,0 +1,124 @@
...
+
+        getVisitedItems: function() {
+            return _.filter(this.get('_items'), function(item) {
+                return item._isVisited;
+            });
+        },
+
+        setItemAtIndexAsVisited: function(index) {
+            var item = this.get('_items')[index];
+            item._isVisited = true;
```
moloko: Would this benefit from a quick check that `item` isn't `undefined`?

src/core/js/models/itemsModel.js
```diff
@@ -0,0 +1,124 @@
...
+        },
+
+        getCompletionStatus: function() {
```
moloko: This function name is a bit misleading, something like `checkAllItemsCompleted` seems a bit more descriptive

moloko: or `areAllIItemsCompleted`?

src/core/js/models/itemsModel.js
```diff
@@ -0,0 +1,124 @@
...
+            return (this.getVisitedItems().length == this.get('_items').length);
```
moloko: Change `this.get('_items').length` to `this.getItemCount()`?

src/core/js/models/itemsModel.js
```diff
@@ -0,0 +1,124 @@
...
+            return this.getActiveItemsIndexes()[0];
+        },
+
+        setItemAtIndexAsActive: function(index, trigger) {
+            var items = this.get('_items');
+            var item = items[index];
+            
+            if (item === undefined) return false;
+
+            item._isActive = true;
```
oliverfoster: probably need to include:
```javascript
item._isVisited = true;
```

src/core/js/models/itemsModel.js
oliverfoster: Dunno about this. I can see the function name ``getCompletionStatus`` transcending the ``ItemsModel`` container as it doesn't really have to be related to the items abstract - except in this case. Then again, the behaviour the function contains is more aptly described by ``areAllItemsCompleted`` / ``getAllItemsCompleted``... so perhaps it is a good idea?

src/core/js/models/itemsModel.js
lc-thomasberger: `item._isVisited` is handled by functions like `setItemAtIndexAsVisited` called in the views.

oliverfoster: I understand. It just seems a little odd to me to have something set active but not visited.
Why have two function calls when one will suffice?

oliverfoster: Could something be active and not marked as visited.

lc-thomasberger: Narratives on mobile can have an active item. Visited is only set when notify is opened.

oliverfoster: 👍 cool

src/core/js/models/itemsModel.js
```diff
@@ -0,0 +1,126 @@
...
+            }
+
+            return indexes;
+        },
+
+        getFirstActiveItemIndex: function() {
+            return this.getActiveItemsIndexes()[0];
+        },
+
+        setItemAtIndexAsActive: function(index, trigger) {
```
oliverfoster: ```setItemIndexActive```
```setItemIndexVisited```

should we drop the ``At`` and ``As``?

Comment (lc-thomasberger) [inline src/core/js/models/itemsModel.js]: fine for me

Comment (oliverfoster) [inline src/core/js/models/itemsModel.js]: the shorter ones would be my preference, i was using this today and it seemed a bit weird having ``getItemIndex`` or ``setItemAtIndexAsActive`` rather than ``setItemActive``, which is nice and short.

File: src/core/js/models/itemsModel.js [hunk 13]
```diff
@@ -0,0 +1,126 @@
...
+
+        setItemVisited: function(index) {
+            var item = this.get('_items')[index];
+            if (item) {
+                item._isVisited = true;
+            }
+        },
+
+        areAllIItemsCompleted: function() {
+            return (this.getVisitedItems().length == this.getItemCount());
```
moloko: Could you change this to use the strict equality operator?

PR #1506, opened 2017-04-05
Added 'ie' class for Internet Explorer 10 and 11

brian-learningpool:
This is added in addition to the 'internet explorer' class.

### DIFF
--- a/src/core/js/device.js
+++ b/src/core/js/device.js
@@ -187,10 +187,16 @@ define(function(require) {
         }
     }
 
-    // Convert 'msie' to 'ie' for backwards compatibility
-    var browserString = (Adapt.device.browser.toLowerCase() === 'msie') ? 'ie' : Adapt.device.browser.toLowerCase();
+    var browser = Adapt.device.browser.toLowerCase();
+    // Convert 'msie' to 'ie' for backwards compatibility.
+    var browserString = (browser === 'msie') ? 'ie' : browser;
     browserString = browserString + ' version-' + Adapt.device.version + ' OS-' + Adapt.device.OS + ' ' + getAppleDeviceType();
     browserString += browserString.replace('.', '-').toLowerCase();
+
+    // Ensure IE 10 and IE 11 also have the 'ie' class.
+    if (browser === 'internet explorer') {
+        browserString += ' ie';
+    }
 
     $("html").addClass(browserString + ' pixel-density-' + pixelDensity());
 });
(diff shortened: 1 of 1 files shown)


src/core/js/device.js
```diff
@@ -187,10 +187,16 @@ define(function(require) {
         }
     }
 
-    // Convert 'msie' to 'ie' for backwards compatibility
-    var browserString = (Adapt.device.browser.toLowerCase() === 'msie') ? 'ie' : Adapt.device.browser.toLowerCase();
+    var browser = Adapt.device.browser.toLowerCase();
+    // Convert 'msie' to 'ie' for backwards compatibility.
+    var browserString = (browser === 'msie') ? 'ie' : browser;
```
oliverfoster:
```javascript
switch(browser) {
    case "msie": case "internet explorer":
        browserString = "ie";
        break;
    case:
        browserString = browser;
}
```

or

```javascript
var browserString = browser.replace(/^msie$|^internet explorer$/gi, "ie");
```

brian-learningpool: Wouldn't the `switch` statement also add the `msie` class though?

oliverfoster:
```javascript
var browserString = browser;
switch(browserString) {
    case "msie": case "internet explorer":
        browserString = " ie";
}
```

or

```javascript
var browserString = browser.replace(/^msie$|^internet explorer$/, " ie" );
```
As per tom's latest suggestion on the original issue.

brian-learningpool: OK, apologies, I missed Tom's comment about disregarding the `internet explorer` class.

src/core/js/device.js
```diff
@@ -187,8 +187,9 @@ define(function(require) {
         }
     }
 
-    // Convert 'msie' to 'ie' for backwards compatibility
-    var browserString = (Adapt.device.browser.toLowerCase() === 'msie') ? 'ie' : Adapt.device.browser.toLowerCase();
+    var browser = Adapt.device.browser.toLowerCase();
+    // Convert 'msie' and 'internet explorer' to 'ie'.
+    var browserString = browser.replace(/^msie$|^internet explorer$/, 'ie');
```
tomgreenfield: Could we perhaps lose the string boundary tokens? I personally find it more readable without and less need to refer back to a regex reference.
```js
var browserString = browser.replace(/msie|internet explorer/, 'ie');
```

oliverfoster: I'm alright with that. They aren't needed, more implied by circumstance.

PR #1511, opened 2017-04-07
Fixes problem with Adapt.offlineStorage not setting ready without Spoor

### LINKED ISSUE #1510: Issue with Adapt.offlineStorage introduced in v2.0.18
While testing a course with the Language Picker extension enabled the adapt content gets stuck on loading screen.

Seems to be a problem with Adapt.offlineStorage

ciaranlp:
This was never setting ready to true.

Converted tabs to spaces as well.

### DIFF
--- a/src/core/js/offlineStorage.js
+++ b/src/core/js/offlineStorage.js
@@ -1,50 +1,54 @@
 define([
-	'core/js/adapt'
+    'core/js/adapt'
 ], function(Adapt) {
 
-	//Basic API for setting and getting name+value pairs
-	//Allows registration of a single handler.
+    //Basic API for setting and getting name+value pairs
+    //Allows registration of a single handler.
 
-	var OfflineStorage = Backbone.Controller.extend({
+    var OfflineStorage = Backbone.Controller.extend({
 
-		/**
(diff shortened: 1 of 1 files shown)


ciaranlp:
Apologies I should have committed the fix separately from the formatting change, the fix is lines 22-24 with the initialize function

oliverfoster:
I don't understand.

https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/adapt-offlineStorage-scorm.js#L14 < spoor sets to _isReady: false on initialize

https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/adapt-contrib-spoor.js#L24 < spoor sets to _isReady: true if disabled

https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/adapt-contrib-spoor.js#L40 < spoor sets to _isReady: true if enabled

if initialize is never called _isReady never becomes false.

brian-learningpool:
`initialize` is called by the `new OfflineStorage()` line though.

oliverfoster:
fair enough

oliverfoster:
This isn't really a breaking change... is it?

brian-learningpool:
It isn't, I'll update the wording on the ticket.

PR #1520, opened 2017-04-11
issue-1519: default disable a11y on touch devices

### LINKED ISSUE #1519: disabled accessibility on mobile as default
causes issues on ios more times than not. courses with accessibility on mobile require specific attention for the moment.

oliverfoster:
fixes #1519

### DIFF
--- a/src/core/js/accessibility.js
+++ b/src/core/js/accessibility.js
@@ -274,7 +274,7 @@ define([
                 this.$accessibilityToggle.remove();
             }
 
-            if (!Modernizr.touch || this.isActive() || Adapt.config.get("_accessibility")._isDisabledOnTouchDevices) return;
+            if (!Modernizr.touch || this.isActive() || Adapt.config.get("_accessibility")._isDisabledOnTouchDevices !== false) return;
 
             //If a touch device and not enabled, remove accessibility button and turn on accessibility
 
--- a/src/core/js/accessibility.js
+++ b/src/core/js/accessibility.js
@@ -273,8 +273,12 @@ define([
                  //Remove button
                 this.$accessibilityToggle.remove();
             }
-
-            if (!Modernizr.touch || this.isActive() || !Adapt.config.get("_accessibility")._isEnabledOnTouchDevices) return;
+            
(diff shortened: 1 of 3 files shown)


andrewhosgood:
Why not reverse the whole logic? Rather than `_isDisabledOnTouchDevices (default true)`, perhaps `_isEnabledOnTouchDevices (default false)`.

```js
if (!Modernizr.touch || this.isActive() || Adapt.config.get("_accessibility")._isEnabledOnTouchDevices !== true) return;
```

oliverfoster:
That's fine if we can isolate which schemas need changing for the AT.

oliverfoster:
@brian-learningpool @taylortom @tomgreenfield  any ideas?

dancgray:
@oliverfoster it's in the [plugins/content/config/model.schema](https://github.com/adaptlearning/adapt_authoring/blob/master/plugins/content/config/model.schema#L57)

oliverfoster:
much obliged sir

oliverfoster:
just need an AT issue + PR.

tomgreenfield:
How to do all this without a breaking change...

oliverfoster:
what?

src/core/js/accessibility.js
```diff
@@ -273,8 +273,12 @@ define([
                  //Remove button
                 this.$accessibilityToggle.remove();
             }
-
-            if (!Modernizr.touch || this.isActive() || Adapt.config.get("_accessibility")._isDisabledOnTouchDevices) return;
+            
+            var config = Adapt.config.get("_accessibility");
+            // Backwards compatibility for _isDisabledOnTouchDevices
+            var isEnabledOnTouchDevices = config._isEnabledOnTouchDevices || (config._isDisabledOnTouchDevices === false);
```
NayanKhedkar: || (config._isDisabledOnTouchDevices === false);  is there any need to have this ??

oliverfoster: Old courses which don't support _isEnabledOnTouchDevices but which do have _isDisabledOnTouchDevices set to false will still work as expected.

PR #1522, opened 2017-04-11
issue/1521: remove navigation from #wrapper

### LINKED ISSUE #1521: move navigation bar outside of wrapper
* in preparation for ios scrolling fixes
* fixed position elements inside scrolling elements are particularly difficult on ios with an iframe

oliverfoster:
fixed #1521 

1. allow default navigation bar to be disabled ```course._navigation._isDefaultNavigationDisabled```
2. add ``navigation:initialize`` event for navigation bar plugins to know when to start
3. put ``#wrapper`` location classes on ``html`` tag to allow styling of navigation as usual
4. insert ``.navigation`` container before the ``#wrapper`` div so that the tab order is preserved

### DIFF
--- a/src/core/js/app.js
+++ b/src/core/js/app.js
@@ -118,11 +118,24 @@ require([
             Adapt.log.error('Error during app:dataReady trigger', e);
         }
 
-        Adapt.navigation = new NavigationView();// This should be triggered after 'app:dataReady' as plugins might want to manipulate the navigation
+        addNavigationBar();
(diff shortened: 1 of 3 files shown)


brian-learningpool:
@moloko, isn't the re-positioning of the navigation a bit of a breaking change, for themes at least?

I'm mentioning this in case our next version should be 2.1.0 rather than 2.0.20.

Issue #1536, opened 2017-04-22
adapt:initialize called before plugins have chance to complete aysnc ops during app:dataReady


chris-steele:
Causes problems such as restoring user answers: `QuestionModel` listens for `adapt:initialize`, but because `adapt-contrib-spoor` performs operations on `app:dataReady` asynchronously, this means data is not necessarily ready in time.

chris-steele:
Accidentally pushed this directly to master. If that's a problem I can revert, but this is a pretty simple yet critical fix.

chris-steele:
@oliverfoster Couldn't live with the guilt. I've moved changes from master to issue/#1536, if you could have a look that would be awesome 👍

PR #1539, opened 2017-04-23
Fix for plugins doing asynchronous work prior to adapt:initialize

chris-steele:
https://github.com/adaptlearning/adapt_framework/issues/1536

### DIFF
--- a/src/core/js/app.js
+++ b/src/core/js/app.js
@@ -118,9 +118,9 @@ require([
             Adapt.log.error('Error during app:dataReady trigger', e);
         }
 
-        if (!Adapt.isWaitingForPlugins()) triggerInitialize(newLanguage);
+        if (!Adapt.isWaitingForPlugins()) triggerInitialize();
         else Adapt.once('plugins:ready', function() {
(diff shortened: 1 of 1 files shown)


src/core/js/app.js
```diff
@@ -118,6 +118,15 @@ require([
             Adapt.log.error('Error during app:dataReady trigger', e);
         }
 
+        if (!Adapt.isWaitingForPlugins()) triggerInitialize();
+        else Adapt.once('plugins:ready', function() {
```
tomgreenfield: Now there's no arguments could we use the named callback instead?
```js
Adapt.once('plugins:ready', triggerInitialize);
```

src/core/js/app.js
```diff
@@ -118,6 +118,13 @@ require([
             Adapt.log.error('Error during app:dataReady trigger', e);
         }
 
+        if (!Adapt.isWaitingForPlugins()) triggerInitialize();
```
brian-learningpool: Can we get braces on this `if-else`?  The `else` on the next line looks off.

PR #1553, opened 2017-05-02
amend to course description as a demonstration of how

moloko:
handlebars-style expressions can now be used to pull in things like the learner's name


chucklorenz:
+1
(Is "Learner" more PC than "Student"?  If so, might consider updating [https://github.com/adaptlearning/adapt_framework/blob/master/src/course/en/course.json#L86](https://github.com/adaptlearning/adapt_framework/blob/master/src/course/en/course.json#L86))

PR #1565, opened 2017-05-18
Fixes #1564 - Ensures core assets, fonts and templates are copied

### LINKED ISSUE #1564: Problem with Grunt copy task for core fonts and assets
The core fonts and assets should not be filtered and should be copied regardless.  Currently they are being filtered by the function intended to only copy enabled plugins.

brian-learningpool:
The includedFilter() helper function would cause problems by running the filter on essential core files.  This affects the authoring tool where the config.json contains the list of plugins to include.

### DIFF
--- a/grunt/config/handlebars.js
+++ b/grunt/config/handlebars.js
@@ -25,7 +25,7 @@ module.exports = function(grunt) {
                     dest: '<%= outputdir %>templates.js',
                     filter: function(filepath) {
                         if (filepath.indexOf(grunt.config('sourcedir') + 'core/') > -1) {
-                            // Always include core teomplates.
+                            // Always include core templates.
                             return true;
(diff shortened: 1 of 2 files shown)


grunt/config/handlebars.js
```diff
@@ -24,6 +24,11 @@ module.exports = function(grunt) {
                     follow: true,
                     dest: '<%= outputdir %>templates.js',
                     filter: function(filepath) {
+                        if (filepath.indexOf(grunt.config('sourcedir') + 'core/') > -1) {
+                            // Always include core teomplates.
```
bell87: typo - teomplates

oliverfoster: nice, i love writing teomplates

Issue #1594, opened 2017-06-20
update inview to latest version


oliverfoster:
* fixes android setTimeout throttling
* stop inview events propagating 

https://github.com/adaptlearning/jquery.onscreen

Issue #1591, opened 2017-06-19
bug: android chrome timeout throttling


oliverfoster:
https://bugs.chromium.org/p/chromium/issues/detail?id=570845

issue: with the updated media component timeouts are being excessively throttled for all of adapt code
impact: onscreen/inview and resize callbacks are delayed or missed
result: trickle show + hide of autohide button is severely delayed on android
solution: move to requestAnimationFrame instead of setTimeout where applicable in both jquery.resize and jquery.onscreen

moloko:
@oliverfoster I've got a few queries about this...
* what 'updated media component'?
* the [chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=570845) you link to is very old, it was opened in Dec '15 and closed + fixed in Jan '16
* if this is high priority should it not be fixed for the current 2.1.x version? You have added it to the #1571 which indicates that it should be left for the v3 release

oliverfoster:
* https://github.com/mediaelement/mediaelement/compare/2.16.3...2.22.1 media element has had a lot of changes since I wrote inview back in 2015, some of which are negatively impacting performance on android (especially on a page with lots of media components)
* that bug may be "fixed" but the throttling is still present on android https://bugs.chromium.org/p/chromium/issues/detail?id=570845#c19
* I have written the fix already I just need to PR it in I will get it in for 2.1.x, it definitely needs to be in before we drop transition to v3 - it was a misunderstanding on the intention of the ie8 epic on my part. I'll remove it from there.

moloko:
@oliverfoster gotcha, thanks for the clarifications!

Issue #1452, opened 2017-03-07
bookmarking visible check error


oliverfoster:
https://github.com/adaptlearning/adapt-contrib-bookmarking/blob/master/js/adapt-contrib-bookmarking.js#L187

doesn't work as intended as ``:visible`` selector is a bit weird. https://api.jquery.com/visible-selector/

@moloko suggests changing to ``measurements.onscreen`` instead. which is calculated as detailed here > https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/libraries/inview.js#L467-L475

moloko:
just giving my alternative version a try on a project first to see how it fares in QA...

moloko:
Cannot remember what project I did this on now but I never heard back so assume it worked OK... just need to try and remember exactly how I changed it :-(

oliverfoster:
https://github.com/adaptlearning/jquery.onscreen/issues/22

oliverfoster:
fixed in above prs

PR #1595, opened 2017-06-20
issue/1594: updated inview to version 2017-07-11

oliverfoster:
#1594 update to latest version
#1591 fix android setTimeout throttling
#1452 fix bookmarking negative area issue

* fixes android setTimeout throttling
* stop inview events propagating
* use window.innerHeight and width to calculate window size
* fixes negative area issue #1452

### DIFF
--- a/src/core/js/libraries/inview.js
+++ b/src/core/js/libraries/inview.js
@@ -1,5 +1,5 @@
 'use strict';
-// jquery.onscreen 2017-06-19 https://github.com/adaptlearning/jquery.onscreen
+// jquery.onscreen 2017-06-22 https://github.com/adaptlearning/jquery.onscreen
 
 (function() {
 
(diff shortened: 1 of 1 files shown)


src/core/js/libraries/inview.js
```diff
@@ -392,6 +420,7 @@
             var body = $("body")[0].getBoundingClientRect();
             //make sure to get height and width independently if getBoundingClientRect doesn't return height and width;
             measurements.supplimentDimensions = (body.width === undefined);
+            loop.hasRaf = (window.requestAnimationFrame && window.cancelAnimationFrame);
```
NayanKhedkar: is it compatible with all browsers ex. Mozilla has window.mozRequestAnimationFrame.

oliverfoster: :/ are you using firefox 10?

oliverfoster: either way, this code is to check if raf is available and if not use ```setTimeout``` and ```clearTimeout``` instead.

moloko: Firefox ESR is the lowest version we support - as of this month that is v52.x

NayanKhedkar: 😃  i am not using firefox 10 , i was giving example of it 
I mean to say can we write this fallback or doesn't need to write this 
window.requestAnimationFrame = window.requestAnimationFrame
    || window.mozRequestAnimationFrame
    || window.webkitRequestAnimationFrame
    || window.msRequestAnimationFrame

oliverfoster: 👍 yup, definitely no longer required

moloko:
I'll kick the tyres on this a bit later today

PR #1607, opened 2017-06-22
Backbone.Collection cleanup

ryanrolds:
While tracking down long load times we noticed a lot of stale event listeners related to Backbone.Collections that were not being cleaned up. Backbone.Collection binds event listeners to every model that is added to it. Unless the models are removed or the Collection is reset those event listeners remain on the models, a performance problem (event dispatching) and a memory leak.

This change set adjusts several core functions so that they no longer return a Backbone.Collection, but instead an array of models. These functions where used by other plugins, which I've also adjusted.


src/core/js/models/adaptModel.js
```diff
@@ -236,9 +245,9 @@ define([
         },
 
         getAvailableChildren: function() {
-            return new Backbone.Collection(this.getChildren().where({
+            return this.getChildren().where({
```
oliverfoster: we're going to need two versions of ```getAvailableChildren``` and ```findDescendants```. the original with a deprecated message and a new version we can switch over to. otherwise this is gonna cause some wonderful breakages in 3rd party extensions.

this is a rather large breaking change otherwise.

src/core/js/models/adaptModel.js
```diff
@@ -345,15 +354,15 @@ define([
         },
 
         setSequentialLocking: function() {
-            var children = this.getAvailableChildren().models;
+            var children = this.getAvailableChildren();
```
moloko: would you mind trimming the trailing whitespace?

src/core/js/models/adaptModel.js
```diff
@@ -172,29 +183,27 @@ define([
...
             // first check if descendant is child and return child
             if (this._children === descendants) {
-                return this.getChildren();
+                return children;
             }
 
             var allDescendants = [];
             var flattenedDescendants;
-            var children = this.getChildren();
             var returnedDescedants;
```
moloko: any chance you could correct the typo in this variable name while you're here? it's been annoying me for ages!

oliverfoster: :O !

src/core/js/models/adaptModel.js
ryanrolds: Will do. Thanks.

src/core/js/models/adaptModel.js
ryanrolds: Of course.

src/core/js/models/adaptModel.js
ryanrolds: Will do.

src/core/js/models/adaptModel.js
oliverfoster: @moloko and I had a conversation about this. it might be worth just swallowing the repercussions and take advantage of the v3 release to just do this outright? that would be my preference even though it'll require everyone who uses these two functions to update their extensions.

ryanrolds: I can go either way. I see the advantage of tandem functions as it will require less work for plugin authors for them to pull up to v3.

oliverfoster: ``duplicate functions, depr warning, remove on v4, give time to update, w. few acute breaking issues``

^ we all decided this in the gitter room, after a brief conversation about the merits of dropping vs deprecating

ryanrolds:
@oliverfoster @moloko Updated with tandem functions and deprecation notices

moloko:
Just in the middle of testing this... one quick note while I think of it: the deprecation warnings don't give you any indication of what code is calling the deprecated function. 

It would be nice (but not essential) if we could give some indication of what code is making the call. 

Possible?

oliverfoster:
not really @moloko  especially not with ie8-10 support. https://developer.mozilla.org/en-US/docs/Web/API/Console/trace < this would probs be what we need otherwise?

ryanrolds:
I can see about including more information based on browser support, but I don't see this as a show stopper. Searching the main directory and all subdirectories for the function name cited in the error will return any instances of the deprecated function.

moloko:
fair enough, lets go with it as-is and if we find people do want more info we can sort that out later (and once IE11 is the minimum supported browser!)

PR #1612, opened 2017-06-24
amends logging module to use matching methods of the console object

moloko:
... when a matching method is available e.g. `Adapt.log.error('this is an error')` will use `console.error` in browsers that support it - but `Adapt.log.fatal('doh!')` fall back to using `console.log`


moloko:
this may as well go in as part of v3 since it will have no use whatsoever in IE8 or 9

Issue #1601, opened 2017-06-22
startController: remove IE8 hack


moloko:
in `setStartLocation`

moloko:
done in dfc60ac75626d5dbdb106169617dbaefc65b2ab3

PR #1613, opened 2017-06-26
removed IE8 hack

sarveshwar-gavhane:
Issue - https://github.com/adaptlearning/adapt_framework/issues/1601

### DIFF
--- a/src/core/js/startController.js
+++ b/src/core/js/startController.js
@@ -17,12 +17,7 @@ define([
 
             if ('replaceState' in window.history) {
                 window.history.replaceState('', '', hash);
-            } else {
-                // IE8 does not support window.history.replaceState
-                // This is the best approximation taken from Backbone.Router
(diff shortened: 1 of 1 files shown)


src/core/js/startController.js
```diff
@@ -17,12 +17,7 @@ define([
 
             if ('replaceState' in window.history) {
```
moloko: I don't think the `if()` statement is required any more

oliverfoster: +1

src/core/js/startController.js
```diff
@@ -14,16 +14,7 @@ define([
             if (!this.isEnabled()) return;
 
             var hash = this.getStartHash();
```
moloko: the local variable `hash` is now redundant, might as well delete this line and change the following line to:
```js
window.history.replaceState('', '', this.getStartHash());
```

moloko:
Please leave this unmerged until we are ready to start with the v3 changes!

PR #1617, opened 2017-06-26
Use DocumentFragment to cut down on page redraws/reflows

ryanrolds:
#1618 

On mobile devices page redraw/reflows can be a significant performance issue. When rendering pages containing 50+ questions we saw noticeable performance impact related to the rendering of the page. Creating a document fragment and rendering children in to the fragment prior to adding the fragment to the course/main document combats this problem.

### DIFF
--- a/src/core/js/router.js
+++ b/src/core/js/router.js
@@ -46,9 +46,9 @@ define([
 
         handleRoute: function() {
             var args = this.pruneArguments(arguments);
-            
+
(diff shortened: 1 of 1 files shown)


ryanrolds:
I need to do some work on this. It's a step in the right direction, but it doesn't solve the issue for blocks and components (add children is only preventing reflows for articles, not their children).

Issue #1614, opened 2017-06-26
Remove legacy CSS code for IE8 hack


sarveshwar-gavhane:
https://github.com/adaptlearning/adapt_framework/blob/master/src/core/less/legacy.less#L1
https://github.com/adaptlearning/adapt_framework/blob/master/src/core/less/legacy.less#L18

akshay-bhavsar:
@sarveshwar-gavhane 
We can removed the legacy file for this?

sarveshwar-gavhane:
PR : https://github.com/adaptlearning/adapt_framework/pull/1626

moloko:
#1626 seems to be broken, there's no source repository

PR #1626, opened 2017-06-30
legacy for IE8 removed

akshay-bhavsar:
#1614


tomgreenfield:
Could we also take out the below while we're here?

* https://github.com/adaptlearning/adapt_framework/blob/master/src/core/less/base.less#L164-L166
* https://github.com/adaptlearning/adapt_framework/blob/master/src/core/less/base.less#L216-L219

moloko:
the repository/branch these changes came from seems to have disappeared?

oliverfoster:
closed as file needs to be deleted and tom's request needs to be met but source has been deleted. inavlid pr.

PR #1627, opened 2017-06-30
issue/1374: Add header title to drawer-toolbar when in drawer item

barryo-learningpool:


sarveshwar-gavhane:
when i click on resource drawer item few times and close the drawer,  i realised that its creating multiple references of title
drawer

src/core/js/views/drawerView.js
```diff
@@ -86,6 +86,19 @@ define([
...
+            var className = view.context.className;
+
+            // Use the class name of the drawer item to retrieve its model from the collection of drawer items.
+            var model = this.collection.find(function(model) {
+                return model.get('className') === className + '-drawer';
+            });
+
+            if (this._hasBackButton) {
+                // Display the title of the drawer item after (next to) the back button.
+                this.$('.drawer-back-button').after('<div class="drawer-custom-view-title">' + model.get('title') + '</div>');
```
sarveshwar-gavhane: can you please use other method /html() its creating instances every time user click on drawer

src/core/js/views/drawerView.js
```diff
@@ -86,6 +86,19 @@ define([
...
+
+            // Get the class name of the drawer item.
+            var className = view.context.className;
+
+            // Use the class name of the drawer item to retrieve its model from the collection of drawer items.
+            var model = this.collection.find(function(model) {
+                return model.get('className') === className + '-drawer';
+            });
+
+            if (this._hasBackButton) {
```
sarveshwar-gavhane: can you please ``` if (this._hasBackButton &&  model.get('title')) ``` or something like this , its always good idea to check current drawer item has title, for example if PLP doesn't have title attribute still its displaying title of resource in drawer of PLP currently

barryo-learningpool: Hi @sarveshwar-gavhane, thanks for the pointers.  I've pushed some changes which seem to fix this behaviour

sarveshwar-gavhane: great @barryo-learningpool

tomgreenfield:
@barryo-learningpool, thanks for your work so far!

Would you mind updating this to be a bit more dynamic, with less manual injection of elements?

For example, the template could include the `<div>` by default:
```hbs
<div class="drawer-custom-view-title">
    <div class="drawer-custom-view-title-inner"></div>
</div>
```
In the JS we could have a new function dedicated to setting the title:
```js
setTitle: function(title) {
    $('.drawer-custom-view-title-inner').html(title);
}
```
`setTitle()` would then be called from two different places:
* Firstly from [`openCustomView()`](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/drawerView.js#L83):
```js
var model = this.collection.findWhere({
    className: view.attr('class') + '-drawer'
});

this.setTitle(model ? model.get('title') : '');
```
* Secondly from [`emptyDrawer()`](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/drawerView.js#L187)
```js
this.setTitle('');
```
@guywillis has also kindly provided the Less needed for the title to be properly centred vertically:
```less
.drawer-custom-view-title {
    float: left;
    display: table;
    width: 192px;
    height: 64px;
    font-size: 18px;
    font-weight: 300;
}
.drawer-custom-view-title-inner {
    display: table-cell;
    padding: 0 10px;
    vertical-align: middle;
}
```
Feel free to have a play around and let us know what you think, cheers.

barryo-learningpool:
Hi @tomgreenfield, thanks for the pointers.  I have updated the code as your suggested.  This will require themes to also add the drawer-custom-view-title and drawer-custom-view-title-inner elements to drawer.hbs template.  Should I make the change on the vanilla theme?

tomgreenfield:
Cheers @barryo-learningpool.

Vanilla v2.1+ no longer includes templates. The code will fail gracefully if the element doesn't exist so this looks good to me.

src/core/js/views/drawerView.js
```diff
@@ -86,7 +86,14 @@ define([
             this._isCustomViewVisible = true;
             Adapt.trigger('drawer:empty');
             this.showDrawer();
+
             this.$('.drawer-holder').html(view);
+
+            var model = this.collection.findWhere({
```
sarveshwar-gavhane: @barry-learningpool it works fine for resource drawer, but  thinking forward , when i think about PLP ``` this.collection ``` always have resource object in it https://github.com/adaptlearning/adapt-contrib-resources/blob/master/js/adapt-contrib-resources.js#L29-L35 . and it never gets empty , when you click on any other drawer item,  undefined will be return

tomgreenfield: Sorry, I'm not fully understanding your issue; would you mind elaborating a bit more?

I tested this PR with both [Resources](https://github.com/adaptlearning/adapt-contrib-resources) and [Glossary](https://github.com/adaptlearning/adapt-contrib-glossary) to check that the title switches out correctly.

sarveshwar-gavhane: good morning ! no issues with resources, can please check with pageLevelProgress once ??

barryo-learningpool: Hi @sarveshwar-gavhane, I tried it with pageLevelProgress but did not see any problems. Could you perhaps give the steps to reproduce any problems you've seen?

sarveshwar-gavhane: can you able to see the title header for PLP ? then my mistake apologies

tomgreenfield: Page Level Progress doesn't provide any title at present. I suggest opening a separate issue if that's what you're referring to, thanks.

sarveshwar-gavhane: sure , i have added the title in PLP object in course.json 
```
 "_pageLevelProgress": {
        "_isEnabled": true,
        "title": "page Level Progress",
        "_showPageCompletion": false,
        "_altText": "of components completed on this page."
    },  ``` 
but unable to see the title for PLP drawer like resources

Comment (moloko) [inline src/core/js/views/drawerView.js]: @sarveshwar-gavhane pageLevelProgress has never had a title

Comment (sarveshwar-gavhane) [inline src/core/js/views/drawerView.js]: yes PLP has never had a title , but if we add title to the PLP object in course.json , what expected is it will show PLP title, 
expectation from this PR is  , if any(PLP,language picker,resource,glossary) drawer item has title then it shows title when we click on that particular drawer item,

Comment (tomgreenfield) [inline src/core/js/views/drawerView.js]: @barryo-learningpool, I think if we want to do this properly, we'll need to update all the plugins that use Drawer to pass their actual view to `Adapt.drawer.triggerCustomView()` rather than their view's element.

For example, in [Resources](https://github.com/adaptlearning/adapt-contrib-resources/blob/19db7d410e96b5462108d3054f010485bb7c47c8/js/adapt-contrib-resources.js#L14-L17), removing the `.$el` from:
```js
Adapt.drawer.triggerCustomView(new ResourcesView({
    model: resourcesModel,
    collection: resourcesCollection
}).$el);
```
Most plugins will then already have their model available (PLP will need one adding in the same vein as [Resources](https://github.com/adaptlearning/adapt-contrib-resources/blob/19db7d410e96b5462108d3054f010485bb7c47c8/js/adapt-contrib-resources.js#L10)).

To keep backwards compatibility, in this PR we'll need to change [this line](https://github.com/barryo-learningpool/adapt_framework/blob/07a3722783e015f4163eb73b173e5a7dbb446676/src/core/js/views/drawerView.js#L90) to
```js
this.$('.drawer-holder').html(view.$el || view);
```
and [`model`](https://github.com/barryo-learningpool/adapt_framework/blob/07a3722783e015f4163eb73b173e5a7dbb446676/src/core/js/views/drawerView.js#L92-L94) will become
```js
var model = view.model;
```
Happy with this approach?

oliverfoster: I'd like to see this replicated in the notify also.

moloko:
where are we with this one - are further changes required or is it ready for review again?

moloko:

Issue #1596, opened 2017-06-20
contrib-narrative: Swiping on touch devices


simondate:
It seems like a more natural interaction for learners to swipe through images rather than pressing a button on touch based devices.

brian-learningpool:
I wonder should we add something like [HammerJS](http://hammerjs.github.io/) to core to facilitate this?

simondate:
I just tried HammerJS on my Android phone and performance was pretty poor (1s delay). Is there a better library to use?

Is there any other components that could use the mobile events treatment?

brian-learningpool:
AlloyFinger looks promising.  I believe there's a JQuery UI Touch Punch library too.  I can't really say at the minute what's better though.

Narrative is the core component most obviously missing gestures.

brian-learningpool:
@simondate, I brought this up on this morning's framework call and the consensus was to go with JQuery Mobile's touch events.

simondate:
Cheers @danielstorey . I'll be happy to work on actually implementing next week.

brian-learningpool:
This is already implemented in https://github.com/adaptlearning/adapt-contrib-narrative/pull/174.

oliverfoster:
@danielstorey would you be able to add the appropriate event listeners to the narrative to close this ticket?

PR #1633, opened 2017-07-04
Add support for swipe events

danielstorey:
This makes #1596 possible whilst also making swipe events available for other plugins to use, rather than adding an exclusive dependency just for narrative.

Note this update does not directly add swipe functionality to the Narrative component.

This file was compiled using the custom download build and contains just the code needed for touch/swipe events. Other jQuery mobile modules are therefore not included.

See http://api.jquerymobile.com/category/events/ for usage


oliverfoster:
can we have an unminified version?

danielstorey:
There you go!

oliverfoster:
ta vm.

oliverfoster:
where did you get this from? @danielstorey can i have a source reference before we merge it?

danielstorey:
http://jquerymobile.com/download-builder/

Select **Touch** under **Events** then **Build My Download**

oliverfoster:
cool, it's a shame there's not permanent link to it.
it it worth putting a comment at the top with the url it came from and what was selected?

sarveshwar-gavhane:
yes it is there are so many modules to select.

danielstorey:
Yeah it is a shame. The only permanent links would be to the full library which is 196Kb. This version is only 8Kb! The good thing is though jQuery allow custom downloads from each branch so it'll always be accessible via the builder in the link above. I've added instructions and also renamed the file to the original name of the downloaded file, to avoid any confusion

Issue #1365, opened 2016-12-22
expose components active item index 


lc-thomasberger:
Components with items should keep track of the active item in their model. This would enable us to listen to the change event and call actions in extensions. Furthermore, setting the components state, should be possible by setting a value for an active item index.

Example usecase: A custom audio extension can play audiotracks depending on the currently active item of a narrative component.

Core components with items:
- accordion
- narrative
- hotgraphic

tomgreenfield:
Directly related to https://github.com/adaptlearning/adapt_framework/issues/1178.

oliverfoster:
closed in https://github.com/adaptlearning/adapt_framework/pull/1638

Issue #1637, opened 2017-07-06
Add items collection


tomgreenfield:
Proposal: use a true Backbone collection for items in a component model.

Benefits include built-in change events and helper functions, and it also works to align with Adapt's existing *parent > collection of children* architecture.

Building on the excellent foundation from @lc-thomasberger, the functions have been whittled down to the below:

### Component level `itemsComponentModel.js`

* `init()`: custom initialise function to set up items and event listeners
* `setUpItems()`: create a collection of `ItemModel`s
* `getItem(index)`: get the item at specified index
* `getVisitedItems()`: return all items where `_isVisited: true`
* `getActiveItems()`: return all items where `_isActive: true`
* `getActiveItem()`: return first item where `_isActive: true`
* `areAllItemsCompleted()`: does what it says on the tin
* `checkCompletionStatus()`: set component to complete if all items are completed (easily overriden)
* `reset()`: resets all items and then component
* `resetActiveItems`: set all items to `_isActive: false`

### Item level `itemModel.js`

* `reset()`: shortcut to set `_isActive: false` and `_isVisited: false`
* `toggleActive(isActive)`: shortcut to `this.set('_isActive', ...);`
* `toggleVisited(isVisited)`: shortcut to `this.set('_isVisited', ...);`

Issue #1178, opened 2016-07-28
Create "items" presentation component model


oliverfoster:
All of these components use the same model behaviour
- Narrative
  * update transformations to css3 inline with dropping ie8-ie10
- Accordion (https://github.com/adaptlearning/adapt-contrib-accordion/pull/85)
- Hotgraphic
  * use [notify subviews](https://github.com/adaptlearning/adapt_framework/pull/1893) and itemscollection

Create a standard presentation component model for them

oliverfoster:
any objection to me putting this in the core?

moloko:
Could you also deal with #1365 at the same time?

PR #1638, opened 2017-07-06
Add items collection

tomgreenfield:
Original issue: https://github.com/adaptlearning/adapt_framework/issues/1637.
fixes: 
https://github.com/adaptlearning/adapt_framework/issues/1365
https://github.com/adaptlearning/adapt_framework/issues/1178

Please test in conjuction with https://github.com/adaptlearning/adapt-contrib-accordion/tree/issue/1637.


oliverfoster:
@lc-thomasberger  what do you think?

danielstorey:
I just started to test this using a new component for which I originally used itemsModel.

I understand the benefit of using a collection but now it feels a bit more laboured now as I can no longer use a single handlebars template to iterate over `_items` or refer to specific indices, and am having to write extra functions and create new templates in order to perform a simple initial render.

The issue is that `.toJSON` returns a shallow copy of the model's `attributes` property. Can we therefore redifine this method on the itemsComponentModel ie
```
toJSON: function() {
    var json = _.clone(this.attributes);
    json._items = this.get('_items').toJSON();
    return json;
},
```
With maybe some defensive programming to ensure we can call `toJSON` on the items collection

oliverfoster:
Sounds fine to me. Need anything else?

danielstorey:
Everything else works great!

src/core/js/models/itemsComponentModel.js
```diff
@@ -0,0 +1,76 @@
...
+        getActiveItem: function() {
+            return this.get('_items').findWhere({ _isActive: true });
+        },
+
+        areAllItemsCompleted: function() {
+            return this.getVisitedItems().length === this.get('_items').length;
+        },
+
+        checkCompletionStatus: function() {
+            if (this.areAllItemsCompleted()) {
```
oliverfoster: return early?

tomgreenfield: Aha, I think defiite overkill / needless expansion for single-line if statements. Only an issue if dealing with multiple nested lines and/or actions after the conditional.

src/core/js/models/itemsComponentModel.js
```diff
@@ -0,0 +1,76 @@
...
+            this.listenTo(this.get('_items'), {
+                'change:_isVisited': this.checkCompletionStatus
+            });
+        },
+
+        setUpItems: function() {
+            var items = this.get('_items');
+            var collection = new Backbone.Collection(null, { model: ItemModel });
+
+            items.forEach(function(item, index) {
```
brian-learningpool: You could actually call `.map()` here and call the Collection `add()` with an array of `items` after.

src/core/js/models/itemsComponentModel.js
```diff
@@ -0,0 +1,76 @@
...
+        },
+
+        checkCompletionStatus: function() {
+            if (this.areAllItemsCompleted()) {
+                this.setCompletionStatus();
+            }
+        },
+
+        reset: function(type, force) {
+            this.get('_items').each(function(item) { item.reset(); });
```
brian-learningpool: Would be more readable on a single line if we were using arrow functions.

oliverfoster: :O * gasp *

brian-learningpool: Can't do that with IE 11 still in the mix though. :-/

oliverfoster: I need to have a chat with you about that kind of stuff.

src/core/js/models/itemsComponentModel.js
oliverfoster: can you write out the code? my brain has gone to sleep... :/

brian-learningpool: Something like this:

```
items.map(function(item, index) {
  item._index = index;
  return item;
});

collection.add(items);
```

danielstorey: This doesn't make use of the array returned by calling .map so .forEach could still be used. If immutability is desirable though
var items = items.slice().map(
would prevent mutation

Also since there's now only one .add call why not just use

var collection = new Backbone.Collection(items, { model: itemModel })

After the array has been mapped.

Sorry for the lack of MD I'm on a phone currently.

brian-learningpool: Good point.

oliverfoster:
could you all re-+1 this so it can go in?

PR #1640, opened 2017-07-06
Ensure jQuery mobile gets loaded  - completes #1633

danielstorey:
#1633 added jqm to the libraries folder but it doesn't get loaded without this change.

### DIFF
--- a/src/core/js/scriptLoader.js
+++ b/src/core/js/scriptLoader.js
@@ -30,7 +30,8 @@
                 a11y: 'libraries/jquery.a11y',
                 scrollTo: 'libraries/scrollTo',
                 bowser: 'libraries/bowser',
-                'enum': 'libraries/enum'
+                enum: 'libraries/enum',
+                jqueryMobile: 'libraries/jquery.mobile.custom'
             }
         });
         loadJQuery();
@@ -74,7 +75,7 @@
         if (isIE8) {
             fixIE8ConsoleLog();
         }
-        
+
     }
 
     function fixIE8ConsoleLog() {
@@ -86,34 +87,35 @@
     //6. Load foundation libraries and templates
     function loadFoundationLibraries() {
         require([
-            "underscore",
-            "backbone",
-            "backbone.controller",
-            "handlebars",
-            "velocity",
-            "imageReady",
-            "inview",
-            "a11y",
-            "scrollTo",
-            "bowser",
-            "enum",
-            "templates"
+            'underscore',
+            'backbone',
+            'backbone.controller',
+            'handlebars',
+            'velocity',
+            'imageReady',
+            'inview',
+            'jqueryMobile',
+            'a11y',
+            'scrollTo',
+            'bowser',
+            'enum',
+            'templates'
         ], loadAdapt);
     }
 
     //7. Load adapt
     function loadAdapt() {
         switch (IE) {
-        case 8: case 9:
+            case 8: case 9:
             //ie8 and ie9 don't do crossdomain with jquery normally
             break;
-        default:
-            //cross domain support for all other browers
-            $.ajaxPrefilter(function( options ) {
-                options.crossDomain = true;
-            });
+            default:
+                //cross domain support for all other browers
+                $.ajaxPrefilter(function( options ) {
+                    options.crossDomain = true;
+                });
         }
-        Modernizr.load("adapt/js/adapt.min.js");
+        Modernizr.load('adapt/js/adapt.min.js');
     }
 
     //1. Load foundation libraries, json2, consoles, requirejs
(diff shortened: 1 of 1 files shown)


oliverfoster:
has the jquery.mobile.custom changed?

danielstorey:
No, I'm not sure why it's showing up in diff

brian-learningpool:
It's because the previous PR was squashed and merged.  @danielstorey, can you rebase your branch or cherry-pick 3a86941?

src/core/js/scriptLoader.js
```diff
@@ -30,7 +30,8 @@
                 a11y: 'libraries/jquery.a11y',
                 scrollTo: 'libraries/scrollTo',
                 bowser: 'libraries/bowser',
-                'enum': 'libraries/enum'
+                enum: 'libraries/enum',
+                jqueryMobile: 'libraries/jquery.mobile.custom'
```
NayanKhedkar: could you please load this library conditionally i.e. load library only for touch devices.

brian-learningpool: I can see where you're coming from but I think loading conditionally is possibly overkill in this instance.

NayanKhedkar: can we try something like this 
``Modernizr.load([
            {
                test: Modernizr.touch,
                yes: 'libraries/jquery.mobile.custom.js'
            }
        ]);``
if it is overkill instance, no problem 👍

danielstorey: It's a fair point. It would eliminate the need for code like this:
```
onSwipe: function(e) {
    // Don't allow swiping with the mouse
    if (!Adapt.device.touch) {
         return;
     }
     doSomething();
}
```

oliverfoster: why not allow swipe with the mouse?

src/core/js/scriptLoader.js
```diff
@@ -30,7 +30,8 @@
                 a11y: 'libraries/jquery.a11y',
                 scrollTo: 'libraries/scrollTo',
                 bowser: 'libraries/bowser',
-                'enum': 'libraries/enum'
+                enum: 'libraries/enum',
```
oliverfoster: please leave speech marks on 'enum', it's a reserved keyword in ie8.

src/core/js/scriptLoader.js
danielstorey: Coz I don't want to.

And to retain distinction between different devices and their methods of interaction. It's probably more just OCD than something necessary.

src/core/js/scriptLoader.js
danielstorey: Of course it is. Good ol' ie

src/core/js/scriptLoader.js
oliverfoster: I meant in general, it's a framework after all and we're not the only people who might want to use the swipe action.

oliverfoster: We're half mobile only, half desktop included at the moment. What are the arguments for either side?

moloko: I'm for loading on touch-capable devices only

danielstorey: I guess the reason I started using swipe interactions for mobile is it offers the user a way of interacting that's intuitive and something they're used to doing on other sites/apps. Allowing swiping with the mouse does the opposite ie it's counter-intuitive and fiddly so there doesn't seem much point. Having said that, most users would therefore not even be aware that they were able to do it, but I guess part of me would like it not to be there if it's not going to be used. I'm happy to add the condition if the consensus is that's something we want to do. Equally happy to continue as is :)

oliverfoster: @brian-learningpool ? Alright for the conditional loading?

dan makes a good ux point about the affordability of swiping on a non-touch device - no-one would be expecting it.

oliverfoster: how come you decided against adding touchpunch and the drag+drop bits @danielstorey

danielstorey: when did I mention that? I can't remember now lol. I'm all for adding touchpunch as well. In fact I vaguely remember experiencing issues when I included it manually on two different plugins so having it loaded once on the document would solve this. Probably a good idea - especially if Adapt is going more and more in this direction.

danielstorey: Having said that, jQuery UI is not included by default is it? So no point including touch punch unless we're including that as well.

oliverfoster: cool, a conversation for another ticket no doubt

oliverfoster: i've added an issue to touch-only loading > https://github.com/adaptlearning/adapt_framework/issues/1648 as this has merged already. i think we've hit a majority consensus on doing that?

Issue #1634, opened 2017-07-04
grunt: post compile minification


oliverfoster:
Minify all code in libraries that isn't already minified (these will be identified by being called *.min.js), offline_API_wrapper.js, templates.js and scriptLoader.js

tomgreenfield:
As we're doing this, reckon we could minify the JSON too?

barry-learningpool:
Absolutely @tomgreenfield , should I raise a separate issue or is it cool to do it as part of this one?

tomgreenfield:
Happy for it to be included here 👍

moloko:
done in https://github.com/adaptlearning/adapt_framework/commit/e018fc5f91ebbefac902c73fb8dbbe25faf00c80

PR #1642, opened 2017-07-06
#1634 - grunt build task: minification for libraries folder and json files

barry-learningpool:
https://github.com/adaptlearning/adapt_framework/pull/1634 Added UglifyJS library to package.json. Added 'libraries' configuration for 'uglify' task and calling it as part of the standard 'build' task. Files which are already minified are ignored. Files, filenames and extensions are all preserved (as these this would break the requires in scriptLoader.js)


oliverfoster:
which libraries have we got that are minified but don't have the .min.js ending? cos that'll reduce the time quite a bit.

barry-learningpool:
@oliverfoster I'm in the process of renaming the files, I've had to update the modernizr.js reference in index.html, however this references also exists in the bundled Spoor HTML files. I can go ahead and create a Spoor PR for these changes, I'm just wondering about versioning. Does this mean there needs to be a framework version bump and the framework version then updated in Spoor?

oliverfoster:
Yup, spoor will need it's minimum framework version setting also.

moloko:
@barry-learningpool sorry but could you leave modernizr alone for now please? we can sort that out when we switch to v3. just a bit concerned about the risk of changing that i.e. someone could update framework but not spoor and then they'd have a broken course.

moloko:
@barry-learningpool thanks for the amends, there's a conflict on scriptLoader.js though I'm afraid...

moloko:
also you have renamed modernizr to modernizr.min.js - but we need that to be left as modernizr.js to avoid having to update the references in the various .html files

barry-learningpool:
@moloko issues have been resolved

oliverfoster:
core/js/libraries/modernizr.min.js needs renaming

grunt/config/json-minify.js
```diff
@@ -0,0 +1,7 @@
+module.exports = function (grunt, options) {
+    return {
+        build: {
```
oliverfoster: can you rename from ``build`` to ``minify`` so the ``minify`` task is responsible for both json and javascript?

grunt/config/uglify.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function (grunt, options) {
+    return {
+        libraries: {
+            files: [{
+                expand: true,
+                cwd: '<%= outputdir %>libraries/',
+                src: ['*.js', '!*.min.js'],
+                dest: '<%= outputdir %>libraries/'
+            }]
```
oliverfoster: Could you also add these, for the ``templates.js`` and the ``scriptLoader.js``
```javascript
            ,{
                expand: true,
                cwd: '<%= outputdir %>/adapt/js',
                src: ['*.js', '!*.min.js'],
                dest: '<%= outputdir %>/adapt/js'
            },{
                expand: true,
                cwd: '<%= outputdir %>',
                src: ['*.js', '!*.min.js'],
                dest: '<%= outputdir %>'
            }
```

oliverfoster:
ta @barry-learningpool

grunt/config/uglify.js
```diff
@@ -0,0 +1,26 @@
+module.exports = function (grunt, options) {
+    return {
+        libraries: {
```
moloko: you need to add 
```js
options: {
    ie8: true
},
```
here or the minified version of the code won't load in IE8. We can take this out again once we move to v3 and drop IE8 support.

PR #1664, opened 2017-07-14
Fixes #1567 Grunt watch doesn't detect new files

### LINKED ISSUE #1567: Grunt check for new files added
If you add a new file grunt does not process it. Would it be possible to enable it to?

danielstorey:
Replace `<%= sourcedir %>` with `src/`

See https://stackoverflow.com/questions/31679375/grunt-contrib-watch-doesnt-see-new-files for reference

### DIFF
--- a/grunt/config/watch.js
+++ b/grunt/config/watch.js
@@ -1,92 +1,92 @@
 // TODO excludes
 module.exports = {
     less: {
-        files: ['<%= sourcedir %>**/*.less'],
+        files: ['src/**/*.less'],
         tasks: ['less:dev']
     },
     handlebars: {
-        files: ['<%= sourcedir %>**/*.hbs'],
+        files: ['src/**/*.hbs'],
         tasks: ['handlebars', 'javascript:dev']
     },
     courseJson: {
-        files: ['<%= sourcedir %>course/**/*.json'],
+        files: ['src/course/**/*.json'],
         tasks : ['jsonlint', 'check-json', 'copy:courseJson', 'schema-defaults', 'create-json-config']
     },
     courseAssets: {
-        files: ['<%= sourcedir %>course/<%=languages%>/*', '!<%= sourcedir %>course/<%=languages%>/*.json'],
+        files: ['src/course/<%=languages%>/*', '!src/course/<%=languages%>/*.json'],
         tasks : ['copy:courseAssets']
     },
     js: {
         files: [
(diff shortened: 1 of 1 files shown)


oliverfoster:
That link just says that relative paths starting with ./ don't work.
<% sourcedir %> is just a string replacement for the absolute path of the working folder concatenated with /src.
https://github.com/adaptlearning/adapt_framework/blob/master/grunt/helpers.js#L77

I'm not entirely sure how changing it is going to help?
On windows all the slashes in grunt are the wrong way round, the come out like "'D:\\working\\src\\" when really we should replace them all with / slashes.
Otherwise there doesn't seem to be much explanation as to why this fix would work.

Which OS are you using?

oliverfoster:
grunt-contrib-watch is notoriously bad for new files. There are libraries created just to solve that problem.

danielstorey:
I'm using windows and it works after the change. I also struggle with fixes that "just work" but if as you say grunt-contrib-watch is notoriously bad for new files then I guess we can expect not to have a nice easy explanation.

I'm guessing that grunt-contrib-watch struggles with full paths in the same way that it struggles with "./". Further diagnosis/fixes would involve getting down and dirty with grunt-contrib-watch I imagine. I did also just test it by replacing "\\" with "/" but that still didn't work.

Try it though with this change - add an image to your course folder and it will update :)

oliverfoster:
Can you change it in the defaults instead of in the task? So that sourcedir just equals the relative src/ rather than proccess.cwd() + src/ . The change you've done basically negates otherwise quite sound code - which you're not removing.

danielstorey:
This will have a knock on effect on other tasks. The less task fails if the default is changed and others may well be affected too. Every task will have to be tested in both fw and AT and potentially a fair bit of refactoring to cope with the change. Probably not worth it given that watch is the only task that doesn't work fully with the way it is currently?

oliverfoster:
after testing this, i'm finding that with/without your changes:
* adding a new folder in a watched folder is fine without adding any changes (like: src/components/adapt-contrib-accordion/less/new folder)
* adding a new folder in an unwatched folder doesn't work regardless (like: src/components/adapt-contrib-new)

danielstorey:
Ok fair enough. Well there's definitely an improvement when adding files in watched folders but accept that there's still an issue adding a new folder.

PR #1666, opened 2017-07-14
Update to previous commit for #1487.

danielghost:
Amend to prevent optional models from being marked as completed when a different model with the same parent has been viewed. Course completion will still be based only on completing mandatory models.

Also fixed issue with condition as per https://github.com/adaptlearning/adapt_framework/commit/6d3031943ec1947f83f9ea73cddf1b379b52b716#diff-de3e37cd39003ccf4ed26bc787b21871R130.


moloko:
hi @danielghost I'm afraid you have a conflict that needs resolving...

moloko:

PR #1669, opened 2017-07-14
Updated amend for #1486.

danielghost:
Locking doesn't account for optional models.

Also, fixed an issue created in #1488, where only listening to a models immediate children prevented unlocking models outside of this hierachy. Now should also listen to changes for those models listed in `_lockedBy`.

### DIFF
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -467,7 +467,10 @@ define([
                 try {
                     var model = Adapt.findById(id);
 
-                    if (!model.get("_isAvailable")) continue;
+                    this.stopListening(model, "change:_isComplete");
+                    this.listenTo(model, "change:_isComplete", this.onIsComplete);
+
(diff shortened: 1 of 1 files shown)


src/core/js/models/adaptModel.js
```diff
@@ -467,7 +467,10 @@ define([
                 try {
                     var model = Adapt.findById(id);
 
-                    if (!model.get("_isAvailable")) continue;
+                    this.stopListening(model, "change:_isComplete");
```
moloko: I feel this needs some kind of explanatory comment... mainly based on the fact that I can't figure out why you're unregistering a listener then immediately re-registering it again!

oliverfoster: it also needs to be changed to include the callback function.
this call to stopListening will remove all "change:_isComplete" event listeners attached to this model, which can override subsequently added behaviour.
```
this.stopListening(model, "change:_isComplete", this.onIsComplete);
```

moloko:

PR #1671, opened 2017-07-16
Fix #1670: RTL vertical align of components

### LINKED ISSUE #1670: RTL vertical align of components is wrong
When placing two components in a block and the course uses RTL as _defaultDirection, then the vertical alignment is done wrong. The component on the right on large screens should be aligned on top on small screens. Right now it is the opposite:

The following screenshot already shows the right behaviour:

rtl-issue-vertical

dasistwas:
Fixes vertical alignment issue for RTL course configuration described in https://github.com/adaptlearning/adapt_framework/issues/1670

### DIFF
--- a/src/core/js/views/adaptView.js
+++ b/src/core/js/views/adaptView.js
@@ -85,7 +85,11 @@ define([
                     if (ChildView) {
                         var $parentContainer = this.$(this.constructor.childContainer);
                         model.set("_nthChild", nthChild);
-                        $parentContainer.append(new ChildView({model:model}).$el);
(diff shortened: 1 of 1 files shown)


brian-learningpool:
@lc-thomasberger, is there an impact on multi-language courses here?

oliverfoster:
is there a course.json specific rtl value? or it is only done in the config.json?

moloko:
@oliverfoster `_defaultDirection` initially gets set via config.json but then can be changed by plugins like [languagePicker](https://github.com/adaptlearning/adapt-contrib-languagePicker)

brian-learningpool:
Perfect, we're covered with this PR then. 👍

oliverfoster:
ta @moloko

PR #1679, opened 2017-07-19
Fixes #1675

### LINKED ISSUE #1675: In-course links in Notify content disable scrolling
If you include a hyperlink to another part of the course inside a Notify popup, the link works but scrolling is subsequently completely disabled.

This is presumably down to notifyView disabling scrolling when a Notify popup gets opened - but is not being given the chance to re-enable it again because clicking the link will cause Adapt to route to the new location without calling `closeNotify`.

Looking at the code I suspect this will also have an impact on accessibility, though I haven't tested that.

It seems reasonable to me that someone would want to add a link to another part of the course to a Notify popup - in this particular case it was to direct the user to a part of the course they needed to revisit because they'd answered a question incorrectly - so it would be great to get this working.

danielstorey:


moloko:
Works for me... tested in Firefox

src/core/js/views/notifyView.js
```diff
@@ -91,7 +92,13 @@ define([
             event.preventDefault();
             //tab index preservation, notify must close before subsequent callback is triggered
             this.closeNotify();
-            Adapt.trigger("notify:cancelled");
+            Adapt.trigger('notify:cancelled');
+        },
+
+        onLinkClicked: function(e) {
```
moloko: I think this could use an explanatory comment, maybe including a link to #1675

danielstorey:
I think a lot of code could do with explanatory comments ;)

danielstorey:
Removed link event handler. Instead call closeNotify on remove which will ensure the necessary happens when the route changes. Adding a listener to page:scrollTo ensures that stuff happens when navigating to an element on the same page.

src/core/js/views/notifyView.js
```diff
@@ -19,7 +19,8 @@ define([
         },
 
         setupEventListeners: function() {
-            this.listenTo(Adapt, 'remove', this.remove);
+            this.listenTo(Adapt, 'remove', this.closeNotify);
```
moloko: Now that there's quite a few of these, would you mind changing this to the object style i.e.
```js
this.listenTo(Adapt, {
    'remove', this.closeNotify,
    'page:scrollTo': this.closeNotify,
    'device:resize': this.resetNotifySize,
    'accessibility:toggle': this.onAccessibilityToggle
});
```

danielstorey: Good call, I've done as you suggested Matt, as well as condensing remove and page:scrollTo to a single key

Issue #1582, opened 2017-06-09
router.js _isCircularNavigationInProgress reset issue


oliverfoster:
_isCircularNavigationInProgress doesn't reset when canNavigate is set to true.
this is was found by resetting an assessment twice in quick succession (see #1584 and #1585)

note: circular navigation is when a course changes the url when it's not allowed to, the url must be corrected.

oliverfoster:
i missed a recheck of canNavigate which broke pageIncompletePrompt https://github.com/cgkineo/adapt-pageIncompletePrompt/issues/13

brian-learningpool:
Code has been merged to `master`, closing issue.

PR #1688, opened 2017-07-26
issue/1582: _canNavigate after router:navigate

oliverfoster:
#1582 

originally from https://github.com/cgkineo/adapt-pageIncompletePrompt/issues/13


moloko:
I think this needs to go into the legacy branch as well

PR #1697, opened 2017-08-04
Added tab index to allow the notify popup body to become focus-able.

ryan-lynch:


ryan-lynch:
Apologies I misunderstood how this accessibility works with child nodes.

Issue #1709, opened 2017-08-10
Remaining attempts text is not accessible.


ryan-lynch:
There is currently no accessibility for "remaining attempts" on components. 

I have two potential solutions: 

1. add an aria-label attribute via JS in core buttonsView 
2. add a tabindex attribute to the template in core and the vanilla theme. 

I would probably prefer to add a concrete aria-label via JS.

moloko:
as discussed I think the way forward here is to amend the final few lines of `updateAttemptsCount` in buttonsView.js from this:
```js
if (shouldDisplayAttempts) {
    this.$('.buttons-display-inner').html(attemptsString);
}
```
to this:
```js
if (shouldDisplayAttempts) {
    this.$('.buttons-display-inner').html(attemptsString).a11y_text();
}
```

danielstorey:
I can foresee this kind of behaviour being needed more and more. Would it be worth creating another helper or even allowing an optional argument [here](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/libraries/jquery.a11y.js#L956-L965) which when present inserts accessible text rather than converting the existing content.

oliverfoster:
fine plan! i'll +1 if you pr.

just as a note.... 
```javascript
this.$('.buttons-display-inner').html($.a11y_text(attemptsString));
```
also does the same thing

danielstorey:
Of course, but less brackets is always good ;) #1725

oliverfoster:
couldn't agree more

moloko:
done in https://github.com/adaptlearning/adapt_framework/commit/9d50f2fc9b71be7b6294decfa05d9aba3e9b6759

PR #1711, opened 2017-08-10
issue/1709 - Added support for accessibility to display remaining attempts

ryan-lynch:
#1709

### DIFF
--- a/src/core/js/views/buttonsView.js
+++ b/src/core/js/views/buttonsView.js
@@ -110,7 +110,7 @@ define([
                 // Enable the button, make accessible and update aria labels and text
                 this.$('.buttons-action').a11y_cntrl_enabled(true).html(buttonText).attr('aria-label', ariaLabel);
 
-                // Make model answer button inaccessible (but still enabled) for visual users due to 
+                // Make model answer button inaccessible (but still enabled) for visual users due to
                 // the inability to represent selected incorrect/correct answers to a screen reader, may need revisiting
                 if (changedAttribute == BUTTON_STATE.SHOW_CORRECT_ANSWER || changedAttribute == BUTTON_STATE.HIDE_CORRECT_ANSWER) {
                     this.$('.buttons-action').a11y_cntrl(false);
@@ -152,7 +152,7 @@ define([
             }
 
             if (shouldDisplayAttempts) {
-                this.$('.buttons-display-inner').html(attemptsString);
+                this.$('.buttons-display-inner').html(attemptsString).a11y_text();
             }
 
         },
@@ -164,7 +164,7 @@ define([
                 .removeClass('display-none')
                 .addClass(this.model.get('_isCorrect') ? 'icon-tick' : 'icon-cross');
         },
-        
+
         refresh: function() {
             this.updateAttemptsCount();
             this.checkResetSubmittedState();
--- a/src/core/js/views/buttonsView.js
+++ b/src/core/js/views/buttonsView.js
@@ -152,7 +152,7 @@ define([
             }
 
             if (shouldDisplayAttempts) {
-                this.$('.buttons-display-inner').html(attemptsString).a11y_text();
+                this.$('.buttons-display-inner').a11y_text(attemptsString);
             }
 
         },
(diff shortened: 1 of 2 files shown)


src/core/templates/buttons.hbs
```diff
@@ -11,7 +11,7 @@
     <div class="buttons-display">
         <div class="buttons-marking-icon icon display-none">
         </div>
-        <div class="buttons-display-inner">
+        <div class="buttons-display-inner" aria-live="rude">
```
oliverfoster: as far as i remember few browse and screen reader combos respect aria-live tags, i think it might be almost completely pointless adding them in event though they are defined in the spec. which platforms, browsers and screen readers have you tested?

oliverfoster: https://www.paciellogroup.com/blog/2014/03/screen-reader-support-aria-live-regions/ < old but probably still relevant

ryan-lynch: Hi @oliverfoster I have updated to use "assertive" as per the aria spec. I have also tested this out on IE11 with JAWS 18 as well as iOS on Safari as well as Chrome browser with ChromeVox and they all work fine. I personally think this is quite an improvement in terms of accessibility.

oliverfoster: can you also do macosx + voiceover? if that works, can add a section about aria-live to the accessibility wiki in adapt? then we can all know how to use it and that it's fine to use.

oliverfoster: there are a few other places where it'd come in handy...

ryan-lynch: I'll fire that up now and test it out. Yeah, you're right there are definitely a lot of improvements to be made when it comes to accessibility.

oliverfoster: beautiful, ta

ryan-lynch: Ok @oliverfoster so this is a bit strange. It doesn't seem to behave itself on Mac osx + safari + voiceover, I was able to get it working intermittently but not consistently. However, on chrome + voiceover it works perfectly without any issues. It's odd because it also seems to work consistently on safari on iOS + iOS voiceover. 

I think this suggests that there could be a bug in Safari because I did hear it several times, intermittently.

oliverfoster: ok, well... 4/5 isn't bad, use or don't use? what do you reckon?

ryan-lynch: I think we should use it, definitely. The "attempts remaining" element is at least tabbable, which it wasn't before now. So if for some reason the user decides to use Safari, then the element is still accessible in some way. I can't think of any other ways to include it, but I'm new to accessibility. 

I would also suggest that the user should get some feedback from their answer when they click submit on a question component. Currently there is no feedback at all, so if they do get the question correct, a partially sighted user wouldn't even know.

oliverfoster: there is an aria label after the component which tells them.

ryan-lynch: oh I know there is a "component complete, correct" but I meant something like an aria-live which reads it out automatically as there isn't really a guarantee that the user tabs down that far. Just a thought.

oliverfoster: it also doesn't seem like there's a guarantee it'll read out using aria-live... always a rock and a hard place, at least when it's hard update and available to tab over the potential for it to read is always there. perhaps adding aria-live to the label at the bottom of the component is in order? so we get both if the aria-live is reading.

ryan-lynch: Yes I get that, the accessibility state element has a default value of "component title Incomplete". aria-live only reads out when the content of the element changes. To make this read out in both scenarios (a correct or incorrect answer), I think maybe this would need to be changed to only contain the state after the user has provided an answer of some description. Certainly some additional thought needs to go in to it.

We should raise that as a separate accessibility improvement.

src/core/js/views/buttonsView.js
```diff
@@ -152,7 +152,7 @@ define([
             }
 
             if (shouldDisplayAttempts) {
-                this.$('.buttons-display-inner').html(attemptsString);
+                this.$('.buttons-display-inner').html(attemptsString).a11y_text();
```
oliverfoster: can you flip this over to use @danielstorey 's new style?
```javascript
this.$('.buttons-display-inner').a11y_text(attemptsString);
```

ryan-lynch: Yeah, will do.

Issue #1723, opened 2017-08-16
jquery.a11y - insert accessible text with single call


danielstorey:
It's quite common for components to dynamically insert text using `$(selector).html(someText)`

In order to make this text accessible we need to add another call `.a11y_text()`

Wouldn't it be nice if one could accomplish this in one single call, perhaps by passing an argument into the above i.e. `.a11y_text(someText)`?!

oliverfoster:
as mentioned on previous issue, i like this idea, the syntax below does the same thing a little less concisely:

```javascript
$(selector).html($.a11y_text(someText));
$(selector).html(someText).a11y_text();
```

danielstorey:
I started doing that for a component I'm working on, then when it got to the 8th one I thought a helper might be called for!

PR #1725, opened 2017-08-16
Addresses #1723

danielstorey:
#1723 

Allow an optional argument to be passed to `$.fn.a11y_text` which when specified will insert and accessible version the given text rather than convert the existing content


ryan-lynch:
+1

Issue #1727, opened 2017-08-16
jquery.a11y - improve combination of style elements and anchor tags


danielstorey:
Steps to replicate:
Create a text component whose body contains formatted text e.g. `<strong>` elements,  as well at least one link  (`<a>` element)
Turn accessibility on

Current behaviour: because a11y has detected a non-style element (`<a>`) it now creates separate tabbable elements for all style elements

Desired behaviour: links should be tabbable and all other text should be grouped together in a single tabbable block

oliverfoster:
We may have come across this before... Does the outcome change if you wrap the whole thing in a ``<p>`` ?

danielstorey:
Yeah, still the same. I understand the challenge involved in trying to achieve the above, especially if the `<a>` is wrapped in a `<strong>` with some other text. Maybe there's a middle ground if the desired behaviour as I've described is not achievable. I'll have a think about possible solutions.

chris-steele:
I agree it's worth revisiting this because of the real impact, particularly for screen reader users, if the read order is punctuated unnaturally. I'll have a think about this and see if I can describe the behaviours we want with this functionality and once we're agreed we can try and implement (good bit of BDD!). This said, there's always the edge cases we miss so let's also implement a feature that allows us to state explicitly in the data how we want the text to be read: e.g. wrapping the blocks of content we want to be a single tabbable unit in a class that will override the default functionality. This will save users having to wait for fixes/updates or hack the code.

oliverfoster:
could you try this extension https://github.com/cgkineo/adapt-a11yText
see if that helps the reading situation? @danielstorey

moloko:
done in 64f635a076cb08cff312e208075f59e17ba9e835

PR #1731, opened 2017-08-17
Addresses #1727

danielstorey:
#1727 

* Combine links and style elements without having to tab through each one.

* Don't create tabbable text if the content is just spaces, commas and full stops

Seems to work well but needs thorough testing

### DIFF
--- a/src/core/js/libraries/jquery.a11y.js
+++ b/src/core/js/libraries/jquery.a11y.js
@@ -199,7 +199,6 @@
                 //CAPTURE DOMNODE CHILDREN
                 var children = $element.children();
 
-                
                 if (children.length === 0) {
                     //IF NO CHILDREN, ASSUME TEXT ONLY, WRAP IN SPAN TAG
                     var textContent = $element.text();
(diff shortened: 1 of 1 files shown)


src/core/js/libraries/jquery.a11y.js
```diff
@@ -221,36 +220,49 @@
...
+                            var $child = $(cloneChild);
+                            //IGNORE STYLING ELEMENTS
+                            if ($child.is(domSelectors.wrapStyleElements)) {
+                                currentText += child.outerHTML;
                             } else {
-                                //DESCEND INTO NODES WITH CHILDREN
-                                makeChildNodesAccessible($child);
+                                //APPEND CURRENT TEXT VALUE BEFORE CONTINUING
+                                if (currentText.length > 0) {
+                                    //MAKE TABBABLE IF THE TEXT CONTAINS CHARACTERS OTHER THAN FULL STOPS, COMMAS & SPACES
```
oliverfoster: this might be a bit too English language specific? why did you choose these characters and are there any potential others in different languages that could have the same impact on a screen reader?

danielstorey: Yeah good question. The examples I was working on had ", " in between different links and it seemed a bit pointless tabbing to text just containing these characters. A comma or fullstop is a pause which you'd get anyway as a the screen reader would stop at the end of the previous element. Other punctuation like "?" or "!"  would have more meaning and so there'd perhaps be more value in a screen reader reading these characters. 

As for other languages, I couldn't say so may need to get a wider consensus on which characters are used here.

oliverfoster:
@danielstorey can you have a look at #1948 see if it solves this issue?

oliverfoster:
closing as out of date and probably fixed in #1948

Issue #1738, opened 2017-08-18
Core: Page.hbs - No instruction text available


guywillis:
There is no support for the instruction text field for the page template file. 

- [x] Update page.hbs file
- [x] Update contentobject.model.schema file

PR #1739, opened 2017-08-18
Resolves Issue/1738

guywillis:
https://github.com/adaptlearning/adapt_framework/issues/1738


sarveshwar-gavhane:
@guywillis 
would it be nice to add  ``` instruction``` to here https://github.com/adaptlearning/adapt_framework/blob/master/src/course/en/contentObjects.json as well ??

guywillis:
Agreed, it would. Thank you for highlighting it! :)

Issue #1435, opened 2017-02-23
review all the core .schema files


moloko:
There's a few entries in course.model.schema that don't look to be used any more and I'm sure there will be others.

Similarly there's some properties that are missing (e.g. `_isVisible`) and some discrepancies between the versions in the framework and those in the AT - for example contentobject.model.schema has `pageBody` in the framework version but not in the authoring tool version.

moloko:
Ones we've found so far:
* `_ariaLabels.menu`
* `_ariaLabels.menuItem`
* `_ariaLabels.menuViewButton`
* `_ariaLabels.navigationBack`
* `_ariaLabels.closeResources`
* `_ariaLabels.drawerBack` (`_ariaLabels.previous` is what's currently assigned to the Drawer's back button in drawer.hbs)
* `_ariaLabels.menuBack`

moloko:
`instruction` seems to be missing for both article and block - it's referenced in article.hbs and block.hbs but not in the .schema files

moloko:
AT develop branch has had `_onScreen` added to contentObject/article/block/component schemas, need to add this to the framework's files

PR #1740, opened 2017-08-18
Issue/1435 - Add instruction to article / block schema

guywillis:
https://github.com/adaptlearning/adapt_framework/issues/1435

### DIFF
--- a/src/core/schema/article.model.schema
+++ b/src/core/schema/article.model.schema
@@ -31,10 +31,10 @@
     "instruction": {
       "type": "string",
       "required": false,
+      "title": "Instruction text",
       "default": "",
       "inputType": "Text",
       "validators": [],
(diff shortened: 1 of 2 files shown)


src/core/schema/article.model.schema
```diff
@@ -28,6 +28,15 @@
       "validators": [],
       "translatable": true
     },
+    "instruction": {
+      "type": "string",
+      "required": false,
+      "default": "",
+      "inputType": "Text",
+      "validators": [],
+      "help": "This is the instruction text",
```
moloko: Sorry I know all the others are exactly like this, but I do think these would look better with a title of 'Instruction text' and no help text...

I don't know why I didn't change these when I did all the select-to-boolean changes

PR #1751, opened 2017-08-21
Resolves #1708, removes v1 button hacks

### LINKED ISSUE #1708: Remove hack in questionModel.js for v1-v2 buttons compatibility
The code here is only required to transition v1 courses to v2, so can probably be dropped for a v3 release:
https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/models/questionModel.js#L51-L92

brian-learningpool:
Removed the version 1->2 button hacks in preparation for the release of version 3.0.

https://github.com/adaptlearning/adapt_framework/issues/1708


src/core/js/models/questionModel.js
```diff
@@ -40,55 +40,30 @@ define([
         setupButtonSettings: function() {
             var globalButtons = Adapt.course.get("_buttons");
```
moloko: just a recommendation and not a requirement... but would be a good opportunity to use `const` here instead of `var`

src/core/js/models/questionModel.js
```diff
@@ -40,55 +40,30 @@ define([
...
+                // Check all the components buttons.
+                // If they are empty use the global defaults.
                 var componentButtons = this.get("_buttons");
 
-                if (typeof componentButtons.submit == 'undefined') {
-                    for (var key in componentButtons) {
-                        if (typeof componentButtons[key] == 'object') {
-                          // ARIA labels
-                          if (!componentButtons[key].buttonText && globalButtons[key].buttonText) {
+                for (var key in componentButtons) {
```
moloko: just a recommendation and not a requirement... but would be a good opportunity to use `const` here instead of `var`

moloko:
@brian-learningpool you have a conflict that needs sorting out I'm afraid

PR #1758, opened 2017-08-23
Resolves #1757, minor tidy up of router.js

### LINKED ISSUE #1757: Tidy up of router.js
While having a look at #1752 I noticed that router.js could use a minor tidy up, specifically:
- code more verbose in places than it has to be
- calls to `console.log()`
- typos in comments
- inconsistencies in use of quote marks

brian-learningpool:
- Replaced console.log() with Adapt.log.warn()
- Removed some duplicated code
- Corrected comment typos
- Fixed inconsistent use of quote marks
- Corrected spacing in curly braces objects for readability


src/core/js/router.js
```diff
@@ -207,45 +216,43 @@ define([
...
                     break;
                 case 2:
-                    this.navigate("#/"+args[0]+"/"+args[1], {trigger:false, replace:false});
-                    break;
                 case 3:
-                    this.navigate("#/"+args[0]+"/"+args[1]+"/"+args[2], {trigger:false, replace:false});
+                    this.navigate('#/' + args.join('/'), options);
                     break;
+                default:
+                    this.handleRoute.apply(this, args);
```
oliverfoster: Are you using this function? @brian-learningpool cos it really needs to be removed.
I would prefer a deprecated warning at the top of this one.

brian-learningpool: No, I'm not using it for anything, @oliverfoster.  I'll add a `warn`.

oliverfoster: Cool. just something like 'use Backbone.history.navigate or window.location.href instead of Adapt.trigger("router:navigateTo");'

PR #1771, opened 2017-09-06
Fixes #1690, resolves console issue when content is rendered in frame…

### LINKED ISSUE #1690: Sample course does not render inside IFRAME on IE 9
It's throwing this error in the console while the page gets stuck on the loading screen:
`Object doesn't support property or method 'apply'`

I think it's coming from the `_addToConsole` function as if I add a try-catch on the call to `console.log.apply()` the course renders fine.

AaronmcL:
…set in IE 9

### DIFF
--- a/src/core/js/logging.js
+++ b/src/core/js/logging.js
@@ -97,8 +97,12 @@ define([
             var log = [level.asUpperCase + ':'];
             data && log.push.apply(log, data);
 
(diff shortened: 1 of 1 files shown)


brian-learningpool:
Already resolved in https://github.com/adaptlearning/adapt_framework/pull/1612

PR #1772, opened 2017-09-07
issue/1407: grunt check for less/plugins/{PLUGIN} in each plugin

### LINKED ISSUE #1407: Omit plugin .less files from Grunt build
When the grunt builds it only includes the plugins that are required - thus saving many selectors helping avoid the ie9 limit. 

But within the vanilla theme there are less files for many components, accordion.less, mcq.less, etc that are compiled even if that component is not included in the course, bloating the course css.

With regards to the grunt build, can we add a further level of filter on the theme less files to exclude the less that's not required?

I understand that this may require that the vanilla less files be renamed, for example, accordion.less be renamed to adapt-contrib-accordion/accordion.less

nilslp:
#1407 

Looks in plugins directory within the less directory in theme and compares against a list of included plugin names as to whether or not to include the less files associated with the included plugins.

Directory structure of theme has been changed in another commit on another forked repository of adapt-contrib-vanilla by @ciaranlp.

It now follows the structure:

```
less
   |_ defaults
   |_ plugin
        |_ adapt-contrib-accordion _ ...files
        |_ adapt-contrib-boxMenu _ ...files
   |_ variables
```


grunt/helpers.js
```diff
@@ -196,11 +196,30 @@ module.exports = function(grunt) {
...
+        var isIncluded;
+        var isExcluded;
+
         // carry on as normal if no includes/excludes
         if (!includes && !excludes) return true;
 
-        var isIncluded = includes && pluginPath.search(exports.getIncludedRegExp()) !== -1;
-        var isExcluded = excludes && pluginPath.search(exports.getExcludedRegExp()) !== -1;
+        // get plugins directory within theme/less
+        var compareVar = grunt.file.doesPathContain((grunt.config('sourcedir') + 'theme' + path.sep + grunt.config('theme') + path.sep + 'less' + path.sep + 'src' + path.sep + 'plugins'), pluginPath);
```
barry-learningpool: Might be cleaner to just use https://nodejs.org/api/path.html#path_path_join_paths here

`path.join(grunt.config('sourcedir'), 'theme', grunt.config('theme'), 'less', 'src', 'plugins')`

grunt/helpers.js
```diff
@@ -196,11 +196,30 @@ module.exports = function(grunt) {
...
+
+        // if plugin directory within theme is in the filepath
+        if (compareVar) {
```
barry-learningpool: 'compareVar' isn't a very meaningful name, what about something like 'themeHasPlugin'?

grunt/helpers.js
```diff
@@ -196,11 +196,30 @@ module.exports = function(grunt) {
...
+
+            // plugins dir is in the file path so check against the list of includes
+            var bits = [];
+            bits = pluginPath.split("/"); // split the file path
```
brian-learningpool: Question: should the split not be `path.sep` rather than `/`?

grunt/helpers.js
```diff
@@ -196,11 +196,30 @@ module.exports = function(grunt) {
         var includes = grunt.config('includes');
         var excludes = grunt.config('excludes');
 
+        var isIncluded;
```
barry-learningpool: Are both of these variables required? Wouldn't just having 'isIncluded' (true/false) suffice?

grunt/helpers.js
brian-learningpool: Can you use `path.join()` here instead of all the concatenation?

Also, `compareVar` would be better called `isThemePluginsFolder` for readability

grunt/helpers.js
```diff
@@ -196,11 +196,30 @@ module.exports = function(grunt) {
...
+        // if plugin directory within theme is in the filepath
+        if (compareVar) {
+
+            // plugins dir is in the file path so check against the list of includes
+            var bits = [];
+            bits = pluginPath.split("/"); // split the file path
+            var match = bits[bits.length - 2]; // get plugin directory name
+
+            // check the plugin directory name against the list of included plugin names
+            (includes.indexOf(match) !== - 1) ? isIncluded = true : isExcluded = true;
```
brian-learningpool: This is using the ternary operator but it's not a ternary statement.  Can you re-write as an `if-else`?

grunt/helpers.js
```diff
@@ -196,11 +196,30 @@ module.exports = function(grunt) {
...
+        // get plugins directory within theme/less
+        var compareVar = grunt.file.doesPathContain((grunt.config('sourcedir') + 'theme' + path.sep + grunt.config('theme') + path.sep + 'less' + path.sep + 'src' + path.sep + 'plugins'), pluginPath);
+
+        // if plugin directory within theme is in the filepath
+        if (compareVar) {
+
+            // plugins dir is in the file path so check against the list of includes
+            var bits = [];
+            bits = pluginPath.split("/"); // split the file path
+            var match = bits[bits.length - 2]; // get plugin directory name
```
brian-learningpool: Does the -2 here assume here that there will be no further directories beneath the plugins folder?

grunt/helpers.js
nilslp: Cheers @barry-learningpool I guess would be, I was just following on with the existing code trying to make as little impact as I could.

Would this be a recommended improvement then?

grunt/helpers.js
nilslp: Yeah great recommendation. Is there a preference for name between @brian-learningpool's recommendation of `isThemePluginsFolder` and your's `themeHasPlugin`?

grunt/helpers.js
nilslp: Thanks @brian-learningpool will make the `path.join()` change.

Could you maybe voice your opinion on the naming of the variable in the other comment please?

grunt/helpers.js
nilslp: Yes it should. I will test that though just to make sure it still performs the split and make the change if so. Thanks

grunt/helpers.js
barry-learningpool: Yeah that's fair enough @nilslp , would be an improvement

grunt/helpers.js
nilslp: Will do. Thanks

grunt/helpers.js
barry-learningpool: Up to you! Either is good

grunt/helpers.js
nilslp: It does. This isn't an argument for it but I am not sure there should be other directories within a plugins folder for less (unless there was to be say an index.less, that included multiple less files from within a sub-directory).

Do you have a recommended better approach I could attempt, for stripping out the plugin name from the filePath?

oliverfoster:
is this behaviour just for plugins in the ``theme`` directory? 
if so, it needs to be changed so that it is global to all plugins.
plugins should be treated equally otherwise we get into a situation where we have to explain all the exceptions and that's a big waste of time.

grunt/helpers.js
brian-learningpool: I prefer the prefix of the style is/has/was because it's immediately obvious that it's a boolean.

nilslp:
Updated code to have a more specific RegEx check and the code now loops through each plugin to see if there are any nested `/less/src/plugins/` plugin directories.

grunt/helpers.js
```diff
@@ -41,13 +41,29 @@ module.exports = function(grunt) {
     });
 
     // privates
-    var generateIncludedRegExp = function() {
+    var generateIncludedRegExp = function(removeSrc = false) {
```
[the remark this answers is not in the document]
nilslp: Sorted. Has been removed in the most recent commit.

grunt/helpers.js
```diff
@@ -238,9 +275,9 @@ module.exports = function(grunt) {
         else return content;
     };
 
-    exports.getIncludedRegExp = function() {
-        var configValue = grunt.config('includedRegExp');
-        return configValue || grunt.config('includedRegExp', generateIncludedRegExp());
+    exports.getIncludedRegExp = function(removeSrc) {
+        if (removeSrc) return grunt.config('includedRegExp', generateIncludedRegExp(removeSrc));
```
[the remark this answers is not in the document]
nilslp: Sorted. Functions were restructured.

danielstorey:
I just tested this by adding a less file with a random name into the theme/less/src/plugins folder. I expected the file to not be included but it was, and the style applied.

On another note several exports don't seem to be used outside of this file (including ones already present before this PR) so could they just be locally declared functions instead? E.g. `getScriptSafeRegExp`, `getExcludedRegExp`...

nilslp:
Hi @danielstorey, can you please check definitely on the correct branch and uptodate.

I have retested this based on what you said and have uncommented the output in helpers.js so I could post the output of it:

* Included /Users/nilssanderson/LearningPool/adapt-authoring/temp/59c26435342ab51256c9fe74/adapt_framework/src/menu/adapt-contrib-boxMenu/less/boxmenu.less
* Included /Users/nilssanderson/LearningPool/adapt-authoring/temp/59c26435342ab51256c9fe74/adapt_framework/src/menu/adapt-contrib-boxMenu/less/src/plugins/adapt-contrib-text/text.less
* Excluded /Users/nilssanderson/LearningPool/adapt-authoring/temp/59c26435342ab51256c9fe74/adapt_framework/src/menu/adapt-contrib-boxMenu/less/src/plugins/temp.less

Everything still seems to be working fine for me.

If you are still having issues can you please check the names of the directories and if all still not working give me more details so I can test with the exact same configuration as yourself please.

danielstorey:
Initially my error due to misunderstanding - I had no "includes" in build config. However I have just tested again and have found that it works on mac but not windows. On windows everything gets excluded!

oliverfoster:
Will test cross platform, ta.

nilslp:
Thanks @danielstorey.

Must be something to do with the slashes (thought there was a function handling all the conversions of slashes so it works cross platform but might need to check that).

Will await @oliverfoster comments and then look into that issue.

Thanks again for taking the time!

grunt/helpers.js
```diff
@@ -245,6 +279,11 @@ module.exports = function(grunt) {
...
     };
 
+    exports.getNestedIncludedRegExp = function() {
+        var configValue = grunt.config('nestedIncludedRegExp');
+        return configValue || grunt.config('nestedIncludedRegExp', generateNestedIncludedRegExp());
+    };
+
     exports.getExcludedRegExp = function() {
         var configValue = grunt.config('excludedRegExp');
         return configValue || grunt.config('excludedRegExp', generateExcludedRegExp());
```
oliverfoster: make the lower functions private as @danielstorey suggests

grunt/helpers.js
```diff
@@ -43,14 +43,28 @@ module.exports = function(grunt) {
...
         var includes = grunt.config('includes') || [];
-        var re = '';
-        for(var i = 0, count = includes.length; i < count; i++) {
-            re += '\/' + includes[i].toLowerCase() + '\/';
-            if(i < includes.length-1) re += '|';
-        }
+        var pluginTypes = exports.defaults.pluginTypes;
+
+        // Return a more specific plugin regExp including src path.
+        var re = _.map(includes, function(plugin) {
```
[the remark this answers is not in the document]
nilssanderson: Done

grunt/helpers.js
```diff
@@ -43,14 +43,28 @@ module.exports = function(grunt) {
...
+            return _.map(pluginTypes, function(type) {
+                return exports.defaults.sourcedir + type + '\/' + plugin + '\/';
```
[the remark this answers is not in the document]
nilssanderson: Done, after `exports.defaults.sourcedir` and `exports.defaults.outputdir` are set.

grunt/helpers.js
```diff
@@ -43,14 +43,28 @@ module.exports = function(grunt) {
...
+            }).join('|');
+        }).join('|');
         return new RegExp(re, "i");
     };
 
+    var generateNestedIncludedRegExp = function() {
+        var includes = grunt.config('includes') || [];
+        var folderRegEx = /(\/less\/src\/plugins)/;
```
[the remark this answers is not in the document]
nilssanderson: Done

grunt/helpers.js
```diff
@@ -201,17 +215,37 @@ module.exports = function(grunt) {
         // carry on as normal if no includes/excludes
         if (!includes && !excludes) return true;
 
+        // Very basic check to see if the file path string contains any
```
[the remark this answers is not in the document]
nilssanderson: Done

grunt/helpers.js
```diff
@@ -201,17 +215,37 @@ module.exports = function(grunt) {
...
         if (isExcluded || isIncluded === false) {
             // grunt.log.writeln('Excluded ' + chalk.red(pluginPath));
             return false;
         }
-        else {
+
+        // Check the LESS plugins folder exists.
+        // The LESS 'plugins' folder doesn't exist, so add the file,
+        // as the plugin has already been found in the previous check.
+        var nestedPluginsPath = !!pluginPath.match(/(?:.)+(?:\/less\/src\/plugins)/g);
```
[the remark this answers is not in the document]
nilssanderson: Done

grunt/helpers.js
nilssanderson: Done

nilssanderson:
Hi, @oliverfoster and @danielstorey I have updated based on your reviews.
If you both get a chance could you give it another test, please.

Thanks

grunt/helpers.js
```diff
@@ -43,21 +43,41 @@ module.exports = function(grunt) {
...
+
+    var generateNestedIncludedRegExp = function() {
+        var includes = grunt.config('includes') || [];
+        var pluginTypes = exports.defaults.pluginTypes;
+        var folderRegEx = /(\/less\/plugins)/;
+
+        // Return a more specific plugin regExp including src path.
+        var re = _.map(includes, function(plugin) {
+            return _.map(pluginTypes, function(type) {
+                return exports.defaults.sourcedir + type + '\/' + plugin + '\/' + folderRegEx;
```
oliverfoster:
```jaavascript
exports.defaults.sourcedir + type + '\/' + plugin + '\/' + folderRegEx;
```
^ this line
```javascript
exports.defaults = "C:/path_to_files/";
type = "components" || "extensions" || "menu" || "theme";
plugin = "adapt-contrib-pluginName";
folderRegEx = /(/less/plugins)/;
```
folderRegExp evaluates to this string
```javascript
"\/(\/less/plugins)\/"
```
so the regex strings come out looking like this:
```javascript
"C:/path_to_files/components/adapt-contrib-pluginName//(/less/plugins)/"
```
when that parses a new regex, it's coming out like the following for every plugin:
```javascript
"C:/path_to_files/components/adapt-contrib-pluginName///less/plugins/"
```
to get rid of that triple slash we need folderRegExp to not be a regex parsed into a string and instead can be a string directly:
```javascript
folderRegEx = "less/plugins";
```
But more than than, we're trying to match the following inside every other plugin:
```javascript
"/less/plugin/adapt-contrib-pluginName/"
```
i think generateNestedIncludeRegExp needs to be:
```javscript
    var generateNestedIncludedRegExp = function() {
        var includes = grunt.config('includes') || [];
        var folderRegEx = "less/plugins/";

        // Return a more specific plugin regExp including src path.
        var re = _.map(includes, function(plugin) {
            return exports.defaults.sourcedir + '([^\/]*)\/([^\/]*)\/' + folderRegEx + plugin + '\/';
        }).join('|');
        return new RegExp(re, "i");
    };
```
which is the equivalent of trying to match:
```javascript
path_to_src/*/*/less/plugins/adapt-pluginName/
```

nilslp: Thanks @oliverfoster have pushed this change after testing it 👍

nilslp:
Thanks @oliverfoster have updated on this commit: https://github.com/adaptlearning/adapt_framework/pull/1772/commits/e3d09e256dbe21ea05050532d6db2d36fee0f85b

oliverfoster:
perfect 👍 
image

image

nilslp:
Great work all 👍

oliverfoster:
@brian-learningpool  can we have your seal?

oliverfoster:
@lc-thomasberger did you manage to have a look at this?

Issue #1777, opened 2017-09-09
grunt: translate task broken if schema destination doesn't exist


oliverfoster:
Example: ```course/course/_globals/_extensions/_languagePicker/navigationBarLabel/``` 

1. The ``schema:defaults`` task adds ``schema.defaults`` to the ``build/course``.
2. The ``translate:export`` task exports from the ``src/course`` and ``schema.defaults`` all of the translatable texts.
3. The ``translate:import`` task imports into ``src/course`` only.
4. When the translate:import task imports into ``src/course``, it will try to write the schema-defaults from the export. The ``schema.defaults`` json structure may never have existed in the ``src/course`` json and so this can create ``"cannot add navigationBarLabel to undefined"`` style errors. As in the above example, the path ```course/course/_globals/_extensions/_languagePicker``` has never existed in the ``src/course`` json.

It's possible for a nest name:value to exist in the ``build/course`` from the ``schema-defaults`` task but not exist in the ``src/course`` data.

To fix, either/both:
a. Allow ``translate:import`` to create missing parents.
b. Make ``schema-defaults`` task perform before the ``copy`` task on ``src/course`` or the ``%outputdir%`` when specified.

PR #1779, opened 2017-09-09
issue/1777: grunt fixed schema issue with translate task

oliverfoster:
#1777 

* fixes issue with translate task and schema defaults

### DIFF
--- a/grunt/tasks/schema-defaults.js
+++ b/grunt/tasks/schema-defaults.js
@@ -54,8 +54,7 @@ module.exports = function(grunt) {
             });
         });
 
-        var sourcedir = grunt.config("sourcedir");
-        if (grunt.option("outputdir")) sourcedir = grunt.option('outputdir');
+        var sourcedir = grunt.option('outputdir') || grunt.config("sourcedir");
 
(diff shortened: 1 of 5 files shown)


grunt/tasks/schema-defaults.js
```diff
@@ -54,8 +54,11 @@ module.exports = function(grunt) {
             });
         });
 
+        var sourcedir = grunt.config("sourcedir");
```
tomgreenfield:
```js
var sourcedir = grunt.option('outputdir') || grunt.config('sourcedir');
```
?

oliverfoster: var sourcedir = the location of the course.json

grunt/tasks/schema-defaults.js
```diff
@@ -54,8 +54,10 @@ module.exports = function(grunt) {
             });
         });
 
+        var sourcedir = grunt.option('outputdir') || grunt.config("sourcedir");
+
         //iterate through lanugage folders
```
moloko: typo 'lanugage'

Issue #1781, opened 2017-09-10
grunt: copy doesn't respect %outputdir%


oliverfoster:
it will copy json from src/course over json at ``%outputdir%``

Issue #1784, opened 2017-09-10
grunt: helpers generateConfigData fails on invalid config.json


oliverfoster:
it uses ``require(fileName)`` to read in json, rather than ``grunt.file.readJSON(fileName)`` which has a lot more error checking bits too it.

PR #1787, opened 2017-09-10
issue/1781: grunt copy task respects outputdir config

oliverfoster:
#1781 

copy task now respects outputdir config

### DIFF
--- a/grunt/config/copy.js
+++ b/grunt/config/copy.js
@@ -1,6 +1,7 @@
 module.exports = function (grunt, options) {
     
-    
+    var _ = require("underscore");
+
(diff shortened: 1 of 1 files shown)


moloko:
So, this works if the directory specified by `%outputdir%` exists and is already populated with an Adapt build. 

But if that directory doesn't exist I get the following error:
issue-1781_error

Apologies if that's as expected, the PR/issue is a bit short of information as to the intention of this change.

oliverfoster:
yup, that's as intended

oliverfoster:
https://github.com/adaptlearning/adapt_framework/issues/1784 < this is the one about invalid config.json

Issue #1783, opened 2017-09-10
grunt: copy task flattens unnecessarily on collated folders


oliverfoster:
# Current
all the fonts, assets and core/js/libraries folders get flattened at their destination. 

* plugin using them cannot segregate files so plugin developers must always use distinct filenames
* references to files which are in hierarchy at source are flattened at destination. this means we have to consider the flattening whilst writing plugins

# Ideal
all sub folders in the fonts/ libraries/ required/ and assets/ at source should remain at destination

taylortom:
I'll have a look into how this affects the authoring tool when I get a min

oliverfoster:
it shouldn't have any impact unless the AT uses plugins which have hierarchical assets / fonts folders that are then referenced by the css or json later. most of the AT assets should be output to the course/en/ folder.

PR #1789, opened 2017-09-10
issue/1783: grunt prevent flatten of fonts/ libraries/ required/ assets/

oliverfoster:
#1783


oliverfoster:
they were fine before, i've just switched them to all use the same function.

PR #1524, opened 2017-04-11
issue/1523: added xapi model state functions

oliverfoster:
as per #1523

PR #1797, opened 2017-09-13
Resolves #1700, moves completion criteria to config.json 

### LINKED ISSUE #1700: Should completion settings move to config.json?
Currently there are course completion attributes which live inside the Spoor extension.  For v3 should these be attributes of the course rather than a specific plugin?

brian-learningpool:
Added a new tracking module which reads the completion criteria from config.json.

This work was branched off https://github.com/adaptlearning/adapt_framework/pull/1524 as part of the xAPI extension development.


src/core/js/tracking.js
```diff
@@ -0,0 +1,123 @@
...
+], function(Adapt, COMPLETION_STATE) {
+
+    var Tracking = Backbone.Controller.extend({
+
+        _config: {
+            _requireContentCompleted: true,
+            _requireAssessmentPassed: false
+        },
+
+        _completionCriteria: [],
```
oliverfoster: i'm not sure this ``_completionCriteria`` abstraction is useful or necessary

```
define([
    'core/js/adapt',
    'core/js/enums/completionStateEnum'
], function(Adapt, COMPLETION_STATE) {

    var Tracking = Backbone.Controller.extend({

        _config: {
            _requireContentCompleted: true,
            _requireAssessmentPassed: false
        },

        _completionCriteria: [],

        initialize: function() {
            Adapt.once('configModel:loadCourseData', this.loadConfig.bind(this));
            Adapt.on('app:dataReady', this.setupEventListeners.bind(this));
        },

            // Check if completion requires passing an assessment. 
            if (this._config._requireAssessmentPassed) {
                // Listen for the 'assessment:complete' event.
                this.listenTo(Adapt, 'assessment:complete', this.onAssessmentComplete);

            }

            // Check if completion requires completing all content.
            if (this._config._requireContentCompleted) {
                // Listen for the 'course:_isComplete' event.
                this.listenTo(Adapt.course, 'change:_isComplete', this.checkCompletion);
            }

        },

        _assessmentState: null,
        onAssessmentComplete: function(assessment) {

            this._assessmentState = assessment
            this.checkCompletion();

        },

        checkCompletion: function() {

            if (this._config._requireAssessmentPassed && (this._assessmentState.attemptsLeft !== 0 ||  this._assessmentState.isPass !=== true)) {
                return;
            }

            if (this._config._requireContentCompleted && !Adapt.course.get("_isComplete")) {
                return;
            }

            var completionData = {
                status: null,
                assessment: null
            };

            if (this._config._requireAssessmentPassed) {
                completionData.status = this._assessmentState.isPass ? COMPLETION_STATE.PASS : COMPLETION_STATE.FAIL;
                completionData.assessment = this._assessmentState
            } else {
                completionData.status = COMPLETION_STATE.COMPLETE;
            }

            Adapt.trigger('tracking:complete', completionData);
            Adapt.log.debug('tracking:complete', completionData);

        },

        /**
         * Set the _config object to the values retrieved from config.json.
         */
        loadConfig: function() {
            if (Adapt.config.has('_completionCriteria')) {
                this._config = Adapt.config.get('_completionCriteria');
            }
        }

    });

    Adapt.tracking = new Tracking();

    return Adapt.tracking;
});
```

src/core/js/tracking.js
```diff
@@ -0,0 +1,123 @@
...
+        setupEventListeners: function() {
+            var self = this;
+
+            // Check if completion requires passing an assessment. 
+            if (this._config._requireAssessmentPassed) {
+                // Add the assessment to the completion criteria.
+                this._completionCriteria.push({ type: 'assessment', isComplete: false, isPass: null });
+
+                // Listen for the 'assessment:complete' event.
+                this.listenTo(Adapt, 'assessment:complete', function(assessment) {
```
oliverfoster: these functions should be on the controller to prevent this deep nesting

src/core/js/tracking.js
```diff
@@ -0,0 +1,123 @@
...
+                    }
+                });
+            }
+
+            // Check if completion requires completing all content.
+            if (this._config._requireContentCompleted) {
+                // Add this to the completion criteria.
+                this._completionCriteria.push({ type: 'content', isComplete: false });
+
+                this.listenTo(Adapt.components, 'change:_isComplete', function (model, isComplete) {
```
oliverfoster: these functions should be on the controller to prevent deep nesting

src/core/js/tracking.js
```diff
@@ -0,0 +1,123 @@
...
+
+        _completionCriteria: [],
+
+        initialize: function() {
+            Adapt.once('configModel:loadCourseData', this.loadConfig.bind(this));
+            Adapt.on('app:dataReady', this.setupEventListeners.bind(this));
+        },
+
+        setupEventListeners: function() {
+            var self = this;
```
oliverfoster: self isn't needed, the ``this.listenTo`` function returns with the correct this context, that is it's function

src/core/js/tracking.js
```diff
@@ -0,0 +1,84 @@
...
+        onAssessmentComplete: function(assessmentState) {
+            this._assessmentState = assessmentState;
+
+            this.checkCompletion();
+        },
+
+        /**
+         * Evaluate the course and assessment completion.
+         */
+        checkCompletion: function() {
```
oliverfoster: Split into two functions?

```js
        getCompletionData: function() {

            var completionData = {
                status: COMPLETION_STATE.INCOMPLETE,
                assessment: null
            };

            if (this._config._requireAssessmentPassed && (!this._assessmentState || !this._assessmentState.isComplete)) {
                return completionData;
            }

            if (this._config._requireContentCompleted && !Adapt.course.get('_isComplete')) {
                return completionData;
            }

            if (this._config._requireAssessmentPassed) {
                completionData.status = this._assessmentState.isPass ? COMPLETION_STATE.PASS : COMPLETION_STATE.FAIL;
                completionData.assessment = this._assessmentState
            } else {
                completionData.status = COMPLETION_STATE.COMPLETE;
            }

            return completionData;
        },

        checkCompletion: function() {

            var completionData = this.getCompletionData();
            switch (completionData.status) {
                case COMPLETION_STATE.INCOMPLETE: 
                case COMPLETION_STATE.FAIL:
                    return;
            }

            Adapt.trigger('tracking:complete', completionData);
            Adapt.log.debug('tracking:complete', completionData);
            
        },
```

src/core/js/tracking.js
```diff
@@ -0,0 +1,84 @@
...
+            if (this._config._requireAssessmentPassed && !this._assessmentState.isComplete) {
```
oliverfoster: If this function is called because the course is complete, but the assessment hasn't been (odd usecase I know), then you'll get a "cannot find isComplete on undefined" error because the this._assessmentState object won't exist yet:
```js
if (this._config._requireAssessmentPassed && (!this._assessmentState || !this._assessmentState.isComplete)) {
```

brian-learningpool:
@oliverfoster, I think this PR should be ok now?

oliverfoster:
Yea, it looks fine. I haven't actually tested it yet. I will do in a bit.

oliverfoster:
@moloko i'd feel safer if you had a look at this

moloko:
this certainly looks OK but I'd like to give it a little test, presumably it would actually work OK as it stands in the current Adapt master? (Obviously it wouldn't send anything back to the LMS at this point.)

brian-learningpool:
Hi Matt, yes, it should work as is in the current `master` branch.  Please do give it a test when you get a moment though.

src/core/js/tracking.js
```diff
@@ -0,0 +1,111 @@
...
+         * @returns An object representing the user's course completion.
+         */
+        getCompletionData: function() {
+            var completionData = {
+                status: COMPLETION_STATE.INCOMPLETE,
+                assessment: null
+            };
+
+            // Course complete is required.
+            if (this._config._requireCourseComplete && !Adapt.course.get('_isComplete')) {
```
moloko: this needs changing to `_requireContentCompleted` as that's what's defined in the .schema

src/core/js/tracking.js
```diff
@@ -0,0 +1,111 @@
...
+            }
+
+            // Assessment completed required.
+            if (this._config._requireAssessmentCompleted) {
+                if (!this._assessmentState) {
+                    // INCOMPLETE: assessment is not complete.
+                    return completionData;
+                }
+
+                if (!this._assessmentState.isPass && this._assessmentState.attempts) {
```
moloko: `_assessmentState` doesn't have an `attempts` property, which means that if you have the criteria set to `_requireAssessmentCompleted` only and the assessment configured to have infinite attempts, you get set to a FAILED status as soon as you fail.

moloko: appreciate this is more an issue with the assessment extension that with this code...

brian-learningpool: There should be an `attempts` property returned by `getState()` here:
https://github.com/adaptlearning/adapt-contrib-assessment/blob/master/js/adapt-assessmentArticleModel.js#L777

In this case there's an early return, with a `status` of incomplete set.  The FAILED status is only set after a fail on the last attempt.

src/core/js/tracking.js
brian-learningpool: Good catch.  I think the property name went through a few iterations.

src/core/js/tracking.js
moloko: I think that's the state you get when _an_ assessment is completed not when _all_ of them are completed.
This is the `assessmentState` object that `onAssessmentComplete` in tracking.js gets:
pr_1797_assessment_state
To be honest, we're actually no worse off here than we are with the current Adapt code; it does exactly the same thing. Really the only issue here is that the code is trying to check something that doesn't actually exist, the outcome is no different to what we currently have.

Issue #1800, opened 2017-09-16
grunt: adapt.min.js not getting minified


oliverfoster:
https://github.com/adaptlearning/adapt_framework/blob/master/grunt/config/uglify.js#L14

PR #1801, opened 2017-09-16
issue/1800: minify adapt.min.js

oliverfoster:
#1800


moloko:
It looks to me very much like adapt.min.js IS minified.

If I do a before and after test with this PR, it does reduce the size of adapt.min.js by 11KB - but also increases execution time of the uglify task from ~6s to ~9s

oliverfoster:
i'll recheck, probably a brain error on my part.

Issue #1811, opened 2017-09-25
Can't determine if data from Spoor has been loaded


danielstorey:
The current way seems to be testing for offlineStorage's ready status [e.g. here](https://github.com/adaptlearning/adapt-contrib-languagePicker/blob/master/js/adapt-languagePicker.js#L31-L35). But what if the offlineStorage model is ready but the data hasn't been loaded yet?

If `plugins:ready` was more reliable we could listen for this but there are currently issues with that (#1810).

It would be great to have a specific trigger for this purpose

brian-learningpool:
Something like this would be good to have for the xAPI work.  

It would be very useful if there was a global event to indicate that the course state had been restored.

danielstorey:
See https://github.com/adaptlearning/adapt-contrib-spoor/pull/164

danielstorey:
Does simply listening to `adapt:start` resolve this?

oliverfoster:
Yes, it should do.

Issue #1810, opened 2017-09-25
Rework plugins:ready


danielstorey:
We need a reliable way of determining whether all plugins are ready.

Currently plugins:ready can fire a number of times. Add a bunch of extensions and stick a console.log [here ](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/adapt.js#L57).

In my understanding this is down to one plugin triggering `plugin:endWait` before another has had the chance to call `plugin:beginWait`

Maybe there needs to be stricter rules on when a plugin can trigger `plugin:beginWait`?

danielstorey:
I'll give that a go. The other issue is that if no plugins present use `plugin:beginWait/endWait` then listening for `plugins:ready` could leave you waiting a rather long time...

oliverfoster:
we'd want core to do at lease one beginWait + one endWait on loading and navigation

PR #1813, opened 2017-09-27
#1810 update plugin wait mechanism

danielstorey:
Also addresses #1811 as we can now listen to `plugins:ready` to determine if spoor, or xapi session data has been restored.


danielstorey:
See #1815

PR #1819, opened 2017-09-29
Fixes #1817, resolves race condition in Drawer (Patch for 'legacy')

### LINKED ISSUE #1817: Race condition with drawerView
This sometimes causes the drawer button to be visible even if there are no drawer items.

The reason for this is because drawerView is instantiated before navigationView.  The logic ([here](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/drawerView.js#L94)), which checks whether the button should be displayed or not, does not find the button in the DOM.

This is because the view is created on `app:dataReady` rather than `adapt:start`.

brian-learningpool:
The init() function now listens out for adapt:start rather than app:dataReady.

### DIFF
--- a/src/core/js/drawer.js
+++ b/src/core/js/drawer.js
@@ -27,7 +27,7 @@ define([
         });
     };
 
-    Adapt.once('app:dataReady', function() {
+    Adapt.once('adapt:start', function() {
         init();
     });
 


brian-learningpool:
This is also a bug in the 2.2.1 legacy branch which I need patched.

Issue #1795, opened 2017-09-13
Framework: Ability to turn accessibility state on / off


guywillis:
At present, the accessibility state for each component is always rendered and read out. 

This means that, if you had an optional graphic component with no defined alt text, the accessibility state would read the component as complete even though the user has not previously tabbed through an image.

Suggestion:
Add config to turn on / off the accessibility completed state on a component by component basis. 

```
{{#if _disableAccessibilityState}}
{{else}}
<div class="accessibility-state">
	{{#if _isComplete}}
	<span class="aria-label a11y-ignore-focus prevent-default" id="buttons-aria-label-complete" role="region" tabindex="0">{{a11y_normalize title}} {{_globals._accessibility._ariaLabels.complete}} {{#if _isQuestionType}}{{#if _canShowFeedback}}{{#if _isCorrect}}{{_globals._accessibility._ariaLabels.correct}}{{else}}{{_globals._accessibility._ariaLabels.incorrect}}{{/if}}{{/if}}{{/if}}</span>
	{{else}}
	<span class="aria-label a11y-ignore-focus prevent-default" id="buttons-aria-label-incomplete" role="region" tabindex="0">{{a11y_normalize title}} {{_globals._accessibility._ariaLabels.incomplete}}</span>
	{{/if}}
</div>
{{/if}}
```

moloko:
I thought you could already do this by adding a class of `no-state`?

oliverfoster:
it can, but guy is right, it would be better if it fit into the normal adapt abstractions

moloko:
done in 2b434143bc94b0c46d8f26ddcabd74cb7f21dddb and documented [here](https://github.com/adaptlearning/adapt_framework/wiki/Core-model-attributes#component-model-attributes)

PR #1834, opened 2017-10-05
Add attribute for _disableAccessibilityState

chris-steele:
issue https://github.com/adaptlearning/adapt_framework/issues/1795

Update rather than recreate component state and added `_disableAccessbilityState` option for components


src/core/js/views/componentView.js
```diff
@@ -25,24 +25,18 @@ define([
...
+            var $rendered = $(Handlebars.partials['state']( _.extend(this.model.toJSON(), {a11yConfig:Adapt.config.get('_accessibility')}) ));
 
-            //restore previous tab index if not on
-            var previousTabIndex = $previousState.find(".aria-label").attr("tabindex");
-            if (previousTabIndex == "-1") {
-                $rendered.find(".aria-label").attr("tabindex", previousTabIndex);
-            }
-
-            this.$el.append( $rendered );
+            $previousState.length ? $previousState.html( $rendered.html() ) : this.$el.append($rendered);
```
oliverfoster: can you switch to an if statement?

```js
var isFirstRun = !($previousState.length);
if (isFirstRun) {
  this.$el.append(rendered);
} else {
  $previousState.html(rendered);
}
```

src/core/js/views/componentView.js
```diff
@@ -25,24 +25,18 @@ define([
...
 			// do not perform if component has .no-state class
             if (this.$el.is(".no-state")) return;
 
-			//remove pre-exisiting states
-            var $previousState = this.$(".accessibility-state").remove();
+            var $previousState = this.$(".accessibility-state");
 
-            //render and append state partial
-            var $rendered = $(Handlebars.partials['state']( this.model.toJSON() ));
+            var $rendered = $(Handlebars.partials['state']( _.extend(this.model.toJSON(), {a11yConfig:Adapt.config.get('_accessibility')}) ));
```
oliverfoster: multiple lines would be good
```js
var template = Handlebars.partials['state'];
var data =  _.extend(this.model.toJSON(), {
    a11yConfig: Adapt.config.get('_accessibility')
};
var rendered = template(data);
```
i don't think we need ``rendered`` as a jquery object and dom nodes, as in both cases we can use the raw html

oliverfoster:
we're going to need a schema amend for the components on this one. @moloko could you take a look?

src/core/js/views/componentView.js
chris-steele: Done

src/core/js/views/componentView.js
chris-steele: Done

moloko:
ah sorry, I should have read the original issue which contains an example ;-)

How about "Disables the hidden label that describes the state of the component to screenreader users. Useful if the component is of no interest to a screenreader user - such as an optional decorative graphic component."

moloko:
And just to flag that the schema change will need to be done on the authoring tool repo as well...

Issue #1833, opened 2017-10-05
Add aria-level to all headings


oliverfoster:
http://www.maxability.co.in/2015/04/aria-level-property/

* add a global config for default component, block, article, page and menu levels
* allow overrides on each instance from the json (including component items, such as accordion item titles, etc)
* use helpers where appropriate for defining headers ``a11y_header`` or such might be useful

chris-steele:
Done in https://github.com/adaptlearning/adapt_framework/pull/1836

PR #1838, opened 2017-10-09
Use helper for aria-level attributes

chris-steele:
#1833 

Chris' changes:
* added a global attribute ``_ariaLevel`` to resolve in all Handlebars contexts from which the ``a11y_attrs_heading`` helper is called
* added ``a11y_attrs_heading`` attribute types ``menu``, ``menuItem``, ``page``, ``article``, ``block``, ``component`` and ``componentItem``
* added default heading type levels
* switched headers to helper ```{{{a11y_attrs_heading}}}```
* added heading types to handlebars files in relevant places

Ollie's changes:
* Add safeString wrapper for a11y helpers so that they work with triple and double curly brace wrappers ``{{{ }}}`` and ``{{ }}``

Tickets to do afterwards:
* Add _ariaLevel to schemas #1941 

Usage:
```hbs
<div role="heading" tabindex="0" class="{{_component}}-title-inner component-title-inner"  aria-level="4">
    {{{compile displayTitle this}}}
</div>
```
```hbs
<div class="{{_component}}-title-inner component-title-inner" {{{a11y_attrs_heading "component"}}}>
    {{{compile displayTitle this}}}
</div>
```


src/core/templates/article.hbs
```diff
@@ -5,7 +5,7 @@
     <div class="article-header">
         {{#if displayTitle}}
         <div class="article-title">
-            <div role="heading" tabindex="0" class="article-title-inner h2"  aria-level="2">
+            <div class="article-title-inner h2" {{{a11y_attrs_heading}}}>
```
oliverfoster: you can remove the "hX" classes now i think. they're not used anywhere.
can we do:
```hbs
{{{a11y_attrs_heading 'article'}}}
{{{a11y_attrs_heading 1}}}
```
instead of the ``instanceof`` checks? the structure is likely to change

src/course/config.json
```diff
@@ -12,7 +12,13 @@
         "_isEnabledOnTouchDevices": false,
         "_shouldSupportLegacyBrowsers": true,
         "_isTextProcessorEnabled": true,
-        "_isSkipNavigationEnabled": true
+        "_isSkipNavigationEnabled": true,
+        "_ariaLevels": {
```
oliverfoster: we need a way of overriding this for each article, block, component etc. either default / specified on the model. this is very much dependent on the course content.

src/core/templates/article.hbs
guywillis: To my knowledge, nothing in the vanilla theme now references the 'hX' classes as the hierarchy is governed by the ```aria-level='X'```

src/core/js/accessibility.js
```diff
@@ -129,7 +129,25 @@ define([
             });
 
             Handlebars.registerHelper('a11y_attrs_heading', function(level) {
```
oliverfoster:
```js
Handlebars.registerHelper('a11y_attrs_heading', function(level, defaultName) {
```
something like ^ that might be a better strategy. it'd be more clear where the values are coming from

```js
                var cfg = Adapt.config.get('_accessibility');

                // if a level has not been passed as a parameter
                if (!_.isString(level)) {
                    level = 0;
                    // look in the global configuration
                    if (cfg._ariaLevels && cfg._ariaLevels['_'+defaultName] !== undefined) {
                        level = cfg._ariaLevels['_'+defaultName]
                    }
                }
                return level ? ' role="heading" aria-level="'+level+'" '+getTabIndex()+' ' : getTabIndex()+' ';
```

chris-steele:
@oliverfoster Having look at your comments I interpreted that we:
1. are able to specify a level integer or level type as a parameter
2. an integer is used as-is, a type is looked up in the global config
3. a level override can be given by the context
4. the default level is 1 if nothing above overrides it

src/core/js/accessibility.js
```diff
@@ -117,25 +117,53 @@ define([
             });
 
             Handlebars.registerHelper('a11y_aria_label', function(text) {
-                return '<div class="aria-label prevent-default" tabindex="0" role="region">'+text+'</div>';
+                return '<div class="aria-label prevent-default'+getIgnoreClass()+'" '+getTabIndex()+' role="region">'+text+'</div>';
```
chetan-hajare: A space is required after prevent-default

src/core/js/accessibility.js
```diff
@@ -117,25 +117,53 @@ define([
...
             });
 
             Handlebars.registerHelper('a11y_aria_label_relative', function(text) {
-                return '<div class="aria-label relative prevent-default" tabindex="0" role="region">'+text+'</div>';
+                return '<div class="aria-label relative prevent-default'+getIgnoreClass()+'" '+getTabIndex()+' role="region">'+text+'</div>';
```
chetan-hajare: A space is required after prevent-default.

src/course/config.json
```diff
@@ -12,7 +12,15 @@
...
```
moloko: this will need adding to config.model.schema in both framework and in the authoring tool - and will need documenting somewhere too if not done already

oliverfoster: I've added a ticket for that work #1941

oliverfoster:
I need to add a componentItem default. Always miss that one.

oliverfoster:
could i get a +1 for the minor amends, just to check i've not messed up anything?

moloko:
ack sorry to be a pain but I've just realised - this doesn't adhere to the rule we try to follow that configuration properties should be prefixed with an underscore...

oliverfoster:
how's that? @moloko

Issue #1840, opened 2017-10-12
Allow template that notify uses to be specified


chris-steele:
issue https://github.com/adaptlearning/adapt_framework/issues/1525

This will allow notify to be more flexible in what it can display and perhaps reduce the reliance on custom overlay/lightbox plugins.

oliverfoster:
Could you turn this into accepting a view rather than a template?

Allowing a template override only means that the view behaviour associated with the template must be external to the notify and must seek the DOM node after render of the notify and before use interaction. If the view is instantiated and handed off to the notify it makes for a much cleaner API.
I would prefer if this behaved more like the drawer; in that it should accept child views. I would see this as a leapfrog step, scrapping the transition to bespoke templates and allowing bespoke, atomized behaviour, templates and styling.

oliverfoster:
@chris-steele is there an open pr for this?

PR #1841, opened 2017-10-12
Permit custom views in notify popups

chris-steele:
issue https://github.com/adaptlearning/adapt_framework/issues/1840

Useful and particularly so to https://github.com/adaptlearning/adapt-contrib-hotgraphic/pull/144

### DIFF
--- a/src/core/js/views/notifyView.js
+++ b/src/core/js/views/notifyView.js
@@ -66,10 +66,11 @@ define([
         },
 
         render: function() {
-            var data, html;
+            var data;
+            var html;
 
             if (this.customView) {
-                html = this.customView.$el;
+                html = this.customView.el;
             } else {
                 data = this.model.toJSON();
                 html = Handlebars.templates['notify'](data);
--- a/src/core/js/views/notifyView.js
+++ b/src/core/js/views/notifyView.js
@@ -7,16 +7,20 @@ define([
         className: 'notify',
         disableAnimation: false,
         escapeKeyAttached: false,
-        template:null,
+        customView:null,
 
-        initialize: function(attrs, options) {
+        initialize: function(options) {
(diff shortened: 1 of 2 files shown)


oliverfoster:
Could you turn this into accepting a view rather than a template?

Allowing a template-override-only means that the view behaviour associated with the template must be external to the notify and must seek the DOM node after render of the notify. If the view is instantiated and handed off to the notify it makes for a much cleaner API.
I would prefer if this behaved more like the drawer; in that it should accept child views. I would see this as a leapfrog step, scrapping the transition to bespoke templates and allowing bespoke, atomized behaviour, templates and styling.

chris-steele:
@oliverfoster Notify now accepts a custom view. Let me know if you want to see a working demo with the latest hotgraphic.

src/core/js/views/notifyView.js
```diff
@@ -60,13 +66,19 @@ define([
         },
 
         render: function() {
-            var data = this.model.toJSON();
-            var template = Handlebars.templates['notify'];
+            var data, html;
```
lc-thomasberger: Out Style-Guide recommends to [declare one variable per var statement](https://github.com/adaptlearning/documentation/blob/master/01_cross_workstream/style_guide.md#declare-one-variable-per-var-statement).

src/core/js/views/notifyView.js
```diff
@@ -60,13 +66,19 @@ define([
...
+
+            if (this.customView) {
+                html = this.customView.$el;
```
lc-thomasberger: `this.customView.el` rather then `this.customView.$el` as it is passed into `jQuery.html()` ?

oliverfoster:
Can we close this and use subviews instead of replacement views?
https://github.com/adaptlearning/adapt_framework/pull/1893

PR #1846, opened 2017-10-17
fixes a bug with  requireCompletionOf that was accidentally introduced in #1607

moloko:
also refactored the code out into a separate function to remove duplication

### DIFF
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -131,7 +131,6 @@ define([
          * @param {string} [completionAttribute] Either "_isComplete" or "_isInteractionComplete". Defaults to "_isComplete" if not supplied.
          */        
         checkCompletionStatusFor: function(completionAttribute) {
-            // in the absence of support for default function parameters in IE & Safari v9...
             if (!completionAttribute) completionAttribute = "_isComplete";
 
             var completed = false;
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -138,9 +138,9 @@ define([
             var requireCompletionOf = this.get("_requireCompletionOf");
 
             if (requireCompletionOf === -1) { // a value of -1 indicates that ALL mandatory children must be completed
-                completed = !(_.find(children, function(child) {
+                completed = (_.find(children, function(child) {
                     return !child.get(completionAttribute) && !child.get('_isOptional');
-                }));
+                }) === undefined);
             } else {
                 completed = (_.filter(children, function(child) {
                     return child.get(completionAttribute) && !child.get('_isOptional');
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -124,11 +124,11 @@ define([
         },
 
         /**
-         * Function for checking whether the supplied completion attribute should be set to true or false
+         * Function for checking whether the supplied completion attribute should be set to true or false. 
          * It iterates over our immediate children, checking the same completion attribute on any mandatory child
          * to see if enough/all of them them have been completed. If enough/all have, we set our attribute to true; 
-         * if not, we set it to false
-         * @param {String} completionAttribute Either "_isComplete" or "_isInteractionComplete"
+         * if not, we set it to false.
+         * @param {string} [completionAttribute] Either "_isComplete" or "_isInteractionComplete". Defaults to "_isComplete" if not supplied.
          */        
         checkCompletionStatusFor: function(completionAttribute) {
             // in the absence of support for default function parameters in IE & Safari v9...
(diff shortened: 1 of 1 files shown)


moloko:
To replicate the issue, add `_requireCompletionOf: 2` to article a-05 (using the OOTB content). Launch the course and complete the text and graphic components in b-05. Check `_isComplete` for co-05, it should be set to `false` as you have only completed one block but in fact gets set to `true` as it was comparing the number of incomplete components to the value of `_requireCompletionOf`. 

This fix reverses the logic so that it compares the number of completed components, as it did prior to #1607

src/core/js/models/adaptModel.js
```diff
@@ -112,57 +112,45 @@ define([
...
+         * It iterates over our immediate children, checking the same completion attribute on any mandatory child
+         * to see if enough/all of them them have been completed. If enough/all have, we set our attribute to true; 
+         * if not, we set it to false.
+         * @param {string} [completionAttribute] Either "_isComplete" or "_isInteractionComplete". Defaults to "_isComplete" if not supplied.
+         */        
+        checkCompletionStatusFor: function(completionAttribute) {
+            // in the absence of support for default function parameters in IE & Safari v9...
+            if (completionAttribute === void 0) completionAttribute = "_isComplete";
+
+            var complete = false;
```
brian-learningpool: (Minor) It might be slightly more readable to call this variable `isComplete`.

src/core/js/models/adaptModel.js
```diff
@@ -112,57 +112,45 @@ define([
...
+        /**
+         * Function for checking whether the supplied completion attribute should be set to true or false. 
+         * It iterates over our immediate children, checking the same completion attribute on any mandatory child
+         * to see if enough/all of them them have been completed. If enough/all have, we set our attribute to true; 
+         * if not, we set it to false.
+         * @param {string} [completionAttribute] Either "_isComplete" or "_isInteractionComplete". Defaults to "_isComplete" if not supplied.
+         */        
+        checkCompletionStatusFor: function(completionAttribute) {
+            // in the absence of support for default function parameters in IE & Safari v9...
+            if (completionAttribute === void 0) completionAttribute = "_isComplete";
```
brian-learningpool: (Minor) Won't `if (!completionAttribute)` work?  There is also `_.isUndefined()`.  

(I know `void 0` will work, just don't think I've seen it used anywhere else in the codebase.)

moloko: Hah I knew someone would comment on this, I totally just pinched this code from the code the TypeScript compiler generates when you do default function parameters! I'm quite happy to change it if it's too outré!

src/core/js/models/adaptModel.js
moloko: The only reason I changed this is that one of the attributes it's checked is `_isComplete` so I felt it was a bit too close to that...

src/core/js/models/adaptModel.js
oliverfoster: It is obtuse.

src/core/js/models/adaptModel.js
moloko: I've gone for `completed` instead which seems slightly clearer?

moloko:
the only thing I wouldn't mind people checking on is that the original `checkInteractionCompletionStatus` had this code:
```
isInteractionComplete = (_.find(children, function(child) {
  return child.get('_isInteractionComplete') === false && child.get('_isOptional') === false;
}) === undefined)
```
(specifically the `=== undefined` bit)
The `checkCompletionStatus` function also used to be written like that but #1607 changed it to:
```
isComplete = !(_.find(children, function(child) {
  return !child.get('_isComplete') && !child.get('_isOptional');
}));
``
I have assumed that it was just an oversight that `checkInteractionCompletionStatus` was not changed in the same way - but if there was a reason for leaving it as-is let me know!

ryanrolds:
That was an oversight on my part. Looking at the Lodash docs for find and findWhere, `result === undefined` is a probably a better check then `!result`, however when working with objects it shouldn't be a problem. It would a problem if `result` is `0`.

moloko:
@ryanrolds thanks for taking the time to check and feed back; I have changed to checking for `undefined`.

moloko:
@chris-steele would you mind reviewing again?

oliverfoster:
the falsey was fine. as ryan says, we're iterating through an array of objects only, using underscore to help. if find returns something it'll always be a truthy.

ryanrolds:
`_.some(...)` would be cleaner, but it's probably not worth the time.

moloko:
> Could always use _.some?

@chris-steele oooh I wish I'd thought of that while I was refactoring but agree with @ryanrolds probably not worth changing now!

> the falsey was fine. as ryan says, we're iterating through an array of objects only, using underscore to help. if find returns something it'll always be a truthy.

@oliverfoster would you like me to change it back or are you happy for it to be left as-is?

oliverfoster:
nah, leave it. no point spending any more time here.

PR #1858, opened 2017-11-08
Selector rather than a classname

NayanKhedkar:


src/core/js/startController.js
```diff
@@ -70,7 +70,7 @@ define([
                         if (model.get("_isComplete")) continue;
                     }
 
-                    if (!className || $("html").is(className)) {
+                    if (!className || $("html").is('.' + className)) {
```
tomgreenfield: I know the previous implementation was incorrect but this could be made backwards compatible if we did something along the lines of:
```js
var $html = $("html");

if (!className || $html.is(className) || $html.hasClass(className)) {

PR #1866, opened 2017-11-14
issue/1828: corrected plugin dependencies in adapt.json

chris-steele:


chris-steele:
@tomgreenfield Yes you are right. As I did this I read them as "greater than or equal to", but for some reason used a caret :/ Will amend!

PR #1869, opened 2017-11-14
Resolves #1868, reverts Grunt copy of core 'libraries'

### LINKED ISSUE #1868: Core libraries folder not copied in AT
An issue has been introduced by aa1cc4e18c710ced750f201b94d29c12058e6f40 which prevents the core libraries folders from copying over.  This can be observed in the authoring tool, where the `includes` plugins are set.

The core libraries should not be passed through the plugin filter.

brian-learningpool:
An issue was introduced in aa1cc4e which passed the core 'libraries' folder through the plugin filter, which would prevent the core libraries copying to the build folder for AT users.

### DIFF
--- a/grunt/config/copy.js
+++ b/grunt/config/copy.js
@@ -195,11 +195,23 @@ module.exports = function (grunt, options) {
                 }
             ]
         },
+        libraries: {
+            files: [
+                {
+                    expand: true,
+                    src: [
+                        '<%= sourcedir %>core/js/libraries/**/*'
+                    ],
+                    dest: '<%= outputdir %>libraries/',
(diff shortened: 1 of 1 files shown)


oliverfoster:
screenshot from 2017-11-14 11-46-13

oliverfoster:
```json
        libraries: {
            files: [
                {
                    expand: true,
                    src: [
                        '<%= sourcedir %>core/js/libraries/**/*'
                    ],
                    dest: '<%= outputdir %>libraries/',
                    rename: _.partial(collate, "libraries")
                }
            ]
        },
```
^ need to rename the files on the way through

Issue #1884, opened 2017-11-28
jquery.resize overwrites $(window).resize() unnecessarily


oliverfoster:
https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/libraries/jquery.resize.js#L235-L249

these lines prevent $(window).resize();
they are also completely unnecessary as this is what jquery does by default, plus some other bits.

comes from bug:
https://github.com/adaptlearning/jquery.resize/issues/10

PR #1885, opened 2017-11-28
issue/1884: updated jquery.resize to latest bugfix

oliverfoster:
#1884 

To test. 
Before replacing this code should not console log:
```js
$(window).on("resize", function() {
  console.log("window resize triggered");
});
$(window).trigger('resize'); // should console log here
$(window).resize(); // won't console log here
```
After replacing it should console log:
```js
$(window).on("resize", function() {
  console.log("window resize triggered");
});
$(window).trigger('resize'); // should console log here
$(window).resize(); // should console log here
```


lc-thomasberger:
see also: https://github.com/adaptlearning/adapt-contrib-trickle/pull/86

Issue #1906, opened 2017-12-20
core: allow templates to access all course data


oliverfoster:
* add ``{{import_adapt}}`` helper
  add this once to the top of the template or the beginning of the json string
* allow ``{{Adapt.config.propertyName}}``
* allow ``{{Adapt.course.propertyName}}``
* allow ``{{Adapt.contentObjects.[id].propertyName}}``
* allow ``{{Adapt.articles.[id].propertyName}}``
* allow ``{{Adapt.blocks.[id].propertyName}}``
* allow ``{{Adapt.components.[id].propertyName}}``

#### Example
```handlebars
{{import_adapt}} {{{Adapt.course.displayTitle}}} {{{Adapt.components.[c-05].displayText}}}
```

PR #1907, opened 2017-12-20
issue/1906: allow templates to access all course data

oliverfoster:
#1906

### DIFF
--- a/src/core/js/helpers.js
+++ b/src/core/js/helpers.js
@@ -115,20 +115,26 @@ define([
         import_data: function(context) {
             
             var directImport = ['config','course'];
-            directImport.forEach(function(name) {
+            for (var i = 0, l = directImport.length; i < l; i++) {
+                var name = directImport[i];
                 context.data.root['_'+name] = Adapt[name].toJSON();
-            });
+            }
 
             var indexedImport = ['contentObjects','articles','blocks','components'];
-            indexedImport.forEach(function(name) {
-                context.data.root['_'+name] = _.indexBy(Adapt[name].toJSON(), function(item) {
-                    return item._id;
-                });
-            });
+            for (var i = 0, l = indexedImport.length; i < l; i++) {
+                var name = indexedImport[i];
+                var importArray = Adapt[name].toJSON();
+                var importIndex = {};
+                for (var i1 = 0, l1 = importArray.length; i1 < l1; i1++) {
+                    var item = importArray[i1];
+                    importIndex[item._id] = item;
+                }
+                context.data.root['_'+name] = importIndex;
+            }
 
             return "";
         }
-        
+
     };
 
     // Compatibility references
--- a/src/core/js/helpers.js
+++ b/src/core/js/helpers.js
@@ -114,7 +114,7 @@ define([
          */
         import_data: function(context) {
             
-            var directImport = ['config', 'course'];
+            var directImport = ['config','course'];
             directImport.forEach(function(name) {
                 context.data.root['_'+name] = Adapt[name].toJSON();
             });
@@ -127,12 +127,8 @@ define([
             });
 
             return "";
-        },
-
-        helperMissing: function() {
-            debugger;
         }
-
+        
     };
 
     // Compatibility references
--- a/src/core/js/helpers.js
+++ b/src/core/js/helpers.js
@@ -101,12 +101,50 @@ define([
 
         /**
          * makes the _globals object in course.json available to a template
-         */ 
+         */
         import_globals: function(context) {
-            if(!context.data.root._globals) {
+            if (!context.data.root._globals) {
                 context.data.root._globals = Adapt.course.get('_globals');
             }
             return "";
+        },
+
+        /**
+         * makes the course data available to a template
+         */
+        import_data: function(context) {
+            if (!context.data.root._config) {
+                context.data.root._config = Adapt.config.toJSON();
+            }
+            if (!context.data.root._course) {
+                context.data.root._course = Adapt.course.toJSON();
+            }
+            if (!context.data.root._contentObjects) {
+                context.data.root._contentObjects = _.indexBy(Adapt.contentObjects.toJSON(), function(item) {
+                    return item._id;
+                });
+            }
+            if (!context.data.root._articles) {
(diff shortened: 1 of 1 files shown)


src/core/js/helpers.js
```diff
@@ -101,11 +101,37 @@ define([
...
+        import_data: function(context) {
+            
+            var directImport = ['config','course'];
+            for (var i = 0, l = directImport.length; i < l; i++) {
+                var name = directImport[i];
+                context.data.root['_'+name] = Adapt[name].toJSON();
+            }
+
+            var indexedImport = ['contentObjects','articles','blocks','components'];
+            for (var i = 0, l = indexedImport.length; i < l; i++) {
```
moloko: I think you could switch to `for...in` for both these `for` loops? or `_.each`?

moloko: although if you want this to go into Adapt v2.x you'll need to include a `hasOwnProperty` check with the `for...in` to stop it from iterating over the methods that the ES5 Shim adds to Array... might be an idea to include that regardless.

oliverfoster: I've replaced ```forEach``` calls for ```for``` loops after Tom G cited performance concerns as this could be used in every ```component.hbs``` template and would need to remap the course data every time.

oliverfoster: ```for...in``` is an odd type of loop to use on an Array. It's an enumerable key loop, which I guess would work fine with arrays but isn't ... really... proper...?

moloko: I don't see why it's not 'proper', in fact the MDN docs refer specifically to using it on an Array! So long as you don't mind that it won't iterate through the Array in a specific order and you do a `hasOwnProperty` check it works great, nice tidy syntax - esp. when you've got an inner loop too.

src/core/js/helpers.js
```diff
@@ -101,11 +101,38 @@ define([
...
+            var directImport = ['config','course'];
+            for (var i = 0, l = directImport.length; i < l; i++) {
+                var name = directImport[i];
+                if (context.data.root['_'+name]) continue;
+                context.data.root['_'+name] = Adapt[name].toJSON();
+            }
+
+            var indexedImport = ['contentObjects','articles','blocks','components'];
+            for (var i = 0, l = indexedImport.length; i < l; i++) {
+                var name = indexedImport[i];
```
moloko: variable redeclaration, strictly speaking (as you've already declared it on L118 and `var` variables aren't block-scoped)

src/core/js/helpers.js
moloko: mmmm so tidy 
```javascript
var directImport = ['config','course'];
for (let name in directImport) {
    if(directImport.hasOwnProperty(name)) {
        context.data.root['_'+name] = Adapt[name].toJSON();
    }
}

var indexedImport = ['contentObjects','articles','blocks','components'];
for (let name in indexedImport) {
    if (!indexedImport.hasOwnProperty(name)) continue;
    var importArray = Adapt[name].toJSON();
    var importIndex = {};
    for (let item in importArray) {
        if(importArray.hasOwnProperty(item)) {
            importIndex[item._id] = item;
        }
    }
    context.data.root['_' + name] = importIndex;
}
```

oliverfoster: I mean that the intention of a ``for...in`` is to iterate through an Object's enumerable properties and not an Array's indices. Even though those two ideas have overlap and are technically similar in that they can both be enumerable properties, if I see a ``for...in`` my assumption is that the code is acting upon and Object's name value pairs and not an Array's indices. To me it would feel as though the code obscures the intention of the loop for the sake of neater syntax. 
As the ``for...in`` doesn't restrict itself to only the numeric indices from the Object's property names, if the Array was extended or had an extra enumerable property for some reason, the ``for...in`` would create unnecessary addition logic problems to solve without an easy or clear solution, even with a ``hasOwnProperty`` check. 
Adding the ```hasOwnProperty``` check means that I have to include an extra function call for each indices. Which kind of negates the point of switching from a ``forEach``.
From my perspective using a ``for...in`` makes the code unnecessary fragile, less performant and more obtuse for very little benefit.

moloko: Although you are looping over the array indices in your code, in this instance it's only for the purposes of getting hold of the array values, which a `for...in` loop does for you

Unnecessarily fragile? Not with the `hasOwnProperty` check. In what circumstances would that not work?

More obtuse? Very subjective, that. Reads a LOT clearer to me. We'll have to agree to disagree on that point I fear.

Less performant? Yes, possibly. Will the impact be more than negligible though? 

"The real problem is that programmers have spent far too much time worrying about efficiency in the wrong places and at the wrong times; premature optimization is the root of all evil (or at least most of it) in programming." - Donald Knuth

Anyway it's all moot as these are just suggestions, happy for it to go in as-is

oliverfoster:
```js
var directImport = ['config','course'];
directImport.prop = "a";
for (let name in directImport) {
    console.log(name);
}
```

moloko: > I don't understand why you're intentionally ignoring this point

Because I don't agree that it's an established pattern of use

> Moving from the `forEach` is not premature optimization

I wasn't referring to that.

I'm sure you'll see I have +1'd this so I think we need to agree to disagree and move on!

oliverfoster: Fair enough.

tomgreenfield:
To avoid ambiguity with existing properties, suggest something like ```{{{Adapt.course.displayTitle}}}``` rather than ```{{{_course.displayTitle}}}```?

src/core/js/helpers.js
```diff
@@ -101,12 +101,47 @@ define([
...
+            return "";
+        },
+
+        /**
+         * makes the Adapt module data available to a template
+         */
+        import_adapt: function(context) {
+
+            if (context.data.root.Adapt) return;
+            var adapt = context.data.root.Adapt = {};
```
tomgreenfield: P.S. this one-liner took me a while to read.

oliverfoster:
@moloko can you re +1?

src/core/js/helpers.js
moloko: wow you're a slow reader ;-)

Issue #1888, opened 2017-12-05
add a 'completed' class to page/article/block/component


moloko:
probably most useful at the component level for styling completed components differently - but might as well add this throughout

NayanKhedkar:
PR: https://github.com/adaptlearning/adapt_framework/pull/1911

PR #1911, opened 2018-01-08
Issue/1888

NayanKhedkar:
#1888 

added a 'completed' class to page/article/block/component

### DIFF
--- a/src/core/js/views/adaptView.js
+++ b/src/core/js/views/adaptView.js
@@ -174,8 +174,7 @@ define([
             this.$el.addClass('display-none');
         },
         
-        onIsCompleteChange:function(model){
-            var isComplete=model.get('_isComplete');
+        onIsCompleteChange:function(model, isComplete){
             this.$el.toggleClass('completed', isComplete);
         }
     });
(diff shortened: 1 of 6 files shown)


src/core/js/views/adaptView.js
```diff
@@ -171,6 +172,10 @@ define([
                 return this.$el.removeClass('display-none');
             }
             this.$el.addClass('display-none');
+        },
+        
+        addCompletionClass:function(){
+            this.$el.addClass('completed');
```
oliverfoster: needs to take account of the value change. it may be changing from true to false.
```js
onChangeComplete: function(value) {
    this.$el[value?'addClass':'removeClass']('completed');
}
```
```js
onChangeComplete: function(value) {
    if (!value) {
        this.$el.removeClass('completed');
    } else {
        this.$el.addClass('completed');
    }
}
```

tomgreenfield:
```js
onIsCompleteChange: function(isComplete) {
    this.$el.toggleClass('completed', isComplete);
}
```

oliverfoster: Nice that toggle takes the second argument. Will have to remember that.

src/core/js/views/adaptView.js
```diff
@@ -171,6 +172,11 @@ define([
                 return this.$el.removeClass('display-none');
             }
             this.$el.addClass('display-none');
+        },
+        
+        onIsCompleteChange:function(model){
+            var isComplete=model.get('_isComplete');
```
tomgreenfield: Can you use the second argument passed to the function rather than a new variable please?

PR #1928, opened 2018-01-22
issue/1603: removed jquery.a11y ie8 hack

oliverfoster:
#1603 

to test:
* check that Jaws can still read text in IE11, test with vanilla project should be fine.

### DIFF
--- a/src/core/js/libraries/jquery.a11y.js
+++ b/src/core/js/libraries/jquery.a11y.js
@@ -227,8 +227,7 @@
                     var cloneChild = $(child.outerHTML)[0];
                     switch(child.nodeType) {
                     case 3: //TEXT NODE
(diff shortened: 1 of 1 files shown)


src/core/js/libraries/jquery.a11y.js
```diff
@@ -227,8 +227,7 @@
                     var cloneChild = $(child.outerHTML)[0];
                     switch(child.nodeType) {
                     case 3: //TEXT NODE
-                        // preserve whitespace in ie8 by adding initial zero-width space
-                        var childContent = child.textContent || "&#8203;" + child.nodeValue;
+                        var childContent = child.textContent;
```
tomgreenfield: Can you remove the variable please?

Issue #1874, opened 2017-11-21
Changing Dest of menuAssets in Grunt task


ciaranlp:
Currently in `grunt/config/copy.js` `menuAssets` **dest** is set to `dest: '<%= outputdir %>assets/'` this should be brough into line with `menuFonts`, `themeFonts` and `themeAssets` that all point to `Adapt/css/[respective-folder]`for example `themeAssets` has `dest: '<%= outputdir %>adapt/css/assets/'`

oliverfoster:
i think all of the asset folders should go to adapt/css/assets not just menu and theme.
it seems silly to have two asset folder locations as we do at the moment. one at  build/assets and one at build/adapt/css/assets.

there isn't a proposal where we could change the location of this folder without breaking any existing plugins. except duplicating the assets over both folders.

oliverfoster:
I had a little more of a think around this...
Assuming that:
* we want to unify the assets folders
* stop code from breaking
* warn people to update their code

We could probably:
* get grunt to output all assets to adapt/css/assets (or move the adapt.css and invert that)
* add a less preprocessor to warn of occurrences of ```../../assets/``` in the LESS source files by path
* after the less is compiled into css, add a css postprocessor to correct the occurances of ```../../assets/``` to ```assets/```

oliverfoster:
any thoughts on this @ciaranlp ?

ciaranlp:
I like the idea of having all assets in one place and the css file moved to root

oliverfoster:
cool. we'd have to move the font and assets folders to root, but that's quite doable.
it would be easier in the first instance to unify them into adapt/css/assets, as fewer things have to change, but i'll happily unify them at /assets instead if we can all agree

any objections? @moloko @brian-learningpool @lc-thomasberger @tomgreenfield @taylortom @dancgray @danielstorey

oliverfoster:
### Changes required
When moving root/adapt/css/adapt.min.css to root/adapt.min.css
* All index.html s need changing as this file is referenced directly
* Change grunt to output all fonts to root/fonts/ instead of  root/adapt/css/fonts
* Change grunt to output all assets to root/assets
* Add less preprocessor to change "../../assets/" references to "assets/" and warn about deprecation

oliverfoster:
@ciaranlp can you have a look at the above pr? see if that helps.
don't install spoor to test, i haven't done a pr for the link tag change in spoor yet.

ciaranlp:
[...]

PR #1932, opened 2018-01-22
issue/1874: amalgamate assets folders

oliverfoster:
#1874 

#### moved
  adapt/css/fonts/ > fonts/
  adapt/css/assets/ > assets/
  adapt/css/adapt.min.css > adapt.min.css

#### added
  preprocessor for less, to change ``url(../../assets`` to ``url(assets/``
  outputs a warning for the user with files that need updating
image

#### to test
* clone adapt_framework
* install plugins
* build course
* check that all assets, css and fonts are displaying correctly


lc-thomasberger:
@oliverfoster adapt.min.css is not created. I get the following warning in the console when running grunt build:
Running "less:compile" (less) task
Warning: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode Use --force to continue.

I am on node 4.4.4 and npm 5.3.0

oliverfoster:
Ahaa, I thought we'd gone up to node 6. I can make it backwards compatible... give me a few minutes.

moloko:
I don't think we need to be supporting Node v4 any more

oliverfoster:
I've updated it @lc-thomasberger can you see if it works in node 4/6 again?

lc-thomasberger:
The authoring does support it, so I think we should at least try to support is a good as possible.
Thanks @oliverfoster

grunt/tasks/less/replaceUrls.js
```diff
@@ -0,0 +1,77 @@
...
+                return URLNode;
+            }
+            if (URLNode.value && URLNode.value.value && URLNode.value.value.indexOf('#') === 0) {
+              // Might be part of a VML url-node value like:
+              // ``behavior:url(#default#VML);``
+              return URLNode;
+            }
+
+            this._options.replaceUrls.forEach(function(replaceObject) {
+                var matches = URLNode.value.value.match(replaceObject.find);
```
brian-learningpool: I think this should have the same defensive programming as line 37.  Otherwise we'll possibly end up with `"Cannot read property 'match' of undefined"`.

oliverfoster: Can you give me a console log over URLNode and URLNode.value?

oliverfoster:
Fingers crossed on that one.

nilslp:
+1 - Tested

PR #1935, opened 2018-01-23
issue/1303: updated to jquery v3

oliverfoster:
* added jquery.min from latest v3
* updated jquery.mobile to v1.5.0 alpha for jquery v3 support
* deleted jquery.v2.min

#1303


brian-learningpool:
@oliverfoster, is this just the 1.5.0-alpha1 download of JQuery Mobile with "Select All" on every option?

oliverfoster:
No, it's the same as the old one. https://github.com/adaptlearning/adapt_framework/blob/1fdcb28e68b358a602b8a8411b758f1523611a04/src/core/js/libraries/jquery.mobile.custom.js#L3 < as detailed up the top

Issue #1934, opened 2018-01-23
Update modernizr


oliverfoster:
### Subject of the issue/enhancement/features
Modernizr needs updating as the our version is 2.8.3 and the current stable is 3.5.0

### Problems
Modernizr.load API has been retired and so needs replacing in the scriptLoader.js
Modernizr.touch identification API has changed slightly and needs to be made backwards compatible.

### Needed Outcomes
Retrofit Modernizr.touchevents to Modernizr.touch + classes
Replace Modernizr.load with loadScript functionality in scriptLoader.js
Implement Adapt.loadScript API with callback as a replacement to Modernizr.load
Remove Modernizr checks which are for <IE11 browsers
Switch to using no-touchevents and touchevents in core
Switch core to use Adapt.device.touch rather than Modernizr.touch

PR #1936, opened 2018-01-23
issue/1934: updated modernizr

oliverfoster:
#1934 

Works done:
* Have retrofitted ``Modernizr.touchevent`` to ``Modernizr.touch``
* Have added ``.touch`` and ``.no-touch`` for backward compatibility 
* Replaced ``Modernizr.load`` with ``loadScript`` function 
* Created ``Adapt.loadScript`` API
* Removed Modernizr checks for <IE11
* Switched core to use new ``.no-touchevents`` and ``.touchevents``  classes instead of using ``.touch`` and ``.no-touch``
* Switch core to use Adapt.device.touch rather than Modernizr.touch

Breaking changes:
* Dropped ``Modernizr.load`` API use ``Adapt.loadScript`` instead


src/core/js/scriptLoader.js
```diff
@@ -34,24 +45,27 @@
 
     // 3. start loading JQuery, wait for it to be loaded
     function loadJQuery() {
-        Modernizr.load([
-            {
-                load: 'libraries/jquery.v2.min.js',
-                complete: checkJQueryStatus
-            }
-        ]);
+        loadScript('libraries/jquery.v2.min.js', checkJQueryStatus);
```
brian-learningpool: Isn't this altered by the switch to JQuery v3?

oliverfoster: Yup, but they are two separate PRs, so I'll hit that merge conflict when I get there,

src/core/js/scriptLoader.js
```diff
@@ -34,24 +45,27 @@
...
-            loadFoundationLibraries();
+            setupModernizr();
         }
     }
-    
-    //5. Load foundation libraries and templates then load Adapt itself
+
+    //5. Backward compatibility for Modernizr
+    function setupModernizr() {
+        Modernizr.touch = Modernizr.touchevents;
```
brian-learningpool: Is the declaration of the `touch` property for backward compatibility between v2 and v3?  

In this PR you also refer to `Adapt.device.touch`.  Should this be the preferred property for plugins to check after the bootstrapping?

oliverfoster: Yes, it's for backward compatibility only. I don't know if there are any other plugins which reference ```Mozernizr.touch```

brian-learningpool: Thanks.  I can't think of a core plugin but I guess there could be non-core ones using it.

src/core/js/scriptLoader.js
```diff
@@ -1,5 +1,16 @@
 (function() {
 
+    function loadScript(url, callback){
+        if (!url || !(typeof url === 'string')) return;
```
tomgreenfield: Am I missing the reason behind using brackets instead of the below?
```js
typeof url !== 'string'
```

PR #1948, opened 2018-01-30
issue/1727: improved screen reading

oliverfoster:
#1727 

* improved the way ``a11y_text`` outputs to take account of style tags - it now clusters readable sections into one tabbable area
* added ability to create abbreviations for screen readers ``<nb>`` and ``<zw>`` tags
* added a better alternative to ``a11y_normalize``, it doesn't strip all html but keeps styling tags intact : ``a11y_remove_breaks``
* made some readability improvements


moloko:
requires a non-Kineo approval/+1 before can be merged

PR #1952, opened 2018-02-01
add _ariaLevels to schema

### LINKED ISSUE #1941: Add _ariaLevel to schema
After [this pr](https://github.com/adaptlearning/adapt_framework/pull/1838) make sure schema is updated to include ``_ariaLevel`` attribute. Default to 0.

moloko:
fixes #1941 

also remove invalid uses of `isSetting`


moloko:
hold off on reviewing until we've made a decision as to whether the child properties of `_ariaLevel` should be prefixed with an underscore or not

oliverfoster:
There are two properties here. ```_ariaLevels``` as you have it are the default levels on the ```config.json``` and ```_ariaLevel```, which should pair with each title attribute. It should perhaps be called, ```_titleArialevel```.

```_ariaLevel```: is used [here](https://github.com/adaptlearning/adapt_framework/pull/1838/files#diff-767c220811484c7ab090ae86d2400d7dR148)
```_ariaLevels```: is used [here](https://github.com/adaptlearning/adapt_framework/pull/1838/files#diff-767c220811484c7ab090ae86d2400d7dR158)

```_ariaLevel``` on the component, block, article and contentObject should default to ``0`` so that it is considered "unset" and not override the default ```_ariaLevels``` value.

moloko:
@oliverfoster added that in, could you check? 

also documented [here](https://github.com/adaptlearning/adapt_framework/wiki/Core-model-attributes#contentobjects-model-attributes)

oliverfoster:
beautiful

lc-thomasberger:
Since it is a limited number range, a dropdown might be easier to use.

moloko:
> Since it is a limited number range, a dropdown might be easier to use.
In theory, it's unlimited. In practice, it seems unlikely that anyone's going to go much higher than, say, 10 I guess?

I'm happy to change it if people would prefer a drop down and are happy with a limit being placed on it

oliverfoster:
10 seems waaay over reasonable. I'd be happy with 1-10

moloko:
Can anyone point me in the direction of a .schema that uses a "Select" `inputType` to populate a `"number"` type field?

The only ones I can find are all for `"string"`...

oliverfoster:
@tomgreenfield is there a select dropdown + number?

tomgreenfield:
Haven't seen one but would most likely work, @oliverfoster.

My preference would be to keep as a [Number](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number) field for simplicity.

lc-thomasberger:
Number field is fine for me.

oliverfoster:
Can we get some +1's ? I know there is implied approval.

Issue #1876, opened 2017-11-23
contrib-tutor: Title not display title in feedback


canstudios-louisem:
Tutor is showing the components title not display title in the feedback notify.

chetan-hajare:
Hi @canstudios-louisem 
We are using the title in the feedback since the beginning of the development of the Adapt framework. The page level progress extension also uses title to represent component in the drawer. If we think to change the feedback title to display title, we might have to consider the page level progress extension too. @brian-learningpool @moloko @oliverfoster @lc-thomasberger Can you please provide your suggestions on this?

oliverfoster:
i think the feedback should be able to specify a title.

oliverfoster:
need to expand this object to take titles, body text, images etc
https://github.com/adaptlearning/adapt_framework/blob/master/src/course/en/components.json#L424

tomgreenfield:
Related: https://github.com/adaptlearning/adapt_framework/issues/1734.

lc-thomasberger:
For PageLevelProgress it is actually good to use title instead of displayTitle (long titles do not work with Drawer), so I think we should stick with this behavior.
For questions a separate title field would be nice. It should default to the component's title if it is not specified.

canstudios-louisem:
Being able to set the feedback's notify title would fix this issue for us really. Most of the time we just want it to say `Feedback` or similar.

oliverfoster:
@canstudios-louisem  one for _partlyCorrect, corract and incorrect? Or just a general title per component?

chetan-hajare:
I would prefer a general title per component. What do you think about this @canstudios-louisem ?

chetan-hajare:
PR: https://github.com/adaptlearning/adapt_framework/pull/1954

canstudios-louisem:
Thanks for your work on this @chetan-hajare much appreciated :+1:

moloko:
should all be done now that https://github.com/adaptlearning/adapt-contrib-mcq/pull/139 is in.

PR #1954, opened 2018-02-05
Updated feedback to show title independent of title/display title of the component.

chetan-hajare:
#1876

### DIFF
--- a/src/core/js/models/questionModel.js
+++ b/src/core/js/models/questionModel.js
@@ -250,6 +250,8 @@ define([
         setupPartlyCorrectFeedback: function() {
 
             var feedback = this.get("_feedback");
+            
+            if(!feedback) return;
 
             if (feedback && feedback._partlyCorrect) {
                 if (this.get('_attemptsLeft') === 0 || !feedback._partlyCorrect.notFinal) {
--- a/src/core/js/models/questionModel.js
+++ b/src/core/js/models/questionModel.js
@@ -239,7 +239,7 @@ define([
         setupCorrectFeedback: function() {
 
             this.set({
-                feedbackTitle: this.get('title'),
+                feedbackTitle: this.get("_feedback") ? this.get('_feedback').title : "",
                 feedbackMessage: this.get("_feedback") ? this.get("_feedback").correct : ""
             });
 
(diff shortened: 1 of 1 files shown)


moloko:
hi @chetan-hajare could you please make this so it's backwards-compatible i.e. if `feedback.title` is not set it should display the component's `title` instead

In fact I'd be open to having it initially fall back to `displayTitle` and only falling back to `title` if that's not set. As the original ticket states, it is a bit odd that it uses `title` when that's really only needed when there's less room on screen i.e. in PLP.

src/core/js/models/questionModel.js
```diff
@@ -238,29 +238,35 @@ define([
 
         setupCorrectFeedback: function() {
 
+            var feedback = this.get("_feedback");
+
             this.set({
-                feedbackTitle: this.get('title'),
-                feedbackMessage: this.get("_feedback") ? this.get("_feedback").correct : ""
+                feedbackTitle: feedback ? feedback.title : "",
```
oliverfoster:
```javascript
feedbackTitle: feedback.correctTitle || feedback.title ||  this.get('displayTitle') ||  this.get('title'),
```
^ would this work?

chetan-hajare: Sure @moloko !
Thanks @oliverfoster !

moloko: I don't think `correctTitle` is necessary but otherwise yes.

it's referenced enough times that it may be worth having a separate function like `setFeedback: function (message, graphic)` that handles adding in the appropriate title

oliverfoster:
What happens if I want to have a different title for correct/incorrect/partially correct?

src/core/js/models/questionModel.js
```diff
@@ -238,29 +238,35 @@ define([
...
         setupPartlyCorrectFeedback: function() {
 
-            if (this.get("_feedback") && this.get('_feedback')._partlyCorrect) {
-                if (this.get('_attemptsLeft') === 0 || !this.get('_feedback')._partlyCorrect.notFinal) {
-                    if (this.get('_feedback')._partlyCorrect.final) {
+            var feedback = this.get("_feedback");
+            
+            if(!feedback) return;
+
+            if (feedback && feedback._partlyCorrect) {
```
moloko: you don't need this check for `feedback` and the one in L254

tomgreenfield:
N.B. doesn't yet pull in `_feedback.title` when using individual feedback e.g.
```json
"_items": [
	{
		"text": "text",
		"_shouldBeSelected": true
	},
	{
		"text": "text",
		"_shouldBeSelected": false,
		"feedback": "feedback",
	},
	{
		"text": "text",
		"_shouldBeSelected": false,
		"feedback": "feedback"
	}
],
"_feedback": {
	"title": "My custom title",
	"correct": "correct"
}
```

moloko:
good catch @tomgreenfield  - adapt-contrib-mcq will need a little updated [here](https://github.com/adaptlearning/adapt-contrib-mcq/blob/master/js/mcqModel.js#L145-L150) which should solve for GMCQ as well. I think those are the only two that currently support option-specific feedback?

moloko:
@tomgreenfield https://github.com/adaptlearning/adapt-contrib-mcq/pull/139

PR #1958, opened 2018-02-07
Issue/1957

lc-thomasberger:
Add ability to add custom classes to notify view.
More settings ...
`_isCancellable` -> leave as is
`_closeOnBackdrop` -> disables call to cancelNotify when user clicks on shadow
`_showCloseButton` -> hides only the close button

### DIFF
--- a/src/core/js/views/notifyView.js
+++ b/src/core/js/views/notifyView.js
@@ -6,9 +6,7 @@ define([
 
         className: function() {
             let classes = 'notify ';
-            if (this.model.has('_classes')) {
-                classes += this.model.get('_classes');
-            }
+            classes += (this.model.get('_classes') || '');
             return classes;
         },
 
@@ -109,7 +107,7 @@ define([
 
         onShadowClicked: function(event) {
             event.preventDefault();
-            if (this.model.get("_closeOnBackdrop") === true) {
+            if (this.model.get("_closeOnShadowClick") === true) {
                 this.cancelNotify();
             }
         },
--- a/src/core/js/views/notifyView.js
+++ b/src/core/js/views/notifyView.js
@@ -107,15 +107,12 @@ define([
 
         onShadowClicked: function(event) {
             event.preventDefault();
-            if (this.model.get("_closeOnShadowClick") === true) {
-                this.cancelNotify();
-            }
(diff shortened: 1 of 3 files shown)


src/core/js/views/notifyView.js
```diff
@@ -101,7 +109,9 @@ define([
 
         onShadowClicked: function(event) {
             event.preventDefault();
-            this.cancelNotify();
+            if (this.model.get("_closeOnBackdrop") === true) {
```
oliverfoster: ```_closeOnShadowClick``` ?

src/core/templates/notify.hbs
```diff
@@ -41,7 +41,7 @@
             </div>
         </div>
         {{! toolbar moved for AA purposes - last tabbable element}}
-        {{#if _isCancellable}}
+        {{#if _showCloseButton}}
```
oliverfoster: This needs to be wrapping in both
```hbs
{{#if _isCancellable }}
```
and
```hbs
{{#if _showCloseButton}}
```
as the close button is "cancel" behaviour which is blocked by ```_isCancellable:false```, so you could end up with an unusable button by having
```json
{
  "_isCancellable": false,
  "_showCloseButton": true
}
```

src/core/js/views/notifyView.js
```diff
@@ -4,8 +4,16 @@ define([
 
     var NotifyView = Backbone.View.extend({
 
-        className: 'notify',
+        className: function() {
+            let classes = 'notify ';
+            if (this.model.has('_classes')) {
```
oliverfoster:
```js
classes += (this.model.get("_classes") || "");
```

src/core/js/views/notifyView.js
```diff
@@ -109,7 +107,7 @@ define([
 
         onShadowClicked: function(event) {
             event.preventDefault();
-            if (this.model.get("_closeOnBackdrop") === true) {
+            if (this.model.get("_closeOnShadowClick") === true) {
```
oliverfoster: this is the perfect place for returning early
```js
if (this.model.get("_closeOnShadowClick") === false) return;
this.cancelNotify();
```

PR #1959, opened 2018-02-07
Items component model

lc-thomasberger:
add setActiveItem method


src/core/js/models/itemsComponentModel.js
```diff
@@ -65,6 +64,11 @@ define([
 
         resetActiveItems: function() {
             this.get('_items').each(function(item) { item.toggleActive(false); });
+        },
+
+        setItemActive: function(index) {
```
oliverfoster: ```setActiveItem``` as the inverse of ```getActiveItem```

PR #1968, opened 2018-02-15
Issue 1699 adds better support for non-Spoor based offlineStorage

brian-learningpool:
A new event allows plugins to initialise offlineStorage conditionally, by waiting until they're ready.  Previously since Adapt.offlineStorage.ready is defaulted to true, plugins such as the Language Picker would skip any aynchronous offline storage setup.
- Added new 'offlineStorage:prepare' event, which is triggered before 'configModel:dataReady'
- Added Adapt.offlineStorage.hasHandler() function, this indicates if an offlineStorage handler has been setup

### DIFF
--- a/src/core/js/models/configModel.js
+++ b/src/core/js/models/configModel.js
@@ -25,29 +25,34 @@ define([
             // Then check if course files can load
             // 'configModel:loadCourseData' event starts the core content collections and models being fetched
             this.fetch({
-                success: _.bind(function() {
-                  Adapt.trigger('offlineStorage:prepare');
-                  Adapt.wait.queue(_.bind(function() {
+                success: function() {
+                    Adapt.trigger('offlineStorage:prepare');
 
-                    Adapt.trigger('configModel:dataLoaded');
-                    if (this.get('_canLoadData')) {
-                        Adapt.trigger('configModel:loadCourseData');
-                    }
-                    if(this.get('_defaultDirection')=='rtl'){//We're going to use rtl style
-                    	$('html').addClass('dir-rtl');
-                    }
-                    // check if animations should be disabled
-                    var disableAnimationArray = this.get('_disableAnimationFor');
-                    if (disableAnimationArray && disableAnimationArray.length > 0) {
-                        for (var i=0; i < disableAnimationArray.length; i++) {
-                            if ($("html").is(disableAnimationArray[i])) {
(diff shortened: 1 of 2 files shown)


src/core/js/models/configModel.js
```diff
@@ -26,6 +26,9 @@ define([
             // 'configModel:loadCourseData' event starts the core content collections and models being fetched
             this.fetch({
                 success: _.bind(function() {
+                  Adapt.trigger('offlineStorage:prepare');
+                  Adapt.wait.queue(_.bind(function() {
```
moloko: it's a small point but now we don't need to support stupid old IE browsers we can safely do `function.bind` instead of `_.bind`

I think this looks a lot cleaner:
```
 this.fetch({
	success: function() {
		Adapt.trigger('offlineStorage:prepare');
		Adapt.wait.queue(function() {
			Adapt.trigger('configModel:dataLoaded');
			if (this.get('_canLoadData')) {
				Adapt.trigger('configModel:loadCourseData');
			}
			
		}.bind(this));
	}.bind(this),
	error: function() {
		console.log('Unable to load course/config.json');
	}
});
```

src/core/js/models/configModel.js
```diff
@@ -26,6 +26,9 @@ define([
             // 'configModel:loadCourseData' event starts the core content collections and models being fetched
             this.fetch({
                 success: _.bind(function() {
+                  Adapt.trigger('offlineStorage:prepare');
```
moloko: indentation needs 4 spaces rather than 2

brian-learningpool:
Thanks @moloko.  I've updated the code.

src/core/js/offlineStorage.js
```diff
@@ -27,6 +27,14 @@ define([
             this._handler = handler;
         },
 
+        /**
+         * Flag to indicate if an offlineStorage handler has been defined.
+         * @returns {boolean} true if an offlineStorage handler has been defined, false otherwise
+         */
+        hasHandler: function() {
+            return !_.isUndefined(this._handler);
```
oliverfoster: `_.isUndefined` is an expensive function call, does it help with readability?
```js
  _.isUndefined = function(obj) {
    return obj === void 0;
  };
```
These are cheaper:
```js
return this._handler !== undefined;
return this._handler !== void 0;
```

Issue #1987, opened 2018-03-01
Lazy load graphics


oliverfoster:
### Subject of the enhancement
We should have the ability to lazy load graphics as the page scrolls.

* Should speed up initial loading on slower bandwidth devices
* Will reduce network overheads if all of the course isn't visited
* Should allow the centralisation of large / small device graphic alternatives

oliverfoster:
#2448

PR #1988, opened 2018-03-01
issue/1987: lazy graphic loading implementation

oliverfoster:
#1987 

### To Test
1. Change all template occurances of `<img src="{{_graphic.src}}"` to `<img {{graphic}}` (accordion, gmcq, graphic, narrative, menu and hotgraphic templates)
2. Remove image swapping functionality 
[graphic](https://github.com/adaptlearning/adapt-contrib-graphic/blob/master/js/adapt-contrib-graphic.js#L56-L58)
[gmcq](https://github.com/adaptlearning/adapt-contrib-gmcq/blob/master/js/adapt-contrib-gmcq.js#L50-L61)
3. Optionally add any of these to your `_graphic` objects in the json
```js
"_graphic": {
    "_width": 900, // natural width of dummy image
    "_height": 506, // natural height of dummy image
    "_color": "#ff00ff", // dummy image fill colour
    "large": "", // preferred source path for large devices
    "small": "", // preferred source path for small and medium devices
    "src": "", // fallback source path
    "_src": "", // fallback source path
    "_canLazyLoad": false // disable lazy load for this graphic
}
```
4. Change `config.json` `_className` property to target more devices

### Enhancement
* Centralised and reusable API for large/small/src/_src images
* Speed up page load when applied (ideally for mobile users)
* Cheap. Lightweight.
* Easy to extend and centralise image alt text and attribution if we wanted

### Issues
* By doing this we lose dynamic switching of images from graphic and gmcq on device size changes.... I'm not sure if that is important but it should be something easy to add back in, this time centrally.
* It should be noted that if the dummy image isn't the same ratio as the actual image, it'll cause a resize to trigger when the actual image gets inserted, might need to not have lazy loading unless height and width are specified?
* We should change all `img` tags where necessary to be `width:100%;` at the moment if the image is too small it doesn't stretch to fit the space.
* This is only enabled on ios and android devices, probably need to disable it by default
* Might want to cache the dummy images as they are created, need to do a performance profile of the canvas code

### Example
[https://rawgit.com/oliverfoster/lazy-images/master/index.html](https://rawgit.com/oliverfoster/lazy-images/master/index.html)
* Lazy loading on all browsers
* Lazy load on accordion, graphic, hotgraphic, narrative, gmcq and menu
* No heights or widths specified (so you get to see a resize)
* Scroll quickly, you can see the images lazy load

### DIFF
--- a/src/core/js/graphics.js
+++ b/src/core/js/graphics.js
@@ -1,181 +1,182 @@
 define([
-  'core/js/adapt',
-  'handlebars',
-  'inview'
-],function(Adapt, Handlebars) {
+    'core/js/adapt',
+    'handlebars',
+    'inview'
+], function(Adapt, Handlebars) {
 
-  var Graphics = Backbone.Controller.extend({
+    var Graphics = Backbone.Controller.extend({
 
-    initialize: function() {
-      _.bindAll(this, "_onScroll", "_delayedThreholdIncrease", "_checkImage");
-      this._onScroll = _.throttle(this._onScroll, 100);
-      this._delayedThreholdIncrease = _.debounce(this._delayedThreholdIncrease, 500);
-      this._offscreenPixelThreshold = 0;
-      this._registerHelpers();
-      this.listenTo(Adapt, "app:dataReady", this._onDataReady);
-    },
+        initialize: function() {
+            _.bindAll(this, "_onScroll", "_delayedThreholdIncrease", "_checkImage");
+            this._onScroll = _.throttle(this._onScroll, 100);
+            this._delayedThreholdIncrease = _.debounce(this._delayedThreholdIncrease, 500);
+            this._offscreenPixelThreshold = 0;
+            this._registerHelpers();
+            this.listenTo(Adapt, "app:dataReady", this._onDataReady);
+        },
 
-    _registerHelpers: function() {
+        _registerHelpers: function() {
 
-      var helpers = {
+            var helpers = {
 
-        "graphic": function(options) {
+                "graphic": function(options) {
 
-          var noArgs = (arguments.length === 1);
-          // if no options were passed and no local _graphic object is defined, return
-          if (noArgs && typeof this._graphic !== "object") return "";
-          // if no options were passed and the _graphic object is defined, use it
-          if (noArgs && this._graphic) options = this._graphic;
-          // if options were passed but were not an object, return
-          if (!options || typeof options !== "object") return "";
+                    var noArgs = (arguments.length === 1);
+                    // if no options were passed and no local _graphic object is defined, return
+                    if (noArgs && typeof this._graphic !== "object") return "";
+                    // if no options were passed and the _graphic object is defined, use it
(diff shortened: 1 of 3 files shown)


src/core/js/graphics.js
```diff
@@ -0,0 +1,181 @@
...
+    },
+
+    isEnabled: function() {
+      var graphics = Adapt.config.get("_graphics");
+      if (!graphics || !graphics._isEnabled) return false;
+      return true;
+    },
+
+    _setUpEventListeners() {
+      this.listenTo(Adapt, "pageView:postRender menuView:postRender", this._onPostRender);
```
moloko: switch to object style of `listenTo` here?

src/core/js/graphics.js
```diff
@@ -0,0 +1,181 @@
...
+      this.listenTo(Adapt, "remove", this._onRemove);
+    },
+
+    _onPostRender: function(view) {
+      // wait for page / menu to be ready
+      if (view.model.get("_isReady")) return this._start();
+      this.listenToOnce(view.model, "change:_isReady", this._start);
+    },
+
+    _start: function() {
```
moloko: rename to `_startScrollListener` to make the comment redundant?

src/core/js/graphics.js
```diff
@@ -0,0 +1,181 @@
...
+    },
+
+    _onScroll: function() {
+      var $progImages = $("img[data-adapt-graphics], video[data-adapt-graphics-poster]");
+      if (!$progImages.length) return this._stop();
+      $progImages.each(this._checkImage);
+      if (!this._offscreenPixelThreshold) this._delayedThreholdIncrease();
+    },
+
+    _stop: function() {
```
moloko: rename to `_stopScrollListener` to make the comment redundant?

oliverfoster:
@chucklorenz + @guywillis "...Multiple images per graphic definition would account for retina displays and/or different images for different devices sizes. Dynamic image switching for responsiveness is absolutely necessary...."

joe-replin:
Another thought. I've noticed on some popular websites an alternative method to lazy loading in which low-quality image placeholders with a blur filter overlay are present until the higher-resolution images are loaded or called when in-view. I found this post to be insightful on the method.
[](https://jmperezperez.com/medium-image-progressive-loading-placeholder/)

joe-replin:
[](https://jmperezperez.com/medium-image-progressive-loading-placeholder/)

moloko:
also worth noting https://addyosmani.com/blog/lazy-loading/

brian-learningpool:
Not to dismiss the effort @oliverfoster has put in but I actually really like the progressive enhancement approach outlined in that article @moloko.  Just a pity it doesn't currently seem to be there in Chrome 76.

oliverfoster:
Cool. I'll revise this so that it just centralises the graphic loading for the moment and drop the lazy loading etc until later.

Something like:
```hbs
{{{graphic}}}
```
Just to pick out the graphic properties and render an image from them. With large / small support etc.

PR #1999, opened 2018-03-06
issue/1998: remove buggy ios fixes

### LINKED ISSUE #1998: Remove ios specific a11y fixes
### Subject of the issue
Buggy code needs removing from a11y library.

oliverfoster:
#1998 

* removed ios9 focus correction from user input controls (text or select)
* removed ios9 focus correction for interacting with `:before` and `:after` elements
* removed restrictions on focus outline as the rest of the code doesn't apply the same restrictions
* removed redundant ie8 class selector for focused elements (ie8 didn't have the `:focus` selector and had a legacy fix to add a `.focused` class, which has since been removed)

### DIFF
--- a/src/core/less/base.less
+++ b/src/core/less/base.less
@@ -29,7 +29,7 @@
 }
 
 .accessibility {
-    .focused, *:focus {
+    *:focus {
         outline: @focus-outline;
     }
 }
--- a/src/core/less/base.less
+++ b/src/core/less/base.less
@@ -29,7 +29,7 @@
 }
 
 .accessibility {
-    *:focus {
+    :focus {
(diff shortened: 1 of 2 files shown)


src/core/less/base.less
```diff
@@ -28,8 +28,8 @@
     display:none!important;
 }
 
-.accessibility.no-touchevents {
-    .focused, button:focus, select:focus, textarea:focus, a:focus, input:focus + label {
+.accessibility {
+    *:focus {
```
tomgreenfield: Would you mind ditching the asterisk? It's implied and we don't seem to use it elsewhere.

Issue #1981, opened 2018-02-22
allow offlineStorage to store data for the duration of the session when no handler is defined


moloko:
Currently if I have a piece of data I want to save 'between sessions' I need to store it in two places - in offlineStorage and else where (typically on the model) - on account of how the offlineStorage API set/get calls do nothing and simply `return` when no handler is defined.

This is duplication and therefore prone to error.

It seems it would be easy enough to implement basic storage so that the offlineStorage get/set calls actually function properly when you're running the course from a regular web server, as you typically do during development. It should NOT be expected to remember data between sessions, that should be the responsibility of plugins like spoor.

PR #2014, opened 2018-03-21
update to allow offlineStorage to store data when no handler has been defined

moloko:
see #1981


moloko:
@tomgreenfield I'll go one better than that and remove the need for the `!` as well

PR #2022, opened 2018-03-22
Added any and all helpers

### LINKED ISSUE #2021: Expand helper js to include ifor and ifand
Update the core helper js to include ifor and ifand statements.

Example of ifor use:
Only render the header section of an article / block / component when **either** the displayTitle, body, or instruction is present. 

```
{{#any displayTitle body instruction}}
<div class="component__header {{_component}}__header">
</div>
{{/any}}
```

Example of ifand use:
Only render the header section of an article / block / component when the displayTitle, body, and instruction are **all** present.

```
{{#all displayTitle body instruction}}
<div class="component__header {{_component}}__header">
</div>
{{/all}}
```

Example of ifand ifor use:
```
{{#all (any displayTitle title) body instruction}}
<div class="component__header {{_component}}__header">
</div>
{{/all}}
```

guywillis:
resolves #2021

### DIFF
--- a/src/core/js/helpers.js
+++ b/src/core/js/helpers.js
@@ -66,9 +66,13 @@ define([
             case "%": return lvalue % rvalue;
             }
         },
-
+        
         /**
-         * if (conditionA || conditionB)
+         * Equivalent to 
+         * @example
+         * {{#any displayTitle body instruction}}
+         * <div class="component__header {{_component}}__header"></div>
+         * {{/any}}
          */
         any: function() {
             var args = Array.prototype.slice.call(arguments, 0, -1);
@@ -78,7 +82,11 @@ define([
         },
 
         /**
-         * if (conditionA && conditionB)
+         * Equivalent to 
+         * @example
+         * {{#all displayTitle body instruction}}
+         * <div class="component__header {{_component}}__header"></div>
+         * {{/all}}
          */
         all: function() {
             var args = Array.prototype.slice.call(arguments, 0, -1);
--- a/src/core/js/helpers.js
+++ b/src/core/js/helpers.js
@@ -69,7 +69,7 @@ define([
         
         /**
          * Equivalent to 
-         * @example
+         * if (conditionA || conditionB)
          * {{#any displayTitle body instruction}}
          * <div class="component__header {{_component}}__header"></div>
          * {{/any}}
@@ -83,7 +83,7 @@ define([
 
         /**
          * Equivalent to 
-         * @example
+         * if (conditionA && conditionB)
          * {{#all displayTitle body instruction}}
          * <div class="component__header {{_component}}__header"></div>
          * {{/all}}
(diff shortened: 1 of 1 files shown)


tomgreenfield:
Suggest some more readable helper names? I keep reading as i-for and i-fand.

Possibly:
```{{#either ...}}```
```{{#all ...}}```

File: src/core/js/helpers.js [hunk 1]
```diff
@@ -67,6 +67,28 @@ define([
             }
         },
 
+        ifor: function() {
+            var args = Array.prototype.slice.call(arguments, 0, -1);
+            var block = Array.prototype.slice.call(arguments, -1)[0];
+
+            for (var i = 0, l = args.length; i < l; i++) {
+                if (args[i]) return block.fn(this);
+            }
```
oliverfoster: use:

```js
return _.some(args) ? block.fn(this) : block.inverse(this);
```

src/core/js/helpers.js
```diff
@@ -67,6 +67,28 @@ define([
...
+            return block.inverse(this);
+        },
+
+        ifand: function() {
+            var args = Array.prototype.slice.call(arguments, 0, -1);
+            var block = Array.prototype.slice.call(arguments, -1)[0];
+
+            for (var i = 0, l = args.length; i < l; i++) {
+                if (!args[i]) return block.inverse(this);
+            }
```
oliverfoster:
```js
return args.every() ? block.fn(this) : block.inverse(this);
```

src/core/js/helpers.js
```diff
@@ -67,6 +67,28 @@ define([
             }
         },
 
+        ifor: function() {
```
oliverfoster: rename to `any`
add comment

src/core/js/helpers.js
```diff
@@ -67,6 +67,28 @@ define([
...
+            var block = Array.prototype.slice.call(arguments, -1)[0];
+
+            for (var i = 0, l = args.length; i < l; i++) {
+                if (args[i]) return block.fn(this);
+            }
+
+            return block.inverse(this);
+        },
+
+        ifand: function() {
```
oliverfoster: rename to `all`
add comment

oliverfoster:
```js
return args.every() ? block.fn(this) : block.inverse(this);
```

src/core/js/helpers.js
```diff
@@ -67,6 +67,26 @@ define([
             }
         },
 
+        /**
+         * if (conditionA || conditionB)
+         */
```
moloko: could you amend to:
```
/**
 * Equivalent to 
 * @example
 * {{#any displayTitle body instruction}}
 * <div class="component__header {{_component}}__header"></div>
 * {{/any}}
 */
```

moloko: and something similar for `any`?

guywillis:
Amended comments

chetan-hajare:
Since the content of both the functions are same except(_.any and _.all), can we have a common function and pass "any"/"all" as parameter? 
Otherwise I am fine with current structure as well.

PR #2053, opened 2018-04-25
accessibility: add aria-roledescription and aria-modal attributes

### LINKED ISSUE #2052: give notify popup some more appropriate ARIA attributes
#2051 suggests adding `aria-haspopup="dialog"` to hot graphic pins, in which case the Notify popup will need to have a matching role of `"dialog"`.

In addition it has been suggested that, as a modal popup, it should have an `aria-modal` attribute

moloko:
resolves #2052 

screenshot of the Notify element HTML with this change applied:
2018-04-25 16_14_37-developer tools - adapt - http___localhost_9001_ _id_co-10

### DIFF
--- a/src/core/js/views/notifyPushView.js
+++ b/src/core/js/views/notifyPushView.js
@@ -19,9 +19,6 @@ define([
                 'change:_index': this.updatePushPosition
             });
 
-            // Include accessibility globals in notify model.
-            this.model.set('_globals', Adapt.course.get('_globals'));
-
             this.preRender();
             this.render();
         },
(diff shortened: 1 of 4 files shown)


oliverfoster:
we should probably remove the popup opened aria label. will test to see if it's needed

oliverfoster:
can you switch to: 
```js
        {
            'role': 'dialog',
            'aria-model': 'true',
            'aria-labelledby': 'notify-heading'
        },
```

In the handlebars, remove the "popup opened" aria label and add `id="notify-heading"` to the `.notify-popup-title` div?

Issue #2071, opened 2018-05-22
ItemsComponentModel has obtuse implementation


oliverfoster:
### Subject of the issue/enhancement/features
The way the `ItemsComponentModel` is written with regards to `_items` breaks backward compatibility with existing functionality unnecessarily.
`_items` is replaced by a Collection of Models which makes any existing code relying on `_items` being an array of objects stop working [at this line](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/models/itemsComponentModel.js#L30)
`toJSON` is then overridden to compensate, [at this line](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/models/itemsComponentModel.js#L8-L13)

### Expected behaviour
`_items` should return an array of objects.
`_children` should represent the collection, as with the other models.

moloko:
I think this is all done now?

PR #2072, opened 2018-05-22
fixes issue/#2071 : itemsComponentModel switched to _children

oliverfoster:
#2071 

Switched to `_children` from `_items`.

### DIFF
--- a/src/core/js/models/itemsComponentModel.js
+++ b/src/core/js/models/itemsComponentModel.js
@@ -5,17 +5,10 @@ define([
 
     var ItemsComponentModel = ComponentModel.extend({
 
-        toJSON: function() {
-            var json = _.clone(this.attributes);
-            json._items = this.get('_items').toJSON();
-
-            return json;
-        },
-
         init: function() {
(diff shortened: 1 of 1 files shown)


src/core/js/models/itemsComponentModel.js
```diff
@@ -7,47 +7,47 @@ define([
...
         setUpItems: function() {
-            var items = this.get('_items').map(function(item, index) {
+            this.get('_items').forEach(function(item, index) {
                 item._index = index;
 
                 return item;
             });
 
-            this.set('_items', new Backbone.Collection(items, { model: ItemModel }));
+            this.set('_children', new Backbone.Collection(this.get('_items'), { model: ItemModel }));
```
tomgreenfield: Cache `items` and remove `return`?
```js
setUpItems: function() {
    var items = this.get('_items');

    items.forEach(function(item, index) { item._index = index; });
    this.set('_children', new Backbone.Collection(items, { model: ItemModel }));
}
```

Issue #2094, opened 2018-05-24
Content bleeds into popup


oliverfoster:
### Subject of the issue
When a popup is open the rest of the document should be inaccessible to the screen reader.
Currently this works for tabbing but not for the screen reader cursor.

oliverfoster:
merged

PR #2095, opened 2018-05-24
accessibility: issue/#2094 popups

oliverfoster:
#2094

* Siblings of the parent branch of the popup tag become `aria-hidden=true tabindex="-1"`
* Tab index original values are better restored by accounting for no tab index rather than defaulting to `0`

### To test

Using JAWS, enable accessibility mode, open the drawer or a notify. Use the up and down arrow keys to navigation around. You should not be able to read any content behind the popup.

### DIFF
--- a/src/core/js/libraries/jquery.a11y.js
+++ b/src/core/js/libraries/jquery.a11y.js
@@ -742,6 +742,7 @@
             floorStack: [$("body")],
             focusStack: [],
             tabIndexes: {},
+            ariaHiddens: {},
             elementUIDIndex: 0,
             scrollDisabledElements: null,
             scrollStartEvent: null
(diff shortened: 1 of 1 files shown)


src/core/js/libraries/jquery.a11y.js
```diff
@@ -1067,6 +1068,10 @@
                 $hideable = $(domSelectors.hideableElements).filter(domFilters.globalTabIndexElementFilter);
             }
 
+            var $branch = this.add(this.parents());
+            var $siblings = $branch.siblings().filter(domFilters.globalTabIndexElementFilter);
```
tomgreenfield: When we spoke briefly about this, you said you may want to switch to `nextAll()`.

oliverfoster: this was a different bit of code i was meaning. i misunderstood you.
this bit gets all of the elements which aren't associated with the branch your element exists on. 
such that all other elements other than the branch from root to tip that you need can become excluded to the screen reader.

oliverfoster:
@tomgreenfield could you +1 ?

src/core/js/libraries/jquery.a11y.js
chris-steele: @oliverfoster was struggling to get my head around this, but I think the key is to realise that when the ancestors and their siblings are made aria-hidden then by virtue _their_ descendants are hidden. Hope I've got that right.

oliverfoster: Exactly. This happens through each parent towards the document. The parent's siblings are made aria hidden. As if picking off all the branches down a twig except the one you need to be available.

Issue #1863, opened 2017-11-10
Double Aria Label Argument


guywillis:
Slightly niche but it would be great to be able to support multiple aria label arguments for Accessibility.

For example:

If I wanted to piece together multiple sections of language to form a new aria label without creating new and custom JSON nodes, I would have to manually write the statement like the below:

```
{{#if duration}}
<span class="aria-label" tabindex="0" role="region">{{{_globals._menu._boxmenu.durationLabel}}} {{{duration}}}</span>
{{/if}}
```

It would be awesome to be able to do something like the below:

```
{{#if duration}}
{{{a11y_aria_label _globals._menu._boxmenu.durationLabel duration}}}
{{/if}}
```

oliverfoster:
merged

PR #2106, opened 2018-05-25
accessibility: issue/#1863 - expand a11y text helpers to take multiple arguments

oliverfoster:
#1863 
* Expanded `a11y_normalize`, `a11y_remove_breaks`, `a11y_aria_label` and `a11y_aria_label_relative` to accept and concatenate multiple arguments


src/core/js/accessibility.js
```diff
@@ -113,23 +113,31 @@ define([
                     }
                 },
 
-                a11y_normalize: function(text) {
-                    return $.a11y_normalize(text);
+                a11y_normalize: function(texts) {
+                    var values = Array.prototype.slice.call(arguments, 0,-1);
+                    values = values.filter(function(v) { return v });
```
tomgreenfield: Remember you can use that syntactic sugar `.filter(Boolean)` here.

Issue #1659, opened 2017-07-13
notifyView::closeNotify can be invoked multiple times


chris-steele:
Can be replicated by multiple clicks of close/alert/prompt buttons or multiple escape key presses (in quick succession).

This has the potential to cause serious problems (yep, it did for me in a bespoke Adapt course) given what `closeNotify` does and the events it triggers (propagating to `a11y` lib etc). Just need to remove it in same way as `drawerView` does.

chris-steele:
A general sweep of lingering event handlers (particularly where things are animated) would be worth doing.

PR #2110, opened 2018-05-29
Issue/#1659

jamesrea83:
re https://github.com/adaptlearning/adapt_framework/issues/1659

### DIFF
--- a/src/core/js/views/notifyView.js
+++ b/src/core/js/views/notifyView.js
@@ -143,7 +143,7 @@ define([
         },
 
         showNotify: function() {
-
+            this.model.set('_isActive', true);
             this.addSubView();
 
             Adapt.trigger('notify:opened', this);
@@ -200,6 +200,8 @@ define([
         },
 
         closeNotify: function (event) {
+            if (!this.model.get('_isActive')) return;
+            this.model.set('_isActive', false);
 
             if (this.disableAnimation) {
 
--- a/src/core/js/views/notifyView.js
+++ b/src/core/js/views/notifyView.js
@@ -14,6 +14,8 @@ define([
 
         escapeKeyAttached: false,
 
+        isOpen: false,
+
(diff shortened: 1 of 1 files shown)


src/core/js/views/notifyView.js
```diff
@@ -200,6 +200,8 @@ define([
         },
 
         closeNotify: function (event) {
+            if (!this.model.get('_isActive')) return;
```
moloko: Picky I know but `_isOpen` feels like a better name...?

Also, could you add a comment:
```
//prevent from being invoked multiple times - see https://github.com/adaptlearning/adapt_framework/issues/1659
```

src/core/js/views/notifyView.js
```diff
@@ -143,7 +145,7 @@ define([
         },
 
         showNotify: function() {
-
+            this._isOpen = true;
```
tomgreenfield: Please remove the underscore from `_isOpen` to match line 17.

src/core/js/views/notifyView.js
```diff
@@ -200,6 +202,9 @@ define([
         },
 
         closeNotify: function (event) {
+            //prevent from being invoked multiple times - see https://github.com/adaptlearning/adapt_framework/issues/1659
+            if (!this._isOpen) return;
+            this._isOpen = false;
```
tomgreenfield: Please remove the underscore from `_isOpen` to match line 17.

moloko:
I suppose we should PR this into the 'legacy' branch as well...?

oliverfoster:
Can we retire legacy now?
Put it out to pasture?

moloko:
@oliverfoster we originally said we'd review at the end of 2018. It's not exactly requiring a lot of effort to maintain though, this is actually the first time we've done anything to it since Adapt v3 was released.

Issue #2122, opened 2018-06-11
Add appropriate aria regions throughout.


oliverfoster:
### Subject of the issue
Some sections need to be grouped by region to make them more easily digested as a unit of learning.
* component
* navigation
* drawer 
* page

oliverfoster:
merged

PR #2123, opened 2018-06-11
accessibility: issue/#2122 - added regions

oliverfoster:
#2122 
* Added appropriate regions


src/core/js/views/componentView.js
```diff
@@ -5,6 +5,14 @@ define([
 
     var ComponentView = AdaptView.extend({
 
+        attributes: function() {
+            return {
+                "data-adapt-id": this.model.get('_id'),
+                "aria-labelledby": this.model.get('_id')+"-heading",
+                "role": "region"
+            };
+        },
```
tomgreenfield: Would like to see this extending the results of the parent's [function](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/adaptView.js#L7-L11) for future-proofing.

_Note: fix results helper to pass through correct context._

Issue #2138, opened 2018-06-18
Drawer isn't hidden from screen reader when closed


oliverfoster:
### Subject of the issue
Aria-hidden doesn't get applied to the lasting drawer element when it is not shown.

oliverfoster:
merged

PR #2139, opened 2018-06-18
accessibility: issue/#2138 drawer screenreader

oliverfoster:
#2138 
* Added aria-hidden to drawer when closed

### DIFF
--- a/src/core/js/views/drawerView.js
+++ b/src/core/js/views/drawerView.js
@@ -11,7 +11,7 @@ define([
             'role': 'dialog',
             'aria-modal': 'true',
             'aria-labelledby': 'drawer-heading',
-            'tabindex': '-1'
+            'aria-hidden': 'true'
         },
 
         initialize: function() {
@@ -118,7 +118,7 @@ define([
         },
 
         showDrawer: function(emptyDrawer) {
-            this.$el.removeClass('display-none');
+            this.$el.removeClass('display-none').removeAttr('aria-hidden');
             //only trigger popup:opened if drawer is visible, pass popup manager drawer element
             if (!this._isVisible) {
                 Adapt.trigger('popup:opened', this.$el);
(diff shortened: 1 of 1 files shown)


src/core/js/views/drawerView.js
```diff
@@ -118,7 +118,7 @@ define([
         },
 
         showDrawer: function(emptyDrawer) {
-            this.$el.removeClass('display-none');
+            this.$el.removeClass('display-none').removeAttr('aria-hidden');
```
lc-thomasberger: With accessibility becoming a default I think we should use aria-attributes to control an elements visibility instead of maintaining classes.
```
.drawer[aria-hidden="true"] {
  display: none;
}
```

oliverfoster: aria-hidden != display: none

there are quite a number of places where you'd want aria-hidden by display:block
i'm not sure this association is a good way to think about it.

in this instance it would work, i can change it if you want?

lc-thomasberger: I did not state that aria-hidden should always be set to display none.
Instead of adding a class to handle the state based styling we could use the attribute which is set anyway. This would make it also more descriptive and make sure that we use the appropriate aria labels.

oliverfoster: The aria spec is intentionally separate from css, html and ecmascript and is specifically for screen readers only, using it for styling seems a counter intuitive to me, like using a data attribute for styling instead of a class. It's the separation of concerns I was trying to speak to.

lc-thomasberger: Fine for me ... just thought we can reuse logic ...

oliverfoster: reference article: https://medium.com/@heyoka/use-aria-attributes-for-javascript-state-setting-styling-efa87457a613

Issue #2058, opened 2018-04-30
add ability to easily filter the results of findDescendantModels


moloko:
With `adaptModel.findDescendants`, if you wanted to do something like get all available, mandatory components for a contentObject, you could do this:
```js
Adapt.findById('co-05').findDescendants('components').where({'_isAvailable': true, '_isOptional': false});
```
However, following #1616 this function is deprecated in favour of `findDescendantModels` - but as this function returns an Array rather than a Backbone Collection it's necessary to do the following to achieve the same result:
```js
_.filter(Adapt.findById('co-05').findDescendantModels('components'), function(comp) {
    return comp.get('_isAvailable') === true && comp.get('_isOptional') === false;
});
```
Which is obviously quite a bit wordier compared to the original.

I think it would therefore be useful to amend `findDescendantModels` to have an optional 2nd parameter `filterBy` that would give us similar functionality to the original version.

moloko:
OK I'm obviously being a bit thick here and completely failing to realise that you could just do:
```js
_.where(Adapt.findById('co-05').findDescendantModels('components'), {'_isAvailable': true, '_isOptional': false});
```
I blame the code in PLP's completionCalculations.js which uses `_.filter` even though you could achieve the same result in less code by using `_.where`...!

However, still might be useful to include this 2nd parameter in the function anyway just to make it even easier? i.e. 
```js
Adapt.findById('co-05').findDescendantModels('components', {'_isAvailable': true, '_isOptional': false});
```

oliverfoster:
You can't do the `_.where` because findDescendantModels returns an array of models. You were right first time.

moloko:
ah of course it does! phew...

oliverfoster:
You might find http://underscorejs.org/#isMatch useful for comparing the `options.filterBy = { name: value }` to the `model.toJSON()` 
```js
if (options && options.filterBy) {
  return _.filter(models, function(model) {
     return _.isMatch(model.toJSON(), options.filterBy);
  });
}
```

moloko:
Oh that's interesting, I'll give that a try - definitely shorter than what I have at the moment:
```js
return _.filter(returnedDescendants, function(descendant){
    for (var property in filterBy) {
        var value = filterBy[property];
        if (descendant.get(property) !== value) {
            return false;
        }
    }
    return true;
});
```

oliverfoster:
It might also be slower than what you have.

oliverfoster:
[...]

PR #2141, opened 2018-06-20
updated findDescendantModels to add ability to filter the results

moloko:


src/core/js/models/adaptModel.js
```diff
@@ -287,24 +289,42 @@ define([
...
+            } else {
+                searchChildren(this.getChildren().models);
+            }
 
-            return returnedDescendants;
+            if (!options) {
+                return returnedDescendants;
+            }
+
+            if (options.where) {
```
tomgreenfield: Please can we return early:
```js
var properties = options.where;

if (!properties) return;
```

src/core/js/models/adaptModel.js
```diff
@@ -287,24 +289,42 @@ define([
...
+                return _.filter(returnedDescendants, function(descendant) {
```
tomgreenfield: Can use native `.filter()`:
```js
return returnedDescendants.filter(function(descendant) {
```

src/core/js/models/adaptModel.js
```diff
@@ -287,24 +289,42 @@ define([
...
+                    for (var property in options.where) {
```
tomgreenfield: Are we particularly against using Underscore here? Should readability outweigh being performant?
```js
return _.findWhere(descendant, properties);
```

src/core/js/models/adaptModel.js
moloko: @tomgreenfield the reason I didn't return early here is that there's an expectation that further `options` will be added in future - see [this suggestion](https://github.com/adaptlearning/adapt_framework/issues/2058#issuecomment-385449881) from @oliverfoster

src/core/js/models/adaptModel.js
moloko: [I did already have a bit of a look at this](https://github.com/adaptlearning/adapt_framework/issues/2058#issuecomment-386673476) and felt that on balance `for...in` was the better option

Issue #2154, opened 2018-07-04
On feedback popup open, background content is scrollable.


NayanKhedkar:
### Subject of the issue/enhancement/features
On feedback popup open, background content is scrollable.

### Your environment
* Framework 3.x.x
* Mobile device(android);

### Steps to reproduce
enable _canShowFeedback on any question component.
on answer submitting feedback popup will open try to scroll content using scrollbar(on desktop) and swipe on mobile devices.

### Expected behaviour
Background content should not be scrollable when feedback popup opened.

### Actual behaviour
on feedback(notify) popup open user can easily scroll background content using scrollbar(on desktop) and swipe on mobile devices.

oliverfoster:
I don't think you'll be able to fix the scroll bar on desktop without moving the page into its own scrolling container.
I've been meaning to do this for a while as it will help mitigate ios iframe scrolling issues and help with accessibility virtual cursor placement, but it is a huge change.

NayanKhedkar:
Is it possible to add **overflow:hidden** to html tag on popup open and remove on popup close.I think it will fix the problem.

oliverfoster:
Nope, the scroll position will get lost.

oliverfoster:
Could you test out #2169? See if that helps?

NayanKhedkar:
Sure.

NayanKhedkar:
It fixed the problem. :+1:

PR #2155, opened 2018-07-04
issue/#2154 fixed passive event attachment

oliverfoster:
#2154 
* Fixed event attachment to allow preventDefault()

### DIFF
--- a/src/core/js/libraries/jquery.a11y.js
+++ b/src/core/js/libraries/jquery.a11y.js
@@ -583,6 +583,9 @@
             return true;
         }
 
+        function nativePreventScroll(event) {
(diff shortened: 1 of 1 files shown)


src/core/js/libraries/jquery.a11y.js
```diff
@@ -583,14 +583,17 @@
             return true;
         }
 
+        function nativePreventScroll(event) {
+            return preventScroll($.event.fix(event));
```
tomgreenfield: Could probably do with a comment here along the lines of what you've already said in the issue.
> I've also used an intermediate function to turn the native event object into a jquery event object - which the preventScroll function is expecting.

Issue #2168, opened 2018-07-05
Change scrolling from document to container


oliverfoster:
### Subject of the enhancement
There are a number of tickets relating to scrolling and iframes which can be solved by incorporating some of the theory and code from [iOSScrollFIx](https://github.com/cgkineo/adapt-iOSScrollFix).

#### Potentially solvable tickets
#585 Responsiveness inside iframes

#### Solvable tickets
#751 Scrolling inside iframes
#835 Item 7
#2154 Document scrolls with scrollbar on desktop
#2167 Content appears under navigation bar when using screen reader

#### Change scrollable element from document to `#app`
0. Add new element `#app` containing the `#wrapper`, fwk + [scorm html](https://github.com/adaptlearning/adapt-contrib-spoor/pull/178) 
1. Make `@navigation-height` variable in core and [theme override](https://github.com/adaptlearning/adapt-contrib-vanilla/pull/223)
2. Make `.navigation` `{ position: relative; }`
3. Make `.navigation` add itself before the `#app` element
3. Make `html, body`  `{ height: 100%; overflow: hidden; }`
4. Make `#app` `{ overflow-y: scroll; height: calc(100% - @navigation-height); }`
#### Make sure window orientated scrolling code scrolls `#app` instead
5. Change `jquery.scrollTo` library functions to use the `#app` instead of `window` 
6. Change `jquery.scrollTop` native function to use the `#app` instead of `window`  
7. Make sure `#app` `scroll` events trigger `window` `scroll` events
8. Change `window.scrollTo` to scroll `#app` instead of `window`
#### Make `#app` scroll on iOS inside iframe
9. `#app` has `{ -webkit-overflow-scrolling: touch; }`
#### Make trickle get resize offset from new function
10. Add `Adapt.getOffset(selector)` to get the offset of `selector` inside `#app`
11. [Make trickle use `Adapt.getOffset` rather than `jquery.offset`](https://github.com/adaptlearning/adapt-contrib-trickle/pull/89)

### Potential issues
* May break any extension with complex scrolling configurations

### Plan for backward compatibility
* Provide a config option for enabling and disabling new behaviour, default enabled
  `config.json` `"_scrolling": { "_isEnabled": true }`
* Add injection of `#app`

oliverfoster:
merged

PR #2169, opened 2018-07-05
accessibility: issue/#2168 Changed scrolling from document to container

oliverfoster:
#2168 
* Changed scrolling from document to container

### Test with
[trickle#89](https://github.com/adaptlearning/adapt-contrib-trickle/pull/89)
[vanilla#223](https://github.com/adaptlearning/adapt-contrib-vanilla/pull/223)
[spoor#178](https://github.com/adaptlearning/adapt-contrib-spoor/pull/178)

### Test in
* iOS inside + outside iframe
* Any other browsers

List of outstanding PRs: https://github.com/adaptlearning/adapt_framework/issues/2206.


src/core/js/scrolling.js
```diff
@@ -0,0 +1,139 @@
...
+        var $app = Adapt.scrolling.$app;
+        var $element = $(selector);
+        var offset = $element.offset();
+        var $stack = $element.parents().add($element);
+        var $scrollParents = $stack.filter('#app');
+        $scrollParents.each(function(index, parent) {
+            var $parent = $(parent);
+            var scrolltop = parseInt($parent.scrollTop());
+            var scrollleft = parseInt($parent.scrollLeft());
+            offset.top += scrolltop - $app.offset()['top'];
```
chris-steele: I can understand adding the scrollTop/scrollLeft values so you can discount the effects of scrolling to get the offset position, but I'm, confused by the subtraction of `$app.offset()['top']`. Let's say the $app was offset from the top by 50px, then if $element is 10 levels deep under the document you'll subtract 10 x 50px (well 9 x 50px) - I must be interpreting this badly - can you help?

oliverfoster: There will only ever bee one `$scrollParents` the `#app` element. I should rewrite this to make that clearer.

Line 127: Make array of parents plus subject element
Line 128: Filter array so that only `#app` element persists
Line 129: Iterate through array.

I'll rewrite it.

chris-steele: Oh that was a bad read of `filter`, sorry! So, with there only being one `#app`, why bother with line 128 and the loop?

oliverfoster: The code evolved over a few iterations. It just ended up like that.

tomgreenfield:
It'd be nice to have one more attribute so we can have a concentrated fix for iOS iframe issues without affecting other devices, e.g.

<table>
<tr>
<th></th><th>Screenreader support</th><th>iOS iframe support</th><th>Native chrome/scrolling support</th>
</tr>
<tr>
<td><pre>"_scrollingContainer": {
    "_isEnabled": false,
    "_limitToSelector": ""
}</pre></td>
<td>❎</td>
<td>❎</td>
<td>✅</td>
</tr>
<tr>
<td><pre>"_scrollingContainer": {
    "_isEnabled": true,
    "_limitToSelector": ""
}</pre></td>
<td>✅</td>
<td>✅</td>
<td>❎</td>
</tr>
<tr>
<td><pre>"_scrollingContainer": {
    "_isEnabled": true,
    "_limitToSelector": ".os-ios"
}</pre></td>
<td>❎</td>
<td>✅</td>
<td>✅</td>
</tr>

src/core/less/scrolling.less
```diff
@@ -0,0 +1,28 @@
...
+
+  #app {
+    position: relative;
+    height: ~"calc(100% - @{navigation-height})";
+    overflow-y: scroll;
+    -webkit-overflow-scrolling: touch;
+  }
+
+  #wrapper {
+    padding: 0 !important;
```
tomgreenfield: Just checking the need for `!important`s?

oliverfoster: Some themes will have #wrapper padding values at 64px or similar, this is just to force it to 0 for adapt-scrolling courses.

tomgreenfield:
I may have jumped the mark on my reivew; the scrolling container seems to have broken `Adapt.navigateToElement()`:
```console
Uncaught TypeError: Cannot read property 'duration' of undefined
    at n.Adapt.scrollTo (scrolling.js:115)
    at n.Adapt.navigateToElement (adapt.js:148)
```

src/core/less/scrolling.less
moloko: worth adding comment (to the .less) to that effect?

guywillis: The specificity of the parent selector, ```html.adapt-scrolling```, should mean that you don't need to add the```!important``` declaration.

Agreed with @moloko that a comment outlining why the override is in place would help resolve any potential confusion.

oliverfoster: Done

tomgreenfield: Would prefer to lose any unnecessary `!important`s since the selectors are already more than specific enough.

src/core/schema/config.model.schema
```diff
@@ -358,6 +358,30 @@
...
+          "title": "Enabled?"
+        },
+        "_limitToSelector": {
+          "type": "string",
+          "required": false,
+          "default": "",
+          "title": "Limit to selector",
+          "inputType": "Text",
+          "validators": [],
+          "help": "A comma-separated list of classes to look for on the HTML element when determining whether this should be active or not."
```
tomgreenfield: Please could we encourage using the selector first and foremost? (I know it now does `.hasClass()` behind the scenes too but that, and the help text, is a hangover from the older implementation).

Issue #2180, opened 2018-07-24
findDescendantModels bug with items _children


oliverfoster:
### Subject of the issue
findDescendantModels is attempting to search components with items models. itemModel doesn't have a getChildren() function compatible with the findDescendantModels function and so an error is thrown.

moloko:
done in 5a58569468de4a05e2b8b9f0abdd679b79cbe448

PR #2181, opened 2018-07-24
issue/#2180 Return if no defined _children type

oliverfoster:
#2180 
* Stop findDescendantModels descending into models with no defined _children type - itemsComponentModel specifically at the moment.

### DIFF
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -260,7 +260,7 @@ define([
 
         /**
          * Returns all the descendant models of a specific type
-         * @param {string} descendants Valid values are 'contentObject', 'article', 'block' or 'component'
+         * @param {string} descendants Valid values are 'contentObjects', 'articles', 'blocks' or 'components'
          * @param {object} options an object that defines the search type and the properties/values to search on. Currently only the `where` search type (equivalent to `Backbone.Collection.where()`) is supported.
          * @return {array}
          * @example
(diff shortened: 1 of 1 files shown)


oliverfoster:
yup, it's about time this one had a mahoosive refactor. it's utterly unreadable, gis a few moments.

oliverfoster:
sorry guys, this needs reviewing again

moloko:
Oh, and I've just spotted that my comment for the `descendants` parameter is wrong, it should read "@param {string} descendants Valid values are 'contentObject**s**', 'article**s**', 'block**s**' or 'component**s**'" - would you mind correcting this while you're here?

oliverfoster:
It used to search [model _children properties](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/models/blockModel.js#L9) but now it searches [model _type attributes](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/models/adaptModel.js#L38). The old ones are pluralised, the new ones are not, I have to strip the 's' from the descendants variable.
thinking about it now, I haven't accounted for 'contentObjects'.

moloko:
yeah I see why you need to strip out the 's'; what I don't get is why the `types` array contains both `"component"` and `"components"` when - I think?! - it could just be `var types = [descendants.slice(0, -1)];` and still work fine?

oliverfoster:
Happy to merge?

moloko:
fine by me. @brian-learningpool ?

PR #2182, opened 2018-07-25
Fixed #2178

### LINKED ISSUE #2178: Page body text is not shown unless body text is given as well on a page
### Subject of the issue
Page body text is not shown unless body text is given as well on a page. 

### Your environment
* Framework: 3.x.x
* Browser: Chrome - v59.0
* Ubuntu 14.04

### Steps to reproduce
* Have a page both with a page body and body, see that they both show as expected.
* Have a page with only the page body - it doesn't show.

### Expected behaviour
A page should be able to have a visible page body without a body being set.

### Actual behaviour
Inputting only a page body means it won't display. If you were to want it to display, you would need to also input something into the body text, but this may not be wanted by the user.

https://github.com/adaptlearning/adapt_framework/blob/1679cd9a8ad8a49305d81e90738af9bdf6625d6f/src/core/templates/page.hbs#L19-L29

I believe the issue is this above, just needs the first if removed.

NayanKhedkar:


NayanKhedkar:
[Issue #2178](https://github.com/adaptlearning/adapt_framework/issues/2178)

Issue #2185, opened 2018-07-26
Facilitate component-style sub views


oliverfoster:
### Subject of the enhancement
In some places in Adapt we inject DOM elements from one extension into another. This practise isn't possible with frameworks like VueJS and React. We need to allow instantiation of registered 'component style' views in Adapt.

### External examples
VueJS: https://vuejs.org/v2/guide/components-registration.html
React: https://reactjs.org/docs/components-and-props.html#composing-components

### Expected outcome
This should allow us to move to a more appropriate style of injection, ready for future frameworks, without changing too much. 

#### Handlebars
Pass the Adapt model by id.
```html
{{subview name="menu:progress" id=this._id}}
```
Create a model from the data specified.
```html
{{subview name="component:completionIndicator" model=this}}
```
#### JavaScript
```js
var MenuProgress = Backbone.View.extend({
    initialize: function() {

    }
});
Adapt.subviews.register("menu:progress", MenuProgress);
```

Link2Twenty:
When it comes time to pick a framework it may be worth looking here for pros and cons https://custom-elements-everywhere.com/

PR #2186, opened 2018-07-26
issue/#2185 Added subview api

oliverfoster:
### TEST WITH
[adapt-audioSubView](https://github.com/cgkineo/adapt-audioSubView) example extension.

### PR Details
#2185 
* Added javascript subview registration
##### View definition:
```js
define([
  "core/js/adapt",
  "core/js/views/subView"
], function(Adapt, SubView) {

  var SubViewClass = SubView.extend({

    className: "outer",

    init: function() {
      // View has been rendered and initialized
    },

    attach: function() {
      // View has been (re)attached to the DOM
      // Reattach events handlers here as likely removed by Adapt
    }

  }, {
    template: "subviewclassname"
  });

  Adapt.subviews.register("subviewclassname", SubViewClass);

});
```
##### Template definition:
```hbs
<div class="inner">
  {{options.value}}
</div>
```
* Added handlebars helper for subview instantiation
##### Invocation definition
```hbs
{{subview name="subviewclassname" id="model-id"}}
{{subview name="subviewclassname" id="model-id" value="pass through"}}
{{subview name="subviewclassname" model={_id: "model-id"} }}
{{subview name="subviewclassname"}}
```
* Defined the lifecycle
1. Instantiated when needed from the handlebars.
2. Reused in a single `Adapt.location` window if required. 
3. They will be destroyed on global `remove` events.
4. Require an `_id` of a model available through the `Adapt.findById` function.

```js
Adapt.on("subView:preRender subView:render subView:postRender", function(view) {

});
Adapt.on("subView:attach subView:attached", function(view) {

});
Adapt.on("subView:remove subView:removed", function(view) {

});
var SubViewClass = Adapt.subviews.get("subviewclassname");
```

* Added `using` helper
```hbs
{{#using 'c-60'}}
  Output c-60 title: {{title}}
{{/using}}
```


src/core/js/subviews.js
```diff
@@ -0,0 +1,107 @@
...
+                subtree: true
+            });
+        },
+
+        /**
+         * Check newly added elements for subviews references
+         * @param  {MutationList} list List of browser mutations
+         */
+        _onMutated: function(list) {
+            var $placeholders = $([]);
```
tomgreenfield: Can use just `$()` since jQuery v1.4.

src/core/js/subviews.js
```diff
@@ -0,0 +1,107 @@
+define([
+    'handlebars',
+    'core/js/adapt'
+], function(Handlebars, Adapt){
+
+    var Subviews = Backbone.Controller.extend({
+
+        initialize: function() {
+            this._register = {};
+            this._hold = {};
```
tomgreenfield: Could we have these as named properties on the class, above `initialize()`?

oliverfoster:
Happy this works as described? @brian-learningpool

oliverfoster:
Please don't merge this. There are some life cycle problems I need to sort out. Leaving open for the moment for visibility.

oliverfoster:
@tomgreenfield 
Use cases:
* Audio/help/resources to component titles
* Video/audio/resources to feedback
* Progress bars to menus
* Anywhere we need to inject behaviour or where the current Adapt API is too rigid or tightly coupled to easily create the required behaviour

oliverfoster:
This is ready for scrutiny again.

src/core/js/views/subView.js
```diff
@@ -0,0 +1,80 @@
...
+        },
+
+        /**
+         * Return string representation of parent element for use as
+         * a placeholder in template rendering.
+         * @return {String} Placeholder string
+         */
+        getOuterHTML: function() {
+            var attributes = [];
+            for (var i = 0, l = this.el.attributes.length; i < l; i++) {
```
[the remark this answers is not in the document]
oliverfoster: `this.el.attributes` is a `NamedNodeMap` not an `Array`.  The `map` function isn't implemented on the `NamedNodeMap` prototype in IE11.

src/core/js/subviews.js
```diff
@@ -0,0 +1,330 @@
...
+            });
+        },
+
+        /**
+         * Destroy all invocations and associated view instances
+         */
+        destroy: function() {
+            // List length is liable to change during remove procedure
+            // perform in reverse order.
+            for (var i = this.items.length-1; i > -1; i--) {
```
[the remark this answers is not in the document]
oliverfoster: The list must be performed in reverse order as the items are removed from the list as it progresses.

oliverfoster:
Ta @moloko 
Did you test it out?

src/core/js/views/subView.js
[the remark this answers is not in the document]
moloko: ah fair point... and of course we can't use `Array.from` to convert it either. curse you, IE11!

src/core/js/subviews.js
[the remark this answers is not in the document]
moloko: > I'm not sure how forEach handles a changing length

Good point, I assumed it would be able to handle it. If we're unsure let's leave as-is, it's no big thing.

src/core/js/subviews.js
```diff
@@ -0,0 +1,319 @@
...
+         */
+        isEqual: function(invocation) {
+            if (this.name !== invocation.name) return false;
+            if (this.id !== invocation.id) return false;
+            if (this.model !== invocation.model) return false;
+            return true;
+        },
+
+        /**
+         * Returns boolean true if this invocation has instantiated a subview
```
moloko: this is redundant now

src/core/js/subviews.js
```diff
@@ -0,0 +1,329 @@
...
+    /**
+     * Represents the handlebars subview statements from the current Adapt location.
+     */
+    var Invocations = Backbone.Controller.extend({
+
+        items: null,
+
+        initialize: function() {
+            this.items = [];
+            this.updatePlaceholders = _.debounce(this.updatePlaceholders, 1);
```
cahirodoherty-learningpool: Isn't the wait of a debounce in milliseconds?

oliverfoster: Yup. This just creates a function which when called multiple times synchronously will callback once on the event queue almost immediately after.

Issue #218, opened 2014-03-17
Unclear Drawer functionality


mgancarzdsi:
This is more on the UI design, but testing the Drawer functionality in the community pre-release testing course, I found when I opened the drawer the following happened to the main course window. Some of this might be intentional but I found it inconsitent
- The 'Back' button in the top left slide out of view, preventing me from navigating the rest of the course
- I could no longer scroll the main window while the drawer was open
- I could interact with hot graphic and question type components, ie: even though the main window seemed to meant to be frozen, I could still interact with clickable components on it, including the progress bar
- The Drawer itself also had a horizontal scroll bar in it's bottom, once I clicked on the 'Resources' button within it.
- Clicking the progress bar while the drawer was open, opened the course progress menu, but clicking the drawer button again did not re-open the resources drawer, instead it closed the drawer. 
- Clicking the progress bar after opening it, does not close it again like the drawer.

If you would like me to split these out into seperate issues, I can do so, but thought I would lump it all together since it's all related to drawer functionality.

Tested on Windows 7 64-bit, Chrome 33.0.1750

PR #2211, opened 2018-08-09
Fix toggleBtn not closing the drawer when only 1 drawer item

nilslp:
Setting `this._isCustomViewVisible = false;` after that creates this UX:

- A user can click `toggleDrawer` it will open and it will show the drawerItems
- A user can click `progressBar` when drawer is open and it will show progressBar stats in the drawer
- A user can then click `toggleDrawer` again and it will show the drawerItems again
- A further click on `toggleDrawer` will hide the drawer

This should be in keeping with the link adaptlearning/adapt_framework#218 @tomgreenfield mentioned but also hiding the drawer using the `toggleBtn`

### DIFF
--- a/src/core/js/views/drawerView.js
+++ b/src/core/js/views/drawerView.js
@@ -139,9 +139,13 @@ define([
                 this.$('.drawer-back').addClass('display-none');
                 this._isCustomViewVisible = false;
                 this.emptyDrawer();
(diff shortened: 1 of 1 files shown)


nilslp:
Thanks @moloko I will update the comment, I should also create a PR for `master`, once your happy with this PR, as it has the same issue.

PR #2225, opened 2018-09-17
Issue/2209-legacy - Fix toggleBtn not closing the drawer when only 1 drawer item

nilslp:
Setting this._isCustomViewVisible = false; after that creates this UX:

A user can click toggleDrawer it will open and it will show the drawerItems
A user can click progressBar when drawer is open and it will show progressBar stats in the drawer
A user can then click toggleDrawer again and it will show the drawerItems again
A further click on toggleDrawer will hide the drawer
This should be in keeping with the link adaptlearning/adapt_framework#218 @tomgreenfield mentioned but also hiding the drawer using the toggleBtn

This was originally here: https://github.com/adaptlearning/adapt_framework/pull/2211

Not sure what happened or why it was closed (or if it was me who closed it) but regardless here is the new one.

### DIFF
--- a/src/core/js/views/drawerView.js
+++ b/src/core/js/views/drawerView.js
@@ -139,9 +139,13 @@ define([
                 this.$('.drawer-back').addClass('display-none');
                 this._isCustomViewVisible = false;
                 this.emptyDrawer();
-                if(this.collection.models.length === 1) {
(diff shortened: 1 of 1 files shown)


brian-learningpool:
@moloko, I think the original PR was closed unintentionally.

PR #2244, opened 2018-10-18
re-write of findAncestor

### LINKED ISSUE #2240: adaptModel.findAncestor triggers an error if no ancestor of that type exists
If you do a [findAncestor](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/models/adaptModel.js#L242-L259) search on a contentObject to see if it has a contentObject as its ancestor - and it doesn't - an error is thrown.

(Kineo's [pageIncompletePrompt extension](https://github.com/cgkineo/adapt-pageIncompletePrompt) does this to check to see if learner is being routed to a child page of the current page)

### Your environment
Any

### Steps to reproduce
Execute the following in the console:
```js
require('core/js/adapt').findById('co-05').findAncestor('contentObjects');
```

### Expected behaviour
no error should occur, the function should simply return `undefined` to indicate that no results were found

### Actual behaviour
The error 'TypeError: returnedAncestor is undefined' is thrown.

moloko:
to fix #2240 

also remove call to `setOptional` from `setupModel` - it wasn't actually doing anything and we wouldn't want to be forcing children to be optional anyway

general tidy-up to:
* change `_.bind` to `Function.bind`
* improve comments
* use best version of `model.set` as appropriate
* remove unused variables

### DIFF
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -42,9 +42,6 @@ define([
                 this._parent = 'contentObjects';
             }
             if (this._children) {
-                //if parent is optional, apply to children
-                if (this.get('_isOptional')) this.setOptional(true);
-
                 this.setupChildListeners();
             }
 
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -240,7 +240,7 @@ define([
 
         /**
          * Searches the model's ancestors to find the first instance of the specified ancestor type
-         * @param {string} [ancestorType] Valid values are 'course', contentObjects', 'articles' or 'blocks'.
+         * @param {string} [ancestorType] Valid values are 'course', 'contentObjects', 'articles' or 'blocks'.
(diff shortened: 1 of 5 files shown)


src/core/js/models/adaptModel.js
```diff
@@ -239,29 +238,28 @@ define([
...
+         * @param {string} [ancestorType] Valid values are 'course', contentObjects', 'articles' or 'blocks'.
+         * If left blank, the immediate ancestor (if there is one) is returned
+         * @return {object} Reference to the model of the first ancestor of the specified type that's found - or `undefined` if none found
+         */
+        findAncestor: function (ancestorType) {
             var parent = this.getParent();
+            if (!parent) return;
 
-            if (this._parent === ancestors) {
+            if (!ancestorType || ancestorType === "" || this._parent === ancestorType) {
```
tomgreenfield: Redundant logic – empty strings are already caught by `!ancestorType`.

Issue #2260, opened 2018-11-05
Adapt.course.findDescendantModels submenu issues


danielstorey:
I have been using `Adapt.course.findDescendantModels('components')` to return all components in a course in the correct order, which is particularly important for courses created in the Authoring Tool.

It would appear that components within a submenu page are omitted.

oliverfoster:
This is probably because the assessment changes the children array.
Could you confirm whether assessment children are missing?

moloko:
@danielstorey could you state which version of the framework you're using? findDescendantModels has been through quite a few changes recently so it's important to know which version you're having problems with

danielstorey:
I thought I was on the latest version but was on a slightly older dev course (3.2.0). Working fine in  latest version. Sorry for the false alarm

danielstorey:
Any chance these fixes could be applied to V2? I have a client who is currently stuck on V2

moloko:
@danielstorey quite possibly, will have a look shortly

moloko:
looking through the code I think it should work OK, will re-open and add a pull request to allow others to review

moloko:
done in [v2.3.0](https://github.com/adaptlearning/adapt_framework/releases/tag/v2.3.0)

PR #2265, opened 2018-11-06
update findDescendantModels (and add getAllDescendantModels)

moloko:
to fix issue #2260


moloko:
(just to add I tested in IE8, IE11 and FF)

Issue #2269, opened 2018-11-13
move inview logic into ComponentView


moloko:
Since the assessmentResults, blank, text, graphic, hotgraphic, narrative & media components all implement `inview` it seems like a sensible idea to move as much of the logic as possible into ComponentView. And now seems like quite a good time to do it!

The implementation in adapt-contrib-text seems like a pretty good starting point - particularly in terms of how it allows some flexibility as to what the 'inview element' should be.

I think we would need three functions:
1. ~~`setupInviewListener`~~`setupInviewCompletion`
1. `onInview`
1. `removeInviewListener`

Plus a model property `inviewElementSelector`

If that sounds OK by everyone I'm happy to pick this up.

oliverfoster:
Yup, perfect.

brian-learningpool:
@moloko, is this something which is intended to go into v4?

moloko:
@brian-learningpool yep that's the idea

PR #2271, opened 2018-11-14
moved inview logic into ComponentView

moloko:
#2269 
allowing both the element to be used for inview and the callback function to be specified, so as to allow for plenty of flexibility

### DIFF
--- a/src/core/js/views/componentView.js
+++ b/src/core/js/views/componentView.js
@@ -23,6 +23,47 @@ define([
             this.renderState();
         },
 
+        /**
+         * Allows components that want to use inview for completion to set that up
+         * @param {string} [inviewElementSelector] Allows to you to specify (via a selector) which DOM element to use for inview.
+         * Defaults to `'.component-inner'` if not supplied.
+         * @param {function} [callback] Allows you to specify what function is called when the component has been viewed.
+         * Defaults to `view.setCompletionStatus` if not specified
+         */
+        setupInviewListener: function(inviewElementSelector, callback) {
+            this.$inviewElement = this.$(inviewElementSelector || '.component-inner');
+            this.inviewCallback = (callback || this.setCompletionStatus);
+
+            this.$inviewElement.on('inview.componentView', this.onInview.bind(this));
+        },
+
+        removeInviewListener: function() {
+            this.$inviewElement.off('inview.componentView');
+        },
+
+        onInview: function(event, visible, visiblePartX, visiblePartY) {
+            if (!visible) return;
+
+            switch (visiblePartY) {
+				case 'top':
+					this.hasSeenTop = true;
+					break;
+				case 'bottom':
+					this.hasSeenBottom = true;
+					break;
+				case 'both':
+					this.hasSeenTop = this.hasSeenBottom = true;
+			}
+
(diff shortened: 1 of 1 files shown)


src/core/js/views/componentView.js
```diff
@@ -23,17 +23,58 @@ define([
...
+            switch (visiblePartY) {
+				case 'top':
+					this.hasSeenTop = true;
+					break;
+				case 'bottom':
+					this.hasSeenBottom = true;
+					break;
+				case 'both':
+					this.hasSeenTop = this.hasSeenBottom = true;
+			}
```
brian-learningpool: Indentation has gone a bit funky here.

moloko: that's what I get for copy-pasting some of @tomgreenfield 's code ;-)

src/core/js/views/componentView.js
```diff
@@ -6,15 +6,15 @@ define([
...
-            return "component " + 
-            this.model.get('_component') + 
-            "-component " + this.model.get('_id') + 
-            " " + this.model.get('_classes') + 
-            " " + this.setVisibility() +
-            " " + this.setHidden() +
-            " component-" + this.model.get('_layout') + 
-            " nth-child-" + this.model.get("_nthChild") +
-            " " + (this.model.get('_isComplete') ? 'completed' : '');
+            return 'component ' +
```
oliverfoster:
```suggestion
            return [ 
                'component',
                this.model.get('_component') +'-component',
                this.model.get('_id'),
                this.model.get('_classes'),
                this.setVisibility(),
                this.setHidden(),
                'component-' + this.model.get('_layout'),
                'nth-child-' + this.model.get('_nthChild'),
                (this.model.get('_isComplete') ? 'completed' : '')
            ].join(" ");
```

src/core/js/views/componentView.js
```diff
@@ -6,15 +6,17 @@ define([
...
-            " " + (this.model.get('_isComplete') ? 'completed' : '');
+            return [
+                'component',
+                this.model.get('_component') +'-component',
+                this.model.get('_id'),
+                this.model.get('_classes'),
+                this.setVisibility(),
+                this.setHidden(),
+                'component-' + this.model.get('_layout'),
+                'nth-child-' + this.model.get('_nthChild'),
```
lc-thomasberger: I think we can start to deprecate the custom nth-child classes in favor to native css selectors.

src/core/js/views/componentView.js
```diff
@@ -23,17 +25,59 @@ define([
...
+         * for an example. Defaults to `view.setCompletionStatus` if not specified.
+         */
+        setupInviewCompletion: function(inviewElementSelector, callback) {
+            this.$inviewElement = this.$(inviewElementSelector || '.component-inner');
+            this.inviewCallback = (callback || this.setCompletionStatus);
+
+            this.$inviewElement.on('inview.componentView', this.onInview.bind(this));
+        },
+
+        removeInviewListener: function() {
```
NayanKhedkar: I think there might be need to be add some check.
```
removeInviewListener: function() {
      if (!this.$inviewElement) return;
      this.$inviewElement.off('inview.componentView');
 }
```

PR #1684, opened 2017-07-24
issue/1683+issue/1678: grunt allow .txt/.json, include build model in output

oliverfoster:
[issue/1678](https://github.com/adaptlearning/adapt_framework/issues/1678):
* allow ``course/**/*.json`` files to be ``course/**/*.txt`` or any other extension in both grunt and the framework.
to test:
1. rename ``config.json``, ``en/course.json``, ``en/contentObject.json``, ``en/articles.json``, ``en/blocks.json`` and ``en/components.json`` to ``config.txt``, ``en/course.txt``, ``en/contentObject.txt``, ``en/articles.txt``, ``en/blocks.txt`` and ``en/components.txt``. 
2. run ```$ grunt dev --jsonext=txt```

[issue/1683](https://github.com/adaptlearning/adapt_framework/issues/1683):
* include build commands, config, plugin bower.json and package json in build output
1. run ```grunt dev``` or ```grunt build```
2. when adapt loads, run 
```
var Adapt = require("core/js/adapt");
Adapt.build.toJSON();
```
<details>
<summary>You should have an object which looks something like:</summary>

```json
{
  "jsonext": "txt",
  "theme": "**",
  "menu": "**",
  "languages": "**",
  "package": {
    "name": "adapt_framework",
    "version": "2.1.3",
    "description": "Adapt Learning output framework",
    "repository": {
      "type": "git",
      "url": "git://github.com/adaptlearning/adapt_framework.git"
    },
    "license": "GPL-3.0",
    "bugs": {
      "url": "https://github.com/adaptlearning/adapt_framework/issues"
    },
    "contributors": [
      {
        "name": "Alan Bourne",
        "email": "alan@zenduo.co.uk"
      },
      {
        "name": "Aniket Dharia",
        "email": "aniket.dharia@vyassystems.com"
      },
      {
        "name": "Belal Almassri",
        "email": "belal.almassri@spongeuk.com"
      },
      {
        "name": "Brian Quinn",
        "email": "brian@learningpool.com"
      },
      {
        "name": "Chris Jones",
        "email": "chris.jones@spongeuk.com"
      },
      {
        "name": "Chris Steele",
        "email": "chris.steele@kineo.com"
      },
      {
        "name": "Chuck Lorenz",
        "email": "chucklorenz@yahoo.com"
      },
      {
        "name": "Dan Ghost",
        "email": "daniel.ghost@kineo.com"
      },
      {
        "name": "Dan Gray",
        "email": "dan@sinensis.co.uk"
      },
[...]

PR #2290, opened 2018-12-05
Back-port #1684 to legacy

taylortom:
...to allow course content files to be non-json file-types.

### DIFF
--- a/grunt/config/build-config.js
+++ b/grunt/config/build-config.js
@@ -1,0 +1,39 @@
+module.exports = function (grunt, options) {
+    return {
+        options:{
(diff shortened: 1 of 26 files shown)


moloko:
@taylortom have you tested all this? or does that still need doing?

taylortom:
@moloko I've tested for my use-case, but could do with some other testing. There's no rush to get this in, and it's currently being user tested in one of our live installs, so may be worth leaving it open for a bit to see if any issues are highlighted.

cahirodoherty-learningpool (approved):
👀

Issue #1180, opened 2016-07-28
Create "mcq" question component model


oliverfoster:

oliverfoster:
this model is done already, but it still lives in the MCQ

we should move it into core

PR #2299, opened 2018-12-17
 issue/1180: Added ItemsQuestionModel and isInteractive

oliverfoster:
#1180
Moved and refactored McqModel into the core as a base for all multiple choice components.
Based upon QuestionModel and ItemsComponentModel.

### DIFF
--- a/src/core/js/models/itemsQuestionModel.js
+++ b/src/core/js/models/itemsQuestionModel.js
@@ -57,37 +57,41 @@ define([
 
         isCorrect: function() {
 
-            var numberOfRequiredAnswers = 0;
-            var numberOfCorrectAnswers = 0;
-            var numberOfIncorrectAnswers = 0;
+            var props = {
+                _numberOfRequiredAnswers: 0,
+                _numberOfIncorrectAnswers: 0,
+                _isAtLeastOneCorrectSelection: false,
+                _numberOfCorrectAnswers: 0,
+                _isCorrect: false
+            };
 
             this.getChildren().each(function(itemModel) {
                 var itemShouldBeActive = itemModel.get('_shouldBeSelected');
                 if (itemShouldBeActive) {
-                    numberOfRequiredAnswers++;
+                    props._numberOfRequiredAnswers++;
                 }
 
-                var itemActive = itemModel.get('_isActive');
-                if (!itemActive) return;
+                if (!itemModel.get('_isActive')) return;
 
-                if (itemShouldBeActive) {
-                    numberOfCorrectAnswers++;
-                    itemModel.set('_isCorrect', true);
+                if (!itemShouldBeActive) {
+                    props._numberOfIncorrectAnswers++;
                     return;
                 }
 
-                numberOfIncorrectAnswers++;
+                props._isAtLeastOneCorrectSelection = true;
+                props._numberOfCorrectAnswers++;
+                itemModel.set('_isCorrect', true);
             });
 
-            this.set({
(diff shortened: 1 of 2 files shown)


src/core/js/models/itemsQuestionModel.js
```diff
@@ -0,0 +1,219 @@
...
+        // This should preserve the state of the users answers
+        storeUserAnswer: function() {
+            var userAnswer = [];
+
+            var items = this.getChildren().slice(0);
+            items.sort(function(a, b) {
+                return a.get('_index') - b.get('_index');
+            });
+
+            _.each(items, function(itemModel, index) {
```
moloko: change to `items.forEach`? either way, the `this` argument is unnecessary.

src/core/js/models/questionModel.js
```diff
@@ -240,11 +240,15 @@ define([
                 feedbackMessage: body
             });
         },
-        
+
         getFeedbackTitle: function() {
             return this.get('_feedback').title || this.get('displayTitle') ||  this.get('title') || "";
         },
 
+        isInteractive: function() {
```
moloko: I think this would benefit from a comment along the lines of 'used to determine whether the learner is allowed to interact with the question component or not'.

src/core/js/models/itemsQuestionModel.js
```diff
@@ -0,0 +1,217 @@
...
+        },
+
+        getLastActiveItem: function(){
+            var selectedItems = this.getActiveItems();
+            return selectedItems[selectedItems.length-1];
+        },
+
+        resetItems: function() {
+            this.resetActiveItems();
+            this.set({
```
tomgreenfield: Suggest
```js
this.set('_isAtLeastOneCorrectSelection', false);
```

src/core/js/models/itemsQuestionModel.js
```diff
@@ -0,0 +1,217 @@
...
+                    description: itemModel.get('text')
+                };
+            });
+
+            var correctItems = this.getChildren().filter(function(itemModel) {
+                return itemModel.get('_shouldBeSelected');
+            });
+
+            interactions.correctResponsesPattern = [
+                _.map(correctItems, function(itemModel) {
```
tomgreenfield: Is it possible to use native `.map()` here?

src/core/js/models/itemsQuestionModel.js
```diff
@@ -0,0 +1,217 @@
...
+            this.set('_userAnswer', userAnswer);
+        },
+
+        isCorrect: function() {
+
+            var numberOfRequiredAnswers = 0;
+            var numberOfCorrectAnswers = 0;
+            var numberOfIncorrectAnswers = 0;
+
+            this.getChildren().each(function(itemModel, index) {
```
tomgreenfield: Unused `index`.

src/core/js/models/itemsQuestionModel.js
oliverfoster:
```
var correctItems = this.getChildren().filter
```
[Array.prototype.filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) Always returns an Array. No need for typing here ;)

oliverfoster: Fair play. [https://underscorejs.org/#filter]
Most of those underscore functions on the Collection return their normal stuffs but use the Collection.models as the subject argument of the function instead of an Array.

oliverfoster:
Ready to go now 👍

Issue #2314, opened 2019-01-07
Provide propagation mechanism for completion.


oliverfoster:
### Subject of the issue
Cannot easily listen to completion changes without listening to all collections or models.

### Expected behaviour
There should be a mechanism to bubble events from component to course.

Proposal:
```js
// core/js/modelEvent.js
define(function() {

  var ModelEvent = Backbone.Controller.extend({

    type: null,
    target: null,
    value: null,
    bubbles: true,
    deepPath: null,
    target: null,
    timeStamp: null,

    initialize: function(type, target, value) {
      this.type = type;
      this.target = target;
      this.value = value;
      this.deepPath = [target];
    },

    stopPropagation: function() {
      this.bubbles = false;
    },

    addPath: function(target) {
      this.deepPath.unshift(target);
    }

  });

  return ModelEvent;

});

// core/js/model/adaptModel.js
define([
    'core/js/adapt',
    'core/js/modelEvent',
], function (Adapt, ModelEvent) {

    var AdaptModel = Backbone.Model.extend({

        bubblingEvents: [
            'change:_isComplete',
            'change:_isInteractionComplete'
        ],

        setupChildListeners: function() {
            this.listenTo(children, {
                "all": this.onAll,
                "bubble": this.bubble
            });
        }

        onAll: function(type, model, value) {
            if (!_.contains(this.bubblingEvents, type)) return;
            var event = new ModelEvent(type, model, value);
            this.bubble(event);
        },

        bubble: function(event) {
            if (!event.bubbles) return;
            event.addPath(this);
            this.trigger("bubble:"+event.type, event);
            this.trigger("bubble", event);
        }

    });

    return AdaptModel;

});


// example
Adapt.course.on("bubble:change:_isComplete", function(event) {
  /* 
   * event.type = "change:_isComplete";
   * event.stopPropagation(); // stop event from bubbling
   * event.bubbles = true/false; // has stopPropagation been called?
   * event.deepPath = [grandParent, parent, target]; // path of bubble
   * event.target; // instance on which event occurred
   * event.value; // value of event
   *
   */
});
```

PR #2315, opened 2019-01-07
issue/2314: Added model event bubbling mechanism

oliverfoster:
#2314 
* Allow change:_isComplete and change:_isInteractionComplete events to bubble from component to course

```js
Adapt.course.on("bubble:change:_isComplete", function(event) {
  /* 
   * event.type = "change:_isComplete";
   * event.stopPropagation(); // stop event from bubbling
   * event.canBubble = true/false; // has stopPropagation been called?
   * event.deepPath = [grandParent, parent, target]; // path of bubble
   * event.target; // instance on which event occurred
   * event.value; // value of event
   *
   */
});

### CODE REVIEW

File: src/core/js/modelEvent.js [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+
+        type: null,
+        target: null,
+        value: null,
+        bubbles: true,
+        deepPath: null,
+        target: null,
+        timeStamp: null,
+
+        initialize: function(type, target, value) {
```
moloko: a JSDoc comment would be useful here to document the function parameter data types at least

src/core/js/models/adaptModel.js
```diff
@@ -739,8 +747,20 @@ define([
 
         onIsComplete: function() {
             this.checkCompletionStatus();
-
             this.checkLocking();
+        },
+
+        onAll: function(type, model, value) {
```
moloko: a JSDoc comment would be useful here to document the function parameter data types at least

src/core/js/models/adaptModel.js
```diff
@@ -739,8 +747,20 @@ define([
...
+            if (!_.contains(this.bubblingEvents, type)) return;
+            var event = new ModelEvent(type, model, value);
+            this.bubble(event);
+        },
+
+        bubble: function(event) {
+            if (!event.bubbles) return;
+            event.addPath(this);
+            this.trigger("bubble:"+event.type, event);
+            this.trigger("bubble", event);
```
moloko: could maybe merge these two lines into one 
```js
this.trigger('bubble bubble:' + event.type, event);
```

src/core/js/modelEvent.js
```diff
@@ -0,0 +1,32 @@
+define(function() {
+
+    var ModelEvent = Backbone.Controller.extend({
+
+        type: null,
+        target: null,
+        value: null,
+        bubbles: true,
```
brian-learningpool: Minor thing, but since this is a Boolean could we possibly rename this to something like `canBubble` for readability?  To me `bubbles` sounds like it _could_ be an array.

moloko: isn’t it named that to match https://developer.mozilla.org/en-US/docs/Web/API/Event/bubbles ?

src/core/js/models/adaptModel.js
oliverfoster: I want `bubble` to call after the more specific `bubble:type`, so imma invert your string but yes.

oliverfoster:
Good to be reviewed @moloko & @brian-learningpool

Issue #1921, opened 2018-01-16
index.html always assumes 'lang' to be English


brian-learningpool:
### Subject of the issue/enhancement/features
The index.html file always assumes the language is English (see https://github.com/adaptlearning/adapt_framework/blob/master/src/index.html#L2)

### Your environment
AT and Framework
Any browser/device/OS

### Steps to reproduce

1. Build and view any course with a language other than English.
2. View the source of index.html.

### Expected behaviour

The default language on index.html's `html` attribute should be set by the `_defaultLanguage` attribute in `config.json`.  Also, this should change dynamically if the language is changed mid attempt via the language picker.

### Actual behaviour

The `lang` attribute on `html` is always set to `en`.

moloko:
just tested using Adapt v2.2.3 and the `lang` attribute does update if you're using the languagePicker extension...

it also seemed to work fine if I disabled the languagePicker and changed the value of `_defaultLanguage` to `"de"`

moloko:
This is what I get if I change the value of `_defaultLanguage` to `"de"`, build and run the course:
![2018-01-16 14_38_05-](https://user-images.githubusercontent.com/1249597/34994336-0ea8a1a2-facb-11e7-816c-acf3e2577549.png)

brian-learningpool:
Thanks @moloko.  The DOM is indeed correctly updated as you have demonstrated.  This issue was raised by a customer who is pointing automated accessibility testing software at index.html.  On disk the index.html is always `en`, but when you launch the course and the JavaScript runs it's as expected.

moloko:
is there something that still needs addressing or can this now be closed?

brian-learningpool:
We could run the replace task to update the `lang` attribute in index.html.

Can we leave this ticket open for a little while so I can see how much of a problem this is actually causing?

moloko:
> We could run the replace task to update the `lang` attribute in index.html.

Now that #2317 is done this should be quite straightforward to do if this is something you feel still needs doing?

moloko:
closing due to lack of activity. feel free to reopen if you feel this needs doing.

PR #2317, opened 2019-01-10
Add ability to insert variables into HTML files

Floppy:
In our theme, we want to add some course data into meta tags, etc, in a custom `required/index.html` file. 

Currently, in `required`, the XML files are processed so you can do things like this from https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/required/imsmanifest.xml#L9

```
 <langstring xml:lang="x-none"><![CDATA[@@course.title]]></langstring>
```

if we add `*.html` to the grunt replace job as well as `*.xml`, we should be able to do the same thing in index.html, meaning that themes can then automatically pull in data from the course. For instance:

```
 <title><![CDATA[@@course.title]]></title>
```

This is my first venture into the internals of the Adapt framework, so if there is a way I should be testing these jobs, or otherwise improving my PR, please tell me!

### DIFF
--- a/grunt/config/replace.js
+++ b/grunt/config/replace.js
@@ -89,7 +89,10 @@ module.exports = function (grunt, options) {
         {
           expand: true,
           flatten: true,
-          src: [path.join(options.outputdir, '*.xml')],
+          src: [
+            path.join(options.outputdir, '*.xml'),
+            path.join(options.outputdir, '*.html')
+          ],
           dest: options.outputdir
         }
       ]


moloko:
@Floppy thanks for this. I think this is a useful addition and may well help with issues like https://github.com/adaptlearning/adapt_framework/issues/1921

There's a guide to contributing to adapt [here](https://github.com/adaptlearning/adapt_framework/wiki/Contributing-code#submitting-your-code) that's worth a quick read - though really you're not far off, I think the only missing piece is that we normally require a matching Github issue for each PR. It's often worth logging this before you start the work to give people an opportunity to comment on what you'd like to do before you start work - this can help avoid re-work or you doing a lot of work submitting something that actually we don't think should be included.

moloko:
@Floppy no, it's all good for this instance, particularly as you have given a comprehensive description of what is is you want to achieve and why

Floppy:
Fantastic, thanks. When this is released I'll do a PR into the vanilla theme to add the course title into the header `title` tag, show how it's used :)

Floppy:
Having checked the contributor guide first of course 😉

PR #2332, opened 2019-01-22
Resolves #2247, adds a shim to support custom manifest 'identifier'

### LINKED ISSUE #2247: adapt-contrib-spoor: Add ability to dynamically set the identifier in manifest.xml
In imsmanifest.xml the identifier is currently set to 'adapt_manifest'.  Some LMSs require that this is a unique value.  I'm proposing that we add a new property to Spoor to allow the user to specify this, with a default value of adapt_manifest in line with current behaviour.

Essentially this leaves it up to the user to configure (if required) but it means that editing the XML file by hand is no longer required.

### Your environment
Adapt Framework v3.2.1

### Steps to reproduce
Create any Adapt course in the framework or authoring tool.

### Expected behaviour
The user should have the option to specify the manifest identifier.

### Actual behaviour
The identifier is hard-coded to 'adapt_manifest'.

brian-learningpool:
This ensures 'adapt_scorm' will be used as a default if the corresponding property is missing.

### DIFF
--- a/grunt/config/replace.js
+++ b/grunt/config/replace.js
@@ -70,7 +70,9 @@ module.exports = function (grunt, options) {
             configJson._spoor._advancedSettings._manifestIdentifier = 'adapt_scorm';
           }
         } else {
-          configJson._spoor._advancedSettings._manifestIdentifier = 'adapt_scorm';
+          configJson._spoor._advancedSettings = { 
+            _manifestIdentifier: 'adapt_scorm'
+          };
         }
       }
 
--- a/grunt/config/replace.js
+++ b/grunt/config/replace.js
@@ -65,15 +65,8 @@ module.exports = function (grunt, options) {
 
       // Shim to preserve the 'adapt_scorm' identifier.
       if (configJson.hasOwnProperty('_spoor')) {
-        if (configJson._spoor._advancedSettings) {
-          if (!configJson._spoor._advancedSettings._manifestIdentifier) {
-            configJson._spoor._advancedSettings._manifestIdentifier = 'adapt_scorm';
-          }
-        } else {
-          configJson._spoor._advancedSettings = { 
-            _manifestIdentifier: 'adapt_scorm'
-          };
-        }
+        configJson._spoor._advancedSettings = configJson._spoor._advancedSettings || {};
+        configJson._spoor._advancedSettings._manifestIdentifier = spoor._advancedSettings._manifestIdentifier || 'adapt_scorm';
       }
 
       // Combine the course and config JSON so both can be passed to replace.  
--- a/grunt/config/replace.js
+++ b/grunt/config/replace.js
@@ -62,7 +62,18 @@ module.exports = function (grunt, options) {
           grunt.log.writeln('WARNING: xAPI activityID has not been set');
         }
       }
-      
+
+      // Shim to preserve the 'adapt_scorm' identifier.
+      if (configJson.hasOwnProperty('_spoor')) {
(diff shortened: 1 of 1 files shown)


grunt/config/replace.js
```diff
@@ -62,7 +62,18 @@ module.exports = function (grunt, options) {
           grunt.log.writeln('WARNING: xAPI activityID has not been set');
         }
       }
-      
+
+      // Shim to preserve the 'adapt_scorm' identifier.
+      if (configJson.hasOwnProperty('_spoor')) {
+        if (comfigJson._spoor._advancedSettings) {
```
moloko: typo 'comfig' here and on all the following lines

brian-learningpool: How embarrassing.

grunt/config/replace.js
```diff
@@ -62,7 +62,18 @@ module.exports = function (grunt, options) {
...
-      
+
+      // Shim to preserve the 'adapt_scorm' identifier.
+      if (configJson.hasOwnProperty('_spoor')) {
+        if (configJson._spoor._advancedSettings) {
+          if (!configJson._spoor._advancedSettings._manifestIdentifier) {
+            configJson._spoor._advancedSettings._manifestIdentifier = 'adapt_scorm';
+          }
+        } else {
+          configJson._spoor._advancedSettings._manifestIdentifier = 'adapt_scorm';
```
oliverfoster: this line will always throw an error as cannot assign `_manifestIdentifier` to undefined `configJson._spoor._advancedSettings`

oliverfoster: perhaps something like this would keep the if statements to a minimum?
```js
      var spoor = configJson._spoor;
      if (spoor) {
       spoor._advancedSettings = spoor._advancedSettings || {};
        _.defaults(spoor._advancedSettings, {
            _manifestIdentifier: 'adapt_scorm'
        });
      }
```
or:
```js
      var spoor = configJson._spoor;
      if (spoor) {
       spoor._advancedSettings = spoor._advancedSettings || {};
       spoor._advancedSettings._manifestIdentifier = spoor._advancedSettings._manifestIdentifier || 'adapt_scorm';
      }
```

brian-learningpool: That's actually much cleaner.

oliverfoster:
The [hasOwnProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) check always seems like a an overkill when you can use falsies to do the same but with less code. JSON objects don't have an inheritance chain to speak of which is part of the usecase of hasOwnProperty.


value | !!value
-- | --
false | false
0 | false
NaN | false
null | false
undefined | false
“” (empty string) | false
” ” (white space) | true
[] | true
{} | true



The former should suffice as an existence check as if the property is undefined because it doesn't exist then it'll fail all the same:
```js
if (jsonObject.propertyName) {
  // this will return true if the property contains anything that doesn't equate to false
  // it doesn't care if the property has been inherited, which it shouldn't have been because we're dealing with JSON
}
if (jsonObject.hasOwnProperty('propertyName')) {
  // this will return true if the property has ever been declared regardless of the value assigned, it may still be undefined
  // if the property has been inherited it will be ignored here, but one shouldn't have been because we're dealing with JSON
}
```

This works because the default value is only assigned if the jsonObject.propertyName is a falsie, in effect the line is skipped if  jsonObject.propertyName has a none false value:
```js
jsonObject.propertyName = jsonObject.propertyName || defaultValue;
```

PR #2341, opened 2019-01-30
Fixes #2247 - Corrects error in Spoor 'adapt_manifest' replace

### LINKED ISSUE #2247: adapt-contrib-spoor: Add ability to dynamically set the identifier in manifest.xml

brian-learningpool:
This was checking a non-existent object.

### DIFF
--- a/grunt/config/replace.js
+++ b/grunt/config/replace.js
@@ -64,9 +64,10 @@ module.exports = function (grunt, options) {
       }
 
       // Shim to preserve the 'adapt_manifest' identifier.
-      if (configJson.hasOwnProperty('_spoor')) {
-        configJson._spoor._advancedSettings = configJson._spoor._advancedSettings || {};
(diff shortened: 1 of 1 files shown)


brian-learningpool:
I agree.

PR #2346, opened 2019-02-05
Update sharing options

### LINKED ISSUE #2345: Add authoring tool sharing options
We're amending course sharing options for **v0.7.0**, so need to add the relevant attributes to the core schemas. See the below issues for more info
- **Schema update:** [adapt_authoring/issues/2177](https://github.com/adaptlearning/adapt_authoring/issues/2177)
- **Course sharing:** [adapt_authoring/issues/2123](https://github.com/adaptlearning/adapt_authoring/issues/2123)

taylortom:
Fixes #2345.

:skull: Need to discuss exactly when this should be merged, as it will break authoring tools older than v0.7.0 due to the new `Users` input type.

### DIFF
--- a/src/core/schema/course.model.schema
+++ b/src/core/schema/course.model.schema
@@ -523,8 +523,21 @@
       "editorOnly": true,
       "inputType": "Checkbox",
       "validators": [],
-      "title": "Share with other users",
-      "help": "Controls whether or not your colleagues will be able to see this course from the 'Shared courses' option"
+      "title": "Share with all users",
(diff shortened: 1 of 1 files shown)


canstudios-nicolaw:
Apologies because I'm not so familiar with the framework side of things I just wanted to check... @taylortom I think this change will need to be in framework version 2,3 & 4. Is this PR going to achieve that? I think you might need to make a PR to the legacy branch?

canstudios-nicolaw:
Discussed in the authoring tool standup and we think there should also be a PR to the legacy branch to avoid confusion caused by having this in some places and not others

PR #2351, opened 2019-02-12
Update sharing options (legacy)

### LINKED ISSUE #2345: Add authoring tool sharing options

taylortom:
Fixes #2345. Duplicate of #2346.

:skull: Need to discuss exactly when this should be merged, as it will break authoring tools older than v0.7.0 due to the new `Users` input type.

### DIFF
--- a/src/core/schema/course.model.schema
+++ b/src/core/schema/course.model.schema
@@ -584,8 +584,21 @@
       "editorOnly": true,
       "inputType": "Checkbox",
       "validators": [],
-      "title": "Share with other users",
-      "help": "Controls whether or not your colleagues will be able to see this course from the 'Shared courses' option"
+      "title": "Share with all users",
+      "help": "Controls whether or not all colleagues will be able to see this course from the 'Shared courses' option"
+    },
+    "_shareWithUsers": {
+      "type": "array",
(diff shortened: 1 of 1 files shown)


src/core/schema/course.model.schema
```diff
@@ -584,8 +584,21 @@
...
+      "inputType": "Users",
+      "items": {
+        "type": "objectid",
+        "inputType": "Text",
+        "required": false,
+        "editorOnly": true,
+        "ref": "user"
+      },
+      "title": "Share with specific users",
+      "help": "Specifies which colleagues will be able to see this course from the 'Shared courses' option"
```
tomgreenfield: * Suggest sticking to `users` instead of `colleagues`.
* Suggest we reword this so it doesn't specifically reference `'Shared courses'` since the terminology/filtering may change in the near future.

Issue #2352, opened 2019-02-12
Provide configuration option for _allowScrollOver


chris-steele:

moloko:
done in fc3e6737dc8a415d5031183b07276c090b0c89cc

PR #2353, opened 2019-02-12
Made jquery.onscreen (inview) _allowScrollOver configurable

chris-steele:
#2352

### DIFF
--- a/src/core/schema/config.model.schema
+++ b/src/core/schema/config.model.schema
@@ -294,7 +294,7 @@
     },
     "_inview": {
       "type": "object",
-      "title": "jquery.onscreen (inview)",
+      "title": "",
       "properties": {
       "_allowScrollOver": {
         "type": "boolean",
--- a/src/core/schema/config.model.schema
+++ b/src/core/schema/config.model.schema
@@ -291,6 +291,19 @@
     },
     "themeSettings": {
       "type": "object"
+    },
+    "_inview": {
+      "type": "object",
+      "title": "jquery.onscreen (inview)",
+      "properties": {
+      "_allowScrollOver": {
+        "type": "boolean",
+        "default": true,
+        "title": "Allow scroll over",
+        "inputType": "Checkbox",
+        "help": "Some browsers pause script execution during scroll (e.g. IE8). If this option is selected the inview library will determine if elements became visible during the scroll event."
+        }
+      }
     }
   }
 }
--- a/src/core/schema/config.model.schema
+++ b/src/core/schema/config.model.schema
@@ -301,7 +301,7 @@
         "default": true,
         "title": "Allow scroll over",
         "inputType": "Checkbox",
-        "help": "Some browsers pause script execution during scroll (e.g. IE8). If this option is selected the inview library will determine if elements became visible during the scroll event."
+        "help": "Some browsers pause script execution during scroll (e.g. some versions of iOS). If this option is selected the inview library will determine if elements became visible during the scroll event."
         }
       }
     }
--- a/src/core/schema/config.model.schema
+++ b/src/core/schema/config.model.schema
@@ -301,7 +301,7 @@
         "default": true,
         "title": "Allow scroll over",
         "inputType": "Checkbox",
-        "help": "Some browsers pause script execution during scroll (e.g. some versions of iOS). If this option is selected the inview library will determine if elements became visible during the scroll event."
+        "help": "This setting relates to browsers that pause script execution whilst the learner is scrolling (e.g. Safari for iOS 6 & 7). This causes problems for components that use 'inview' completion. Once scrolling has stopped 'scroll over' works around the paused execution by checking to see if the content was made visible during the scroll. Unless you need support for browsers that pause script execution on scroll, you can disable this setting."
         }
       }
     }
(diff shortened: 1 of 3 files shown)


src/core/schema/config.model.schema
```diff
@@ -291,6 +291,19 @@
...
+    "_inview": {
+      "type": "object",
+      "title": "jquery.onscreen (inview)",
+      "properties": {
+      "_allowScrollOver": {
+        "type": "boolean",
+        "default": true,
+        "title": "Allow scroll over",
+        "inputType": "Checkbox",
+        "help": "Some browsers pause script execution during scroll (e.g. IE8). If this option is selected the inview library will determine if elements became visible during the scroll event."
```
oliverfoster: I don't think IE8 pauses execution on scroll. This was originally for iOS6+7 which do pause execution.

chris-steele: Thanks @oliverfoster will amend

src/core/schema/config.model.schema
```diff
@@ -291,6 +291,19 @@
...
     "themeSettings": {
       "type": "object"
+    },
+    "_inview": {
+      "type": "object",
+      "title": "jquery.onscreen (inview)",
+      "properties": {
+      "_allowScrollOver": {
+        "type": "boolean",
+        "default": true,
```
moloko: This raises an interesting question... if I understand this correctly, `_allowScrollOver: true` was really only needed for 'browsers that pause script execution on scroll' i.e. Safari for iOS 6 & 7, neither of which are supported any more (by Adapt or by Apple).

So, do we continue to default it to `true` (to maintain backwards compatibility with existing courses) or change it to `false` because that's the better setting for the current releases of iOS?

(From what I can see it defaults to `true` in the code if there's no setting present in config.json)

src/core/schema/config.model.schema
```diff
@@ -291,6 +291,19 @@
...
+        "title": "Allow scroll over",
+        "inputType": "Checkbox",
+        "help": "Some browsers pause script execution during scroll (e.g. some versions of iOS). If this option is selected the inview library will determine if elements became visible during the scroll event."
```
moloko: I think this help text needs re-wording a bit so it is more helpful to the (generally-speaking) less technical authoring tool user. Probably need to resolve the issue of the correct default setting before we can improve on this.

src/core/schema/config.model.schema
chris-steele: @moloko good question; although the legacy branch is designed to support IE8-10, what is the support policy for iOS and Android?

moloko: for iOS it's that we support latest version and one version back (so currently 11 & 12). for Android we're considerably more generous and say that it's any version capable of running Chrome.

If no-one has a view on this let's leave it as-is and I'll just write something in the help text explaining that it defaults to `true` for backwards-compatibility reasons but that you'd probably want to change it to `false`.

src/core/schema/config.model.schema
oliverfoster: "Some browsers pause script execution during scroll (e.g. some versions of iOS). If this option is selected the inview library will determine if elements became visible during the scroll event."

This would be more accurate:

"Some browsers pause script execution during scroll (e.g. some versions of iOS). If this option is selected the inview library will determine if elements became visible after the scroll events have stopped, once the javascript begins executing again."

moloko: That's really helpful, thanks. How about this:
> "This setting relates to browsers that pause script execution whilst the learner is scrolling (e.g. Safari for iOS 6 & 7). This causes problems for components that use 'inview' completion. 'Scroll over' works around this by checking to see if content was made visible once scrolling has stopped. Unless you need support for browsers that pause script execution on scroll, you can disable this setting."

oliverfoster: Yup that's pretty close. It reads imprecisely in my brain with the order:  "... was made visible once.. [event occurrence].".

Could you rearrange the sentence such that event occurs and scroll over works around it afterwards?

"Once scrolling has stopped 'scroll over' works around the paused execution by checking if the content was made visible during the scroll."

src/core/schema/config.model.schema
```diff
@@ -291,6 +291,19 @@
     },
     "themeSettings": {
       "type": "object"
+    },
+    "_inview": {
+      "type": "object",
+      "title": "jquery.onscreen (inview)",
```
tomgreenfield:
```suggestion
      "title": "",
```
The extra information is probably not too useful for the average user – the key is already transformed into the main title. With the current schema we'd get a double title:
double

src/core/schema/config.model.schema
moloko:
```suggestion
        "help": "This setting relates to browsers that pause script execution whilst the learner is scrolling (e.g. Safari for iOS 6 & 7). This causes problems for components that use 'inview' completion. Once scrolling has stopped 'scroll over' works around the paused execution by checking to see if the content was made visible during the scroll. Unless you need support for browsers that pause script execution on scroll, you can disable this setting."
```

chris-steele:
Thanks @moloko, @tomgreenfield, @oliverfoster; amendments made

tomgreenfield:
I'll fix the change log on release.

PR #2382, opened 2019-03-14
update to use 2 spaces for indentation

nilslp:
Framework files (except for libraries) converted to use 2 spaces for indentation.
This should be the biggest PR of all to come.

### DIFF
--- a/src/core/js/accessibility.js
+++ b/src/core/js/accessibility.js
@@ -1,266 +1,266 @@
 define([
-    'core/js/adapt',
-    'a11y'
+  'core/js/adapt',
+  'a11y'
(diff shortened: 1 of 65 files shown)


moloko:
hi @nilsp @guywillis my understanding was that we would only be changing the LESS files to use 2 spaces..?

nilslp:
> hi @nilslp @guywillis my understanding was that we would only be changing the LESS files to use 2 spaces..?

https://www.youtube.com/watch?v=7fyppMgjuWo

oliverfoster:
you can commit these straight in once checked and 1x+1 @brian-learningpool & @guywillis

PR #2402, opened 2019-04-03
Resolves #2398, re-populates _assessmentState from 'assessments:regis…

### LINKED ISSUE #2398: Incorrect "Unknown" lesson status when completing assessment and course in different attempts
### Subject of the issue
Reported by @muthuselvan86 on Gitter.  If a course has an assessment and course content tracked as part of the completion criteria, in instances where the assessment is completed first, the session closed and then resumed, when the rest of the course content is completed the lesson status is incorrectly showing as "Unknown".

### Your environment
* Adapt Framework v3/4

### Steps to reproduce
As above.

### Expected behaviour
A "Completed" status would be written to the LMS.

### Actual behaviour
An "Unknown" status is being written because the assessment is not found.

brian-learningpool:
…ter'

This corrects an issue with incorrect course completion statuses being passed back
when a course was being completed over multiple sessions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "adapt_framework",
-  "version": "3.5.2",
+  "version": "3.5.3",
   "description": "Adapt Learning output framework",
   "repository": {
     "type": "git",
--- a/src/core/js/tracking.js
+++ b/src/core/js/tracking.js
@@ -20,7 +20,10 @@ define([
         setupEventListeners: function() {
             // Check if completion requires passing an assessment. 
             if (this._config._requireAssessmentCompleted) {
-                this.listenTo(Adapt, 'assessment:complete', this.onAssessmentComplete);
+                this.listenTo(Adapt, {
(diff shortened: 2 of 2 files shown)


oliverfoster:
Did it solve the other issue @brian-learningpool ?

brian-learningpool:
It corrects the issue with the incorrect course status and passes the correct `assessment` state when the `tracking:complete` event is triggered.

oliverfoster:
It reads much better also. With a clear api separation. (still, the event names are crepe).

oliverfoster:
👁

muthuselvan86:
Hi @brian-learningpool , it initializes the _assessmentState as soon as the app is ready(my understanding is as soon as app:dataready event triggers it calls assessment register to initialize the data). But does that bring old assessment data from session? We need to know whether the user has previously attempted the assessment(without completing lesson) and either Passed/Failed before triggering tracking:complete ?  Let me know if my understanding is not correct. Thanks.

PR #2403, opened 2019-04-03
Resolves #2398, re-populates _assessmentState from 'assessments:regis…

### LINKED ISSUE #2398: Incorrect "Unknown" lesson status when completing assessment and course in different attempts

brian-learningpool:
…ter'

This corrects an issue with incorrect course completion statuses being passed back
when a course was being completed over multiple sessions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "adapt_framework",
-    "version": "4.1.0",
+    "version": "4.1.1",
     "description": "Adapt Learning output framework",
     "repository": {
         "type": "git",
--- a/src/core/js/tracking.js
+++ b/src/core/js/tracking.js
@@ -20,7 +20,10 @@ define([
         setupEventListeners: function() {
             // Check if completion requires passing an assessment. 
             if (this._config._requireAssessmentCompleted) {
-                this.listenTo(Adapt, 'assessment:complete', this.onAssessmentComplete);
+                this.listenTo(Adapt, {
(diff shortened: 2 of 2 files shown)


moloko:
does this work OK if there are multiple assessments in the course? From a (cursory, admittedly) glance at the code it looks like it will call `onAssessmentRestored` once for each assessment...

brian-learningpool:
@moloko, good catch -- you're exactly right.  I'd only tested this with a single assessment.

I think we may have to actually update the Assessment extension to trigger a new event with the overall state.

brian-learningpool:
@oliverfoster / @moloko, `assessments:restored` actually fires once for every assessment too.  Do we actually require a new event?

moloko:
Yeah I think that might be best TBH, everything else I can think of is pretty hacky

moloko:
make sure to give the event a really unintuitive name though, in keeping with the style of this extension

PR #2407, opened 2019-04-05
issue fixes #2392: Add _htmlClassName

### LINKED ISSUE #2392: Feature request: Start controller add html class
### Original issue
It would be incredibly useful for Start Controller to add a custom class to the html tag. The class would be removed when the user navigates away from the start page.

Preference is for the class to be automatically defined so appropriate navigational elements, such as back or home buttons, can be removed in the base theme - e.g. `.is-start-page`.

It could be extended to support user defined classes if different behaviour is required across multiple start pages.

> Preference is for [...] [a] class to be [...] defined so appropriate navigational elements, such as back or home buttons, can be removed in the base theme [...]

> It could [...] support user defined classes if different behaviour is required across multiple  [...] pages.

sarveshwar-gavhane:
#2392 

implementation of feature request to add html class for start page.


src/core/js/router.js
```diff
@@ -345,13 +345,59 @@ define([
...
+        /**
+         * Sets the start page class as per configuration mentioned in "_start" attribute.
+         * Below code also considers all the example scenarios mentioned in below URL
+         * https://github.com/adaptlearning/adapt_framework/wiki/Content-starts-with-course.json#example-1
+         */
+        setStartPageClass: function(type, id) {
+            var startConfig = Adapt.course.get('_start');
+
+            if (!startConfig && !startConfig._isEnabled) return;
+
```
jamesrea83: This check should use || rather than &&. If startConfig is undefined, startConfig._isEnabled will throw an error.

src/core/js/router.js
```diff
@@ -345,13 +345,59 @@ define([
...
+                } else {
+                    var touchClass = Adapt.device.touch ? 'touch' : 'no-touch';
+                    if (this.$html.is(classes) && this.$html.hasClass(touchClass)) {
+                        startIdIndex = index;
+                    }
+                }
+            }.bind(this));
+
+                //Get the "_id" from _startIds to be applicable 
+                if (startIdIndex != null) {
```
danielstorey: maybe `typeof startIdIndex  === 'number'` would be a stronger check

src/core/js/router.js
```diff
@@ -345,13 +345,59 @@ define([
...
+                    pageID = startIds[startIdIndex]._id;
+                }
+
+                if (type === 'page' && pageID === id) {
+                    this.$html.addClass('is-start-page');
```
danielstorey: this is not a dealbreaker for me but I'm personally not a big fan of unnecessary booleans in classes - would `start-page` not suffice?

src/core/js/router.js
```diff
@@ -345,13 +350,59 @@ define([
                 .attr('data-location', Adapt.location._currentLocation);
 
             this.setDocumentTitle();
+            this.setStartPageClass(type, id);
```
oliverfoster: `setStartPageClass` is no longer needed

src/core/js/startController.js
```diff
@@ -21,7 +21,27 @@ define([
             var hasStartId = (startId)
                 ? true
                 : false;
+            
+            if (hasStartId) {
```
oliverfoster: This section won't work,  as explained on [this comment](https://github.com/adaptlearning/adapt_framework/issues/2392#issuecomment-483215043). Adding classes to the page through the `startController` won't keep them permanently on the page.

src/core/js/router.js
```diff
@@ -322,6 +322,9 @@ define([
             }
 
             Adapt.location._currentLocation = currentLocation;
+            
+            var locationModel = Adapt.findById(id);
+            var startControllerClass = (locationModel && locationModel.get('isStartPage')) ? locationModel.get('_htmlClassName') : '';
```
oliverfoster: This switch is unnecessary, as I explained in [this comment](https://github.com/adaptlearning/adapt_framework/issues/2392#issuecomment-483215043) adding classes through the `startController` won't add the classes permanently to the page.

```js
var htmlClassName = locationModel && locationModel.get('_htmlClassName') || '';
```

src/core/js/startController.js
rakeshjaware-exult: I agree that this section will execute only once  in a lifetime of application. But I would put this in a different way that if we are saying  `"_skipIfComplete": true` then it is desirable that we should not apply the class and we should allow navigation buttons to show. 

OR
 
The `"_skipIfComplete": true`  is only to avoid jumping to the specified page and rest of the things should work as is i.e. class should be applied even if we are saying "skip"?
I think this is contradictory behaviour.

oliverfoster: > It would be incredibly useful [...] to [be able to] add a custom class to the html tag.

> Preference is for the class to be [...] defined so appropriate navigational elements, such as back or home buttons, can be removed in the base theme

> ... The class would be removed when the user navigates away from the [...] page.

This behaviour is made complicated by the original description. It does not need to be dependent on `"_skipIfComplete": true` or the `startController`. We can achieve the same outcome by allowing `_htmlClassName` only.

sarveshwar-gavhane: i think, we will remove the start controller logic then and will add class directly.

src/core/schema/contentobject.model.schema
```diff
@@ -45,6 +45,14 @@
       "validators": [],
       "translatable": true
     },
+    "_htmlClassName": {
+      "type": "string",
+      "required": false,
+      "default": "",
+      "inputType": "Text",
+      "validators": [],
+      "help": "If set, this class will be applied to the html"
```
danielstorey: The following might make the help text clearer to less technical users
```suggestion
      "help": "If set, this class will be applied to the <html> element"
```

src/core/schema/course.model.schema
```diff
@@ -48,6 +48,14 @@
       "validators": [],
       "translatable": true
     },
+    "_htmlClassName": {
+      "type": "string",
+      "required": false,
+      "default": "",
+      "inputType": "Text",
+      "validators": [],
+      "help": "If set, this class will be applied to the html"
```
danielstorey: Same as above

src/core/schema/contentobject.model.schema
```diff
@@ -45,6 +45,14 @@
       "validators": [],
       "translatable": true
     },
+    "_htmlClassName": {
+      "type": "string",
+      "required": false,
+      "default": "",
+      "inputType": "Text",
+      "validators": [],
+      "help": "If set, this class will be applied to the <html> element"
```
moloko:
```suggestion
      "help": "If set, this class will be applied to the <html> element whenever this is the active page in the course"
```

moloko: just a thought - anyone know what will happen to < and > when rendered in AAT help tooltip?

src/core/schema/course.model.schema
```diff
@@ -48,6 +48,14 @@
       "validators": [],
       "translatable": true
     },
+    "_htmlClassName": {
+      "type": "string",
+      "required": false,
+      "default": "",
+      "inputType": "Text",
+      "validators": [],
+      "help": "If set, this class will be applied to the <html> element"
```
moloko:
```suggestion
      "help": "If set, this class will be applied to the <html> element when the top-level menu in the course is displayed"
```

Issue #2389, opened 2019-03-21
RTL layout issues


moloko:
### Subject of the issue/enhancement/features
In RTL content, the DOM order of the components is changed in a way that doesn't seem right. This seems to be do with the fix for #1670 

### Your environment
* FW 4.0.1
(Should be possible to replicate in any FW version since v2.2.0)

### Steps to reproduce
Using the out-of-the-box content, either set `_defaultDirection` to `"rtl"` in config.json or use the language picker extension to add an RTL language. Launch the course and go to the presentation components topic

### Expected behaviour
The order of the components should change to match the RTL layout i.e. the text component should be on the right and the graphic component on the left. Furthermore, in mobile layout the order should remain the same as for the LTR layout i.e. the text component should be shown before the graphic component .

### Actual behaviour
Although the order of the component in the DOM has changed, the visual layout has not: the text component is still on the left and the graphic component is still on the right. In mobile layout the graphic component is shown above the text component.

### Screenshots (if you can)
![adapt-rtl_desktop](https://user-images.githubusercontent.com/1249597/54746420-0c16ba00-4bc4-11e9-92a2-cb4e38d95f33.png)

![adapt-rtl_mobile](https://user-images.githubusercontent.com/1249597/54746430-10db6e00-4bc4-11e9-8d76-59ddcda1b10e.png)

amygroshek:
If this helps, as discussed in gitter, at present `AdaptModel.getChildren()` rearranges left/right positioning of 2 components in a block, checking a `_sortComponents` variable. `AdaptView.addChildren()` is the place where components are reordered if lang dir is rtl.

oliverfoster:
^ Agree with @amygroshek 

DOM order swapping (prepend) needs removing and LESS needs fixing.

moloko:
@guywillis any thoughts about how best to go about fixing the LESS? Swapping the `.component-left` and `.component-right` classes certainly worked for me - but whether that's the 'right' way to go about it or not I'm not sure...

guywillis:
On the face of it, this seems like a relatively straight forward fix. But after a bit of digging I'm finding issues with all the solutions.

The different scenarios tested:

1. Leave auto sort on
1. Turn off auto sort 
1. Turn off auto sort and add CSS to swap components
1. Turn off auto sort and swap component layout in js
1. Turn off auto sort and use display flex instead of floats

So where does that leave us...

[...]

PR #2413, opened 2019-04-10
Resolves #2395 and #2389 - Capture child views

### LINKED ISSUE #2395: Store view references
Store references to views on the page so that view instances can be retrieved at will without having to first store them by listening to all the different 'someView:postRender' events.

View references should be stored as child views on the parent. This will ensure that references remain in context of the Adapt model hierarchy, and that all references are removed when the location changes, so as to prevent memory leaks.

A `findDescendantViews` method should be created, to work in the same way as `findDescendantModels`

A `getViewByModelId` method would also be quite handy

danielstorey:
#2395
* Capture child views
* Provide findViewByModelId api

#2389
* Remove rtl prepend

### DIFF
--- a/src/core/js/adapt.js
+++ b/src/core/js/adapt.js
@@ -240,36 +240,27 @@ define([
 
     Adapt.findViewByModelId = function(id) {
         var model = Adapt.findById(id);
-
         if (!model) {
             return;
         }
 
-        // Create an array of ids which will lead to the specified view
-        var idPath = [model.get('_id')];
-        while (model && model !== Adapt.parentView.model) {
-            model = model.getParent();
+        var idPathToView = [];
+        var currentLocationId = Adapt.location._currentId;
+        var currentLocationModel = _.find(model.getAncestorModels(), function(model) {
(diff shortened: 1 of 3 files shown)


moloko:
Useful!

Two things:
1. Any chance you could add a brief bit of documentation about these to the [Adapt API Wiki page](https://github.com/adaptlearning/adapt_framework/wiki/Adapt-API)?
1. I can totally see why you did it - but it does seem a little odd to have some code that addresses a totally separate issue (#2389) in here. It also doesn't completely fix the issue (in that although the DOM order is now correct the layout - in medium/large screen size - isn't. Using block b-05 as an example, when in RTL mode, the text component c-05 should be on the right and the graphic component on the left). I wonder if it would be better to do this as a separate PR that includes the layout fix? Not sure whether this is best done by re-assigning the `.component-left` and `.component-right` classes or by amending those classes for RTL mode. Suggestions welcome...!

oliverfoster:
https://github.com/adaptlearning/adapt_framework/wiki/Adapt-API updated @moloko

PR #2429, opened 2019-05-13
refactor adapt.js so that all the functions are encapsulated properly in AdaptModel

moloko:
amend app.js to call new `Adapt.init()` function (instead of `Adapt.initialize`) and to change double quotes to single


moloko:
@oliverfoster somehow my changes in 5890ee711334c25a1b15c34791924bf4779082af are stopping require from loading adapt.js, but for the life of me I can't figure out what's wrong - could you take a look?

PR #2437, opened 2019-05-17
refactor app.js and adapt.js

moloko:
add data.js and a `Promise` polyfill

resolves #2429


src/core/js/app.js
```diff
@@ -25,201 +16,33 @@ require([
...
         Adapt.off('adaptCollection:dataLoaded courseModel:dataLoaded');
-    }
+    });
 
     function addNavigationBar() {
 
-        var adaptConfig = Adapt.course.get("_navigation");
+        var adaptConfig = Adapt.course.get('_navigation');
 
         if (adaptConfig && adaptConfig._isDefaultNavigationDisabled) {
```
moloko: as far as I can see, this is the only reference to `_isDefaultNavigationDisabled` (and there's no documentation or schema entry for it)

is this something we actually need?

if not we could drop this, move L46 into `triggerInitialize` and remove the entire `addNavigationBar` function

moloko:
Tested in IE11, Firefox, Chrome and iOS-safari v12 for iPad and iPhone

moloko:
Bug with multilanguage courses: if you switch languages using the icon in the navbar, that icon and the drawer icon are not shown after Adapt switches to the new language. 

Seems to be because the `.navigation-inner` doesn't exist at the time the language picker extension tries to append `languagePickerNavView` to it. In fact, it looks like the `NavigationView` doesn't get recreated after the language is changed...

Will do some more digging next week.

src/core/js/app.js
oliverfoster: This is a precursor to getting rid of the navigation bar altogether and putting it in a separate plugin. Ideally we'd just document the property?

moloko:
closing this so that I can set up a 'draft' PR instead

PR #2442, opened 2019-05-24
Resolves #2439, adds a timeout to Adapt.wait

### LINKED ISSUE #2439: Adapt.wait should implement a timeout in order to prevent blocking of rendering
The `Adapt.wait` API has `begin()` and `end()` calls but in cases where calls to `begin()` do not have a corresponding call to `end()` (for whatever reason) it can result in a course failing to load and difficult to diagnose errors.

### Your environment
Adapt Framework v3.5.3

### Steps to reproduce
Inside a plugin call `Adapt.wait.begin()` and do not call `Adapt.wait.end()`.

### Expected behaviour
After a given timeout, the queue should be flushed together with a debug message logged to the console to indicate as such.

### Actual behaviour
The Adapt content appears to get stuck in an intermediate state between loading and rendering.

brian-learningpool:
To prevent an unhandled expection or a call to begin() not having a corresponding end() from blocking rendering of the Adapt content, a 7-second timeout is now added on every begin.  If the time has elapsed since the last begin() the outstanding calls to Adapt.wait.end() will be triggered.

Also added a _.once() call to the for() callback and replaced double quote marks in strings with single quotes.

### DIFF
--- a/src/core/js/wait.js
+++ b/src/core/js/wait.js
@@ -3,11 +3,13 @@ define(function() {
      var Wait = Backbone.Controller.extend({
 
         initialize: function() {
-            _.bindAll(this, "begin", "end");
+            _.bindAll(this, 'begin', 'end');
         },
 
         _waitCount: 0,
         _callbackHandle: null,
+        _timeoutHandlerId: null,
+        _timeoutInSeconds: 7,
 
         /**
          * Returns true if there are items in the waiting count.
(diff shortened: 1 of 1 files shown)


src/core/js/wait.js
```diff
@@ -18,6 +20,35 @@ define(function() {
...
 
+        /**
+         * Starts or re-starts a timer to ensure that pending calls to end()
+         * are actually executed after a timeout period.
+         */
+        startTimer: function() {
+            this.stopTimer();
+
+            this._timeoutHandlerId = setInterval(function() {
+              // Flush Adapt.wait due to timeout
```
oliverfoster: Mixed indentation

src/core/js/wait.js
```diff
@@ -98,7 +129,7 @@ define(function() {
 
             this.begin();
             _.defer(function() {
-                callback(this.end);
+                _.once(callback(this.end));
```
oliverfoster:
```js
callback(_.once(this.end));
```

brian-learningpool: Whoops!  Good catch. 👍

brian-learningpool:
@oliverfoster, it wasn't intended.  I had multiple framework version checkouts and it looks like I pushed one in an intermediate state.

I intended to trigger `startTimer()` whenever `begin()` is called.

src/core/js/wait.js
```diff
@@ -36,6 +67,8 @@ define(function() {
                 this._callbackHandle = null;
             }
 
+            this.startTimer();
```
oliverfoster: Ahh yea, s'all good.

PR #2444, opened 2019-05-30
fixed #2441

### LINKED ISSUE #2441: Grunt, build & changes to libraries.
Start `grunt dev`, then make a change to a file in core/js/libraries. Grunt will detect the change and start a build process, but the corresponding core/js/libraries file in the build folder will be unchanged.

In order to get the change into the build folder, you must manually stop then restart `grunt dev`.

Is this by design, or should libraries get updated on change like other files in src/?

NayanKhedkar:
Fixed #2441 and also changes the calling of copy task now it's only allowed to copying modified files.

### DIFF
--- a/grunt/config/watch.js
+++ b/grunt/config/watch.js
@@ -34,7 +34,7 @@ module.exports = {
             '!<%= sourcedir %>templates/templates.js',
             '!<%= sourcedir %>core/js/scriptLoader.js',
             '!<%= sourcedir %>core/js/libraries/**/*.js',
-            '!<%= sourcedir %>extensions/*/libraries/**/*.js',
+            '!<%= sourcedir %>extensions/**/libraries/**/*.js',
             '!<%= sourcedir %>components/**/libraries/**/*.js',
             '!<%= sourcedir %>menu/<%= menu %>/libraries/**/*.js', 
             '!<%= sourcedir %>theme/<%= theme %>/libraries/**/*.js'
@@ -82,17 +82,17 @@ module.exports = {
         tasks: ['newer:copy:scriptLoader']
     },
     libraries: {
-        files: ['<%= sourcedir %>core/js/libraries/**/*'],
-        tasks: ['newer:copy:libraries']
-    },
-    required: {
         files: [
-            '<%= sourcedir %>extensions/*/required/**/*',
-            '<%= sourcedir %>extensions/*/libraries/**/*',
+            '<%= sourcedir %>core/js/libraries/**/*',
+            '<%= sourcedir %>extensions/**/libraries/**/*',
             '<%= sourcedir %>components/**/libraries/**/*',
             '<%= sourcedir %>menu/<%= menu %>/libraries/**/*',
             '<%= sourcedir %>theme/<%= theme %>/libraries/**/*'
         ],
-        tasks: 'newer:copy:required'
+        tasks: ['newer:copy:libraries']
+    },
+    required: {
+        files: ['<%= sourcedir %>extensions/*/required/**/*'],
+        tasks: ['newer:copy:required']
     }
 }
--- a/grunt/config/watch.js
+++ b/grunt/config/watch.js
@@ -18,11 +18,11 @@ module.exports = {
     },
     courseJson: {
         files: ['<%= sourcedir %>course/**/*.<%= jsonext %>'],
-        tasks : ['jsonlint', 'check-json', 'copy:courseJson', 'schema-defaults', 'create-json-config']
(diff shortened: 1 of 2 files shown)


grunt/config/watch.js
```diff
@@ -33,62 +33,66 @@ module.exports = {
...
     },
     required: {
-        files: ['<%= sourcedir %>extensions/*/required/**/*', '<%= sourcedir %>extensions/*/libraries/**/*'],
-        tasks: 'copy:required'
+        files: [
+            '<%= sourcedir %>extensions/*/required/**/*',
+            '<%= sourcedir %>extensions/*/libraries/**/*',
+            '<%= sourcedir %>components/**/libraries/**/*',
+            '<%= sourcedir %>menu/<%= menu %>/libraries/**/*',
+            '<%= sourcedir %>theme/<%= theme %>/libraries/**/*'
```
oliverfoster: these `libraries` directives should be `required` directives

grunt/config/watch.js
```diff
@@ -33,62 +33,66 @@ module.exports = {
...
+        tasks: ['newer:copy:scriptLoader']
     },
     libraries: {
-        files: [
-            '<%= sourcedir %>core/js/libraries/require.min.js',
-            '<%= sourcedir %>core/js/libraries/modernizr.js',
-            '<%= sourcedir %>core/js/libraries/jquery.min.js'
-        ],
-        tasks: ['copy:libraries']
+        files: ['<%= sourcedir %>core/js/libraries/**/*'],
```
oliverfoster: this bit should include the `libraries` folder directives for extensions, menus, components etc

grunt/config/watch.js
```diff
@@ -18,11 +18,11 @@ module.exports = {
     },
     courseJson: {
         files: ['<%= sourcedir %>course/**/*.<%= jsonext %>'],
-        tasks : ['jsonlint', 'check-json', 'copy:courseJson', 'schema-defaults', 'create-json-config']
+        tasks : ['jsonlint', 'check-json', 'newer:copy:courseJson', 'schema-defaults', 'create-json-config']
```
oliverfoster: did the inclusion of the `newer:` library give any benefits?

NayanKhedkar: Yes, it facilitates to copy only modified files instead of all the files.

oliverfoster: is it faster?

grunt/config/watch.js
NayanKhedkar: I have included the reference of the directive  for extensions, menus, components into required directives, here is the reference,

https://github.com/adaptlearning/adapt_framework/blob/b6988c335cd9f8b077c7d9f52fc379a5ddc6721b/grunt/config/copy.js#L211

grunt/config/watch.js
NayanKhedkar: yes, as compared to the previous one.

grunt/config/watch.js
oliverfoster: the `required` folder and the `libraries` folder are two different things.

grunt/config/watch.js
oliverfoster: it looks as though newer only compares file modified times one way 
https://github.com/tschaub/grunt-newer/blob/master/lib/util.js#L27-L28

this means it probably won't work if i paste an older file over a newer file?

NayanKhedkar: If the file in build folder the same as the file in src folder then it doesn't copy to build folder.when you paste an older file over a newer file. watch task listens to the changes and calls newer:copy task and it works in the above case also.

grunt/config/watch.js
NayanKhedkar: Do we really need 

> 

{
                    expand: true,
                    src: ['components/**/libraries/**/*', 'extensions/**/libraries/**/*', 'menu/<%= menu %>/libraries/**/*', 'theme/<%= theme %>/libraries/**/*'],
                    cwd: '<%= sourcedir %>',
                    dest: '<%= outputdir %>/libraries/',
                    filter: function(filepath) {
                        return grunt.config('helpers').includedFilter(filepath);
                    },
                    rename: _.partial(collate, "libraries")
                }

> at  https://github.com/adaptlearning/adapt_framework/blob/b6988c335cd9f8b077c7d9f52fc379a5ddc6721b/grunt/config/copy.js#L212

so can we move this to 
https://github.com/adaptlearning/adapt_framework/blob/b6988c335cd9f8b077c7d9f52fc379a5ddc6721b/grunt/config/copy.js#L199

oliverfoster: Yup, that works.

NayanKhedkar: yes, I will do the change an update it soon.

Issue #2440, opened 2019-05-21
Hot Graphic scrolling background


leahyjulian:
### Bug adapt-contrib-hotgraphic
Scrolling background when pop up is showing

### Your environment
* Adapt Framework  4.1.1
* Latest Chrome Version 74.0.3729.157 (Official Build) (64-bit)
* Microsoft Windows 10

### Steps to reproduce
Within the popup body of a Hot Graphic Component add a lot of text so when viewed in the browser the text goes below the view port

### Expected behaviour
Should be able to scroll within the box to read all text (default behaviour in firefox)

### Actual behaviour
Instead of the box background scrolling the main background scrolls

### Screenshots (if you can)
Note background position - before scroll
![image](https://user-images.githubusercontent.com/48165156/58064510-7d7be480-7bd6-11e9-9c58-1edb876f72da.png)
After scroll
![image](https://user-images.githubusercontent.com/48165156/58064556-b0be7380-7bd6-11e9-871e-54c8f8137743.png)

brian-learningpool:
The console log gives the following error:
`jquery.min.js:formatted:2117 [Intervention] Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/6662647093133312`

moloko:
Tried updating to the latest version of JQuery just on the off-chance that might solve it - but sadly not.

Looks to be originating from `a11y_setupScrollListeners` in jquery.a11y.js

brian-learningpool:
Yes, the error message is being thrown by this line:
https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/libraries/jquery.a11y.js#L76

oliverfoster:
This concept of specifying the passive property needs applying to the other scroll events
https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/libraries/jquery.a11y.js#L556

moloko:
> 
> 
> This concept of specifying the passive property needs applying to the other scroll events
> https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/libraries/jquery.a11y.js#L556

They may have to be changed to vanillajs first as it doesn't look like JQuery supports passive event listeners yet https://github.com/jquery/jquery/issues/2871

oliverfoster:
That's what's needed. jQuery doesn't support event attachment options, as you said.

moloko:
[...]

PR #2446, opened 2019-06-03
issue/2440: Chrome bug - scrolling background

oliverfoster:
#2440 
* Added `passive:false` to natively passive event attachment, where required

### DIFF
--- a/src/core/js/libraries/jquery.a11y.js
+++ b/src/core/js/libraries/jquery.a11y.js
@@ -549,21 +549,21 @@ define([
 
     // PRIVATE $.a11y FUNCTIONS
         function a11y_setupScrollListeners() {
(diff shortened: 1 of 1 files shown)


oliverfoster:
@moloko 
`preventScrollKeys` and `nativePreventScroll` both use `preventScroll` just that `preventScrollKeys` passes a jquery `Event` object and the native events don't. `jQuery.on` is needed for key event propagation at the moment because key propagation doesn't work natively using the `addEventListener` api. It'll break my brain today to try to simplify these functions but yes, they do need simplifying.

brian-learningpool:
This issue affects v3 also.

moloko:
> This issue affects v3 also.

see #2454

Issue #2449, opened 2019-06-06
Ongoing accessibility integrations


oliverfoster:
### Part 2 of v4

1. Create APIs from existing functionality:
* Adapt.a11y
  * toggleHidden
  * toggleAccessibleEnabled
  * toggleAccessible
  * toggleEnabled
  * findFirstTabbable
  * findFirstReadable
  * findTabbable
  * findReadable
  * isTabbable
  * isReadable
  * focusNext
  * focusFirst
  * focus
  * normalize
  * removeBreaks
  * scrollEnable
  * scrollDisable
  * popupOpened
  * popupClosed
2. Segregate complex behaviour into separate files
* Browser focus modifications
* Popup aria and tab management
* Scroll enable and disable
* Popup focus wrapping
3. Added warnings for all old APIs
4. Extend configuration options for control over features
5. Move all helpers into "core/js/helpers"
6. Add new features
* Keyboard enabled tab onlines / outline disablement

Should all be backwards compatible.

PR #2453, opened 2019-06-10
issue/2449: Accessibility API centralisation

oliverfoster:
#2449 
* Accessibility API centralisation


src/core/js/a11y/keyboardFocusOutline.js
```diff
@@ -0,0 +1,71 @@
...
+            this._start();
+        },
+
+        /**
+         * Add styling classes if required.
+         */
+        _start: function() {
+            var config = Adapt.a11y.config;
+            if (config._isFocusOutlineDisabled) {
+            this.$html.addClass('a11y-disable-focusoutline');
```
moloko: indentation

src/core/js/a11y/keyboardFocusOutline.js
```diff
@@ -0,0 +1,71 @@
...
+
+        /**
+         * Handle key down events for on a tabbable element.
+         *
+         * @param {Object} event jQuery event object.
+         */
+        _onKeyDown: function(event) {
+            var config = Adapt.a11y.config;
+            if (config._isFocusOutlineDisabled) {
+            this.$html.addClass('a11y-disable-focusoutline');
```
moloko: indentation

src/core/js/a11y/browserFocus.js
```diff
@@ -0,0 +1,76 @@
...
+            this.$body = $('body');
+            this.listenTo(Adapt, {
+                'accessibility:ready': this._attachEventListeners
+            });
+        },
+
+        _attachEventListeners: function() {
+            this.$body
+                .on('blur', '*', this._onBlur)
+                .on("blur", this._onBlur);
```
moloko: double > single quotes

src/core/js/a11y.js
```diff
@@ -0,0 +1,686 @@
...
+            // Stop document reading
+            _.defer(function() {
+                Adapt.a11y.toggleHidden('.page, .menu', true);
+            });
+        },
+
+        _onNavigationEnd: function(view) {
+            // Prevent sub-menu items provoking behaviour
+            if (view && view.model && view.model.get('_id') !== Adapt.location._currentId) return;
+            if (!this.isEnabled()) return;
```
brian-learningpool: Minor: since there are two early returns here is it worth combining this with an OR statement in the `if` above?

src/core/js/a11y.js
```diff
@@ -0,0 +1,686 @@
...
+        },
+
+        /**
+         * Adds or removes `aria-hidden` attribute to elements.
+         *
+         * @param {Object} $elements
+         * @param {boolean} [isHidden=true]
+         * @returns {Object} Returns `Adapt.a11y`
+         */
+        toggleHidden: function($elements, isHidden) {
```
brian-learningpool: `$elements` is actually a JQuery selector.  It might make sense to rename this parameter and declare `var $element = $(selector);` below.

src/core/js/a11y.js
```diff
@@ -0,0 +1,686 @@
...
+            $elements = $($elements);
+            var config = Adapt.a11y.config;
+            if (!config._isEnabled) return this;
+            if (!config._isAriaHiddenManagementEnabled) return this;
```
brian-learningpool: Minor: combine lines 160-161 with an OR since they both return the same thing?

src/core/js/a11y.js
```diff
@@ -0,0 +1,686 @@
...
+         * @param {boolean} $elements
+         * @param {boolean} [isReadable=true]
+         * @returns {Object} Returns `Adapt.a11y`
+         */
+        toggleAccessible: function($elements, isReadable) {
+            $elements = $($elements);
+            var config = Adapt.a11y.config;
+            if (!config._isEnabled) return this;
+            if (!config._isAriaHiddenManagementEnabled) return this;
+            if ($elements.length === 0) return this;
```
brian-learningpool: Minor: Should lines 195-197 these be combined?

src/core/js/a11y.js
```diff
@@ -0,0 +1,686 @@
...
+            return $element.find('*').filter(function(index, element) {
+                return this.isReadable(element);
+            }.bind(this));
+        },
+
+        /**
+         * Check if the element is natively or explicitly tabbable.
+         *
+         * @param {Object} $element
+         * @returns {boolean}
```
brian-learningpool: This can also return `undefined`.

src/core/js/a11y.js
```diff
@@ -0,0 +1,686 @@
...
+            _.find($nextSiblings, function(sibling) {
+                var $sibling = $(sibling);
+                var value = iterator($sibling);
+
+                // skip this sibling if explicitly instructed
+                if (value === false) return;
+
+                if (value) {
+                // sibling matched
+                $found = $sibling;
```
brian-learningpool: Indentation is off here.

src/core/js/a11y.js
```diff
@@ -0,0 +1,686 @@
...
+                return $notFound;
+            }
+
+            // keep walked+passed children in a stack
+            var stack = [{
+                item: $element[0],
+                value: undefined
+            }];
+            var i = 0;
+            var c = i+1;
```
brian-learningpool: There is possibly a better variable name than `c`.

src/core/js/a11y.js
```diff
@@ -0,0 +1,686 @@
...
+         * @param {FocusOptions} options
+         * @returns {Object} Returns `Adapt.a11y`
+         */
+        focus: function($element, options) {
+            options = new FocusOptions(options);
+            $element = $($element).first();
+            var config = Adapt.a11y.config;
+            if (!config._isEnabled) return this;
+            if (!config._isFocusAssignmentEnabled) return this;
+            if ($element.length === 0) return this;
```
brian-learningpool: Minor: These early returns could be one `if`.

src/core/js/a11y.js
```diff
@@ -0,0 +1,686 @@
...
+            function perform() {
+                if (options.preventScroll) {
+                    var y = $(window).scrollTop();
+                    try {
+                    if ($element.attr('tabindex') === undefined) {
```
brian-learningpool: Indentation goes off here.

src/core/js/a11y/browserFocus.js
```diff
@@ -0,0 +1,76 @@
...
+         * check to see if it needs the `data-a11y-force-focus` attribute removing
+         * and check to see if it was blurred because a disabled attribute was added.
+         * If a disabled attribute was added, the focus will be moved forward.
+         *
+         * @param {Object} event jQuery event object.
+         */
+        _onBlur: function(event) {
+            var config = Adapt.a11y.config;
+            if (!config._isEnabled) return
+            if (!config._isFocusNextOnDisabled) return;
```
brian-learningpool: Minor: Can be combined in one `if`.

src/core/js/a11y.js
oliverfoster: Even though it _CAN_ be an selector, `$element = $($element)` is there to ensure that it is always converted into a jQuery object where it receives a string or Array or jQuery object [API](https://api.jquery.com/jQuery/), I'd prefer to leave it as is if that's ok?

brian-learningpool: Good job we're not using Typescript.  :)

Yeah, that's ok.  Can you maybe update the JSDoc in that case though to indicate that the parameter can be any of those types?

brian-learningpool:
👀 Looks good! 👍

src/core/js/a11y.js
oliverfoster: If it were typescript we'd just palm it off onto whatever madness the jQuery selector uses to define its attribute or return type, rather than just the generic "Object".

oliverfoster: https://github.com/andraaspar/jquery-ts/blob/master/src/IStatic.d.ts#L22

`string|any` ... how boring...

src/core/js/a11y/browserFocus.js
```diff
@@ -0,0 +1,84 @@
...
+            }
+            // Move focus to next readable element
+            Adapt.a11y.focusNext($element);
+        },
+
+        /**
+         * Force focus when clicked on a tabbale element,
+         * making sure `document.activeElement` is updated.
+         *
+         * @param {Object} event jQuery event object.
```
moloko: If you add `"@types/jquery": "^3.3.31"` to package.json then you can change this to
```suggestion
         * @param {JQuery.Event}
```
and get proper type definitions appearing:
image

src/core/js/a11y/popup.js
```diff
@@ -0,0 +1,212 @@
...
+                    if (previousAriaHidden === '') {
+                        $item.removeAttr('aria-hidden');
+                    } else {
+                        $item.attr({
+                            'aria-hidden': previousAriaHidden
+                        });
+                    }
+                }
+            }.bind(this));
+            return document.activeElement = this._focusStack.pop();
```
moloko: wrap in brackets to stop linter from complaining
```suggestion
            return (document.activeElement = this._focusStack.pop());
```

src/core/js/a11y/scroll.js
```diff
@@ -0,0 +1,304 @@
+define(function() {
+
+    /**
+     * Controller for blocking scroll events on specified elements.
+     * @class
+     */
+    var Scroll = Backbone.Controller.extend({
+
+        initialize: function() {
+            _.bindAll(this, `_onTouchStart`, `_onTouchEnd`, '_onScrollEvent', '_onKeyDown');
```
moloko:
```suggestion
            _.bindAll(this, '_onTouchStart', '_onTouchEnd', '_onScrollEvent', '_onKeyDown');
```

src/core/js/a11y/scroll.js
```diff
@@ -0,0 +1,304 @@
...
+         */
+        _isScrolling: function($target, directionY) {
+            var scrollTop = Math.ceil($target.scrollTop());
+            var innerHeight = $target.outerHeight();
+            var scrollHeight = $target[0].scrollHeight;
+            var hasScrollingSpace = false;
+            switch (directionY) {
+                case 'down':
+                    hasScrollingSpace = scrollTop + innerHeight < scrollHeight;
+                    if (hasScrollingSpace) {
```
moloko: if it's deliberate that it should fall through to the 'up' case (if `hasScrollingSpace` is `false`), could you mark it with a /* falls through */ to stop linters from flagging this - and maybe to provide a short explanation why

oliverfoster: done

src/core/js/a11y/browserFocus.js
oliverfoster: done

src/core/js/a11y/deprecated.js
```diff
@@ -0,0 +1,174 @@
...
+            return this;
+        },
+
+        a11y_selected: function() {
+            Adapt.a11y.log.removed('$("..").a11y_selected is removed. Please use aria-live instead.');
+            return this;
+        },
+
+        a11y_on: function(isOn) {
+            Adapt.a11y.log.deprecated('$("..").a11y_on, use Adapt.a11y.findTabbable($element); and Adapt.a11y.toggleAccessible($elements, isAccessible); instead.');
```
moloko: Is there any way of giving more detail about what's calling the deprecated function? At the moment you get a lot of warnings in the console but no detail on where to look to track down the source of the warning...
image

src/core/js/helpers.js
```diff
@@ -3,6 +3,16 @@ define([
...
 ], function(Handlebars, Adapt){
 
+    var defaultAriaLevels = {
+        '_menu': 1,
+        '_menuItem': 2,
+        '_page': 1,
+        '_article': 2,
+        '_block': 3,
+        '_component': 4,
+        '_componentItem': 5
```
moloko: as per https://github.com/adaptlearning/adapt_framework/pull/2496
```suggestion
        '_componentItem': 5,
        '_notify': 1
```

src/core/js/a11y/deprecated.js
oliverfoster: Use a browser which outputs the callstack. I'm afraid you can't change the call stack output to the console without logging or throwing from the original location.

image

Or we could output a separate call stack [console.trace()](https://developer.mozilla.org/en-US/docs/Web/API/Console/trace)? It'll really pollute the console output though.

moloko: erm that's weird I tried that in all browsers (other than Safari). the above screenshot was from Chrome. maybe I'm missing a setting?

moloko:
that's a lot of settings being added to config.json - might be worth considering how likely people are to actually need to change those and consider instead having them in something like the `_advancedSettings` of the spoor plugin - or even `_playerOptions` like the media component

src/core/js/a11y/deprecated.js
moloko: Nope, can't see any setting to change this... weird.

oliverfoster: image

Windows 10

oliverfoster: [console.groupCollapsed](https://developer.mozilla.org/en-US/docs/Web/API/Console/groupCollapsed) + [console.trace](https://developer.mozilla.org/en-US/docs/Web/API/Console/trace) could work?

moloko: yeah I'm on the same version of Chrome... I can't figure it out at all, it's most annoying as obviously that would be very useful! It's it's just some quirk of my computer then let's leave it; after all once all the core plugins have been updated it should be much easier to track down purely by searching the code for that function name.

moloko:
config.model.schema needs updating with the new settings (or a media-component style JSON input), otherwise this looks good. I haven't tested in Jaws but have given it a bit of a go in IE11, Chrome & FF just to make sure there were no obvious showstoppers.

src/core/js/a11y/browserFocus.js
```diff
@@ -0,0 +1,84 @@
...
+            // due to the addition of a disabled class
+            if (!$element.is('[disabled]')) {
+                return;
+            }
+            // Move focus to next readable element
+            Adapt.a11y.focusNext($element);
+        },
+
+        /**
+         * Force focus when clicked on a tabbale element,
```
moloko:
```suggestion
         * Force focus when clicked on a tabbable element,
```

src/core/js/helpers.js
```diff
@@ -196,21 +230,158 @@ define([
...
-    helpers['numbers'] = helpers['inc'];
-    helpers['lowerCase'] = helpers['lowercase'];
+    _.extend(helpers, {
 
-    for (var name in helpers) {
-        if (helpers.hasOwnProperty(name)) {
-             Handlebars.registerHelper(name, helpers[name]);
+        if_value_equals: function() {
+            Adapt.a11y.log.deprecated('if_value_equals, use equals instead.');
+            return helpers['equals'].apply(this, arguments);
```
moloko: I don't think there's any need to use 'associative array' style lookup here (or on L372 and L377)
```suggestion
            return helpers.equals.apply(this, arguments);
```

src/core/js/helpers.js
```diff
@@ -196,21 +230,158 @@ define([
...
-        if (helpers.hasOwnProperty(name)) {
-             Handlebars.registerHelper(name, helpers[name]);
+        if_value_equals: function() {
+            Adapt.a11y.log.deprecated('if_value_equals, use equals instead.');
+            return helpers.equals.apply(this, arguments);
+        },
+
+        numbers: function() {
+            Adapt.a11y.log.deprecated('numbers, use inc instead.');
+            return helpers['inc'].apply(this, arguments);
```
moloko:
```suggestion
            return helpers.inc.apply(this, arguments);
```

src/core/js/helpers.js
```diff
@@ -196,21 +230,158 @@ define([
...
+        },
+
+        lowerCase: function() {
+            Adapt.a11y.log.deprecated('lowerCase, use lowercase instead.');
+            return helpers['lowercase'].apply(this, arguments);
```
moloko:
```suggestion
            return helpers.lowercase.apply(this, arguments);
```

PR #2462, opened 2019-06-26
Resolves #2377, adds ESLint as a dev dependency

### LINKED ISSUE #2377: v5: Linting / Code styles
### Subject of the issue
* Remove [jscs and jslint](https://github.com/adaptlearning/adapt_framework/commit/316cd5771d1ba21d4ff7a820f4cdc2b03b8f05c6) as [jscs](https://www.npmjs.com/package/jscs) is deprecated 
* Use eslint or an eslint derivative (https://standardjs.com/#standardjs--the-rules or https://github.com/standard/eslint-config-standard/blob/master/eslintrc.json)
* Add npm/grunt task to allow systems integration

brian-learningpool:
Includes .eslintrc file defaulted to standardjs with some caveats, e.g. 4 spaces rather than 2, semicolons allowed, etc.  (See 'rules' for details.)

Some common common globals (Backbone, underscore) have been added but there are bound to be more required.


brian-learningpool:
I'm not suggesting this be merged right away, but I've put the PR up as a placeholder for a conversation on how to best roll this out.  Enforcing this over the framework core codebase will impact probably every .js file due to legacy inconsistencies.

.eslintignore
```diff
@@ -0,0 +1,6 @@
+# Ignore minified library files.
+/src/core/js/libraries/*.min.js
```
oliverfoster: depending on this [pr#2473](https://github.com/adaptlearning/adapt_framework/pull/2473), these paths will need to change to `/src/core/libraries/`

brian-learningpool:
@moloko we do have some JS in that folder which we do maintain.  The third party libraries we require are usually minified, which is why I chose to ignore those.

oliverfoster:
I think exceptions aren't worth catering for. Can we just assume that everything in the libraries folder is fine as is?

brian-learningpool:
@oliverfoster, I've updated the PR to leave the `libraries` as is.  If you run `npm install` it'll install ESLint as a dev dependency, which I've aliased with the command `npm run lint`.

I have it down to 22 lint warnings, with a few maybe you should take a look at, e.g.`Return statement should not contain assignment`, etc.

grunt/config/translate.js
```diff
@@ -1,10 +1,10 @@
...
   return {
-    masterLang: "en",
+    masterLang: 'en',
     targetLang: null,
-    format: "csv",
-    csvDelimiter: ",",
-    "shouldReplaceExisting": false
+    format: 'csv',
+    csvDelimiter: ',',
+    'shouldReplaceExisting': false
```
cahirodoherty-learningpool: Minor: Is the quote wrapping necessary for this 'shouldReplaceExisting'?

brian-learningpool: It isn't, but it was just the linter's auto-fix which replaced the previous quote marks in this instance.

moloko:
> I have it down to 22 lint warnings, with a few maybe you should take a look at, e.g.`Return statement should not contain assignment`, etc.

These will likely just be cases where brackets need to be added, I have been fixing these as and when I've been doing code tidy stuff e.g. https://github.com/adaptlearning/adapt_framework/blob/74873f9c12a63e1675cc75c8242e1736c7614a4f/src/core/js/data.js#L245

grunt/tasks/less.js
```diff
@@ -33,28 +34,29 @@ module.exports = function(grunt) {
...
+        ? screenSize.small / baseFontSize
+        : screenSize.small;
 
       imports += `\n@adapt-device-large: ${largeEmBreakpoint}em;`;
       imports += `\n@adapt-device-medium: ${mediumEmBreakpoint}em;`;
       imports += `\n@adapt-device-small: ${smallEmBreakpoint}em;\n`;
     }
 
     if (options.mandatory) {
+      // eslint-disable-next-line no-redeclare
```
moloko: would be preferable to switch the `for` loop counter to be declared with `let` rather than use this eslint directive (same goes for line 69)

src/core/js/views/questionView.js
```diff
@@ -27,9 +27,9 @@ define([
       ].join(' ');
     },
 
-    //////
+    /// ///
```
moloko: might as well just delete this comment block I think...

src/core/js/views/questionView.js
```diff
@@ -153,58 +153,58 @@ define([
     // Blank method used just like postRender is for presentational components
     onQuestionRendered: function() {},
 
-    //////
+    /// ///
```
moloko: this one's a bit pointless too

brian-learningpool:
Thanks for the review @moloko.  I'd imagine our rules will evolve but I think the first pass is to get _something_ in there, then refine it to suit our needs.

package.json
```diff
@@ -44,5 +45,14 @@
         "time-grunt": "^2.0.0",
         "underscore": "^1.9.1",
         "underscore-deep-extend": "^1.1.5"
+    },
+    "devDependencies": {
```
oliverfoster: Are Adapt Framework node modules installed in production mode in the AAT?

brian-learningpool: They appear to be:
https://github.com/adaptlearning/adapt_authoring/blob/master/lib/installHelpers.js#L666

brian-learningpool:
@moloko we're now back to old style multiline ternary statements. 

@oliverfoster we're down to two lint errors.  Is there anything you'd like to change about these or should I add an ignore?
```
/adapt_framework/grunt/tasks/help.js
  62:12  error  Expected a conditional expression and instead saw an assignment  no-cond-assign

/adapt_framework/src/core/js/a11y.js
  107:29  error  Expected a conditional expression and instead saw an assignment  no-cond-assign
```

package.json
oliverfoster: Cool. ta.

oliverfoster:
Both of those lines seem reasonable-ish. If they'd work with wrapping brackets I think I'm happy.
https://eslint.org/docs/rules/no-cond-assign
:+1:

Issue #2469, opened 2019-07-02
Content objects that are children of content objects not having _isLocked set.


jamesrea83:
### Subject of the issue/enhancement/features
I have a course which uses a content object as a menu, with a number of content objects set as its children.

The menu content object and its children content objects are all set to be locked until an intro page is completed. This intro page has a quickNav component.

If the quickNav next button is set to navigate to the menu content object, as expected the button is disabled until the intro page has completed. However if it is pointed at one of the child content objects, the button is active before the intro page is completed, and the course is allowed to navigate there.

It seems that `setCustomLocking` in `adaptModel.js` loops through the direct children of the course and sets them all to locked, but it doesn't do the same for any further descendants.

### Your environment
* FW v4.1.1

PR #2470, opened 2019-07-02
setCustomLocking will now check for descendant content objects

jamesrea83:
in reference to #2469


oliverfoster:
Original function just checks the immediate children using `this.shouldLock` (which is a help that doesn't care about its own instance).
```js
        setCustomLocking: function() {
            var children = this.getAvailableChildModels();

            for (var i = 0, j = children.length; i < j; i++) {
                var child = children[i];

                child.set("_isLocked", this.shouldLock(child));
            }
        },
```
You're recursively calling the `this.setCustomLocking` function on this parent instance with the immediate children's children.
```js
        setCustomLocking: function(childModels) {
            var children = childModels || this.getAvailableChildModels();

            for (var i = 0, j = children.length; i < j; i++) {
                var child = children[i];

                child.set("_isLocked", this.shouldLock(child));

                if (child.get('_type') === 'menu') this.setCustomLocking(child.getAvailableChildModels());
            }
        },
```
Would not calling it on the immediate child work just as well as `this.shouldLock` is instance agnostic?
```js
        setCustomLocking: function() {
            var children = this.getAvailableChildModels();

            for (var i = 0, j = children.length; i < j; i++) {
                var child = children[i];

                child.set("_isLocked", this.shouldLock(child));

                if (child.get('_type') === 'menu') child.setCustomLocking();
            }
        },
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "adapt_framework",
-    "version": "4.2.0",
+    "version": "4.2.1",
```
oliverfoster: we'd normally do this when the fw is released.

src/core/js/models/adaptModel.js
```diff
@@ -719,6 +719,8 @@ define([
                 var child = children[i];
 
                 child.set("_isLocked", this.shouldLock(child));
+
+                if (child.get('_type') === 'menu') child.setCustomLocking();
```
oliverfoster: I've had another thought on this today (apologies).

### Assumption
Works for only course or menu contentObjects with child menus.
The children will have custom locking if the parent has custom locking.

### Issue
The child menu might have an alternative locking type.

### Suggestion
Switch from `child.setCustomLocking();` to `child.checkLocking();`

```suggestion
                if (child.get('_type') === 'menu') {
                    child.checkLocking();
                }
```

moloko:
@jamesrea83 you have a conflict I'm afraid...

PR #2475, opened 2019-07-11
Amend to fix #2474

### LINKED ISSUE #2474: "soft" _isResetOnRevisit breaks questions
Questions set to use `"_isResetOnRevisit": "soft"` within the components.json, cannot be answered. The `_buttonState` is undefined and causes an error, and the question doesn't think its interactive, so items can't be selected.

danielghost:
QuestionModel.reset was never being called.

`_userAnswer` isn't reset via `resetQuestionOnRevisit` but that doesn't directly affect this particular issue.

### DIFF
--- a/src/core/js/models/itemsQuestionModel.js
+++ b/src/core/js/models/itemsQuestionModel.js
@@ -161,7 +161,7 @@ define([
 
         reset: function(type, force) {
             QuestionModel.prototype.reset.apply(this, arguments);
-            ItemsComponentModel.prototype.reset.call(this, type, force);
+            ItemsComponentModel.prototype.reset.apply(this, arguments);
         },
 
         getInteractionObject: function() {
--- a/src/core/js/models/itemsQuestionModel.js
+++ b/src/core/js/models/itemsQuestionModel.js
@@ -160,7 +160,7 @@ define([
         },
 
         reset: function(type, force) {
-            QuestionModel.prototype.reset.call(this, type, force);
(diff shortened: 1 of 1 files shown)


panthimanshu79:
It indeed solve my problem too

panthimanshu79:
+1

src/core/js/models/itemsQuestionModel.js
```diff
@@ -159,6 +159,11 @@ define([
             this.set('_isAtLeastOneCorrectSelection', false);
         },
 
+        reset: function(type, force) {
+            QuestionModel.prototype.reset.call(this, type, force);
```
oliverfoster: `.apply(this, arguments);` is a bit more future-safe as it forwards all supplied arguments without bias.

moloko:
```suggestion
            QuestionModel.prototype.reset.apply(this, arguments);
```

src/core/js/models/itemsQuestionModel.js
```diff
@@ -159,6 +159,11 @@ define([
             this.set('_isAtLeastOneCorrectSelection', false);
         },
 
+        reset: function(type, force) {
+            QuestionModel.prototype.reset.apply(this, arguments);
+            ItemsComponentModel.prototype.reset.call(this, type, force);
```
moloko:
```suggestion
            ItemsComponentModel.prototype.reset.apply(this, arguments);
```

PR #2482, opened 2019-07-29
Store last selected theme preset id

### LINKED ISSUE #2481: (For AT) Should be able to store last selected theme preset
See https://github.com/adaptlearning/adapt_authoring/issues/2379 for full info.

The AT needs to store the last selected theme preset id so that the preset dropdown can be correctly selected when the user returns to the edit theme page.

This should not be merged until the corresponding AT release is out.

canstudios-nicolaw:
Fixes #2481 

Do not merge until corresponding AT release has happened

### DIFF
--- a/src/core/schema/course.model.schema
+++ b/src/core/schema/course.model.schema
@@ -553,6 +553,9 @@
     "themeVariables": {
       "type": "object"
     },
+    "lastSelectedThemePresetId": {
+      "type": "string"
+    },
     "menuSettings": {
       "type": "object"
     },


canstudios-nicolaw:
Just tested this and it looks to me like nothing gets saved to _themePreset

What do you think _themePreset should be? It seems to me as though it should just be the last selected preset id...I can look into saving it here instead - then we won't need a framework change?

canstudios-nicolaw:
Closing as this is not necessary - we will use existing _themePreset

PR #2485, opened 2019-07-29
issue/2484 allow overridden jQuery.fn.offset to set

### LINKED ISSUE #2484: jQuery.fn.offset override doesn't set only gets
### Subject of the issue

jQuery.fn.offset override doesn't set only gets
https://api.jquery.com/offset/#offset2 < defined set behaviour
https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/scrolling.js#L53 < override location

oliverfoster:
fixes #2484 
* added set behaviour


catchshiv:
+1

PR #2490, opened 2019-07-31
Resolves #2480, adds a guard on '_items' property

### LINKED ISSUE #2480: Defensive checks required around setUpItems()
This is a bit of an edge case, but if an accordion component is added which does not contain any _items, the `setUpItems` function here (https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/models/itemsComponentModel.js#L24) fails and never runs for any subsequent components.  As such the `toJSON()` call in the same code fails because `_children` is null and prevents the course from rendering.

brian-learningpool:
If '_items' was not set this would throw an error which would prevent the content from loading.

### DIFF
--- a/src/core/js/models/itemsComponentModel.js
+++ b/src/core/js/models/itemsComponentModel.js
@@ -21,7 +21,7 @@ define([
         },
 
         setUpItems: function() {
-            var items = this.get('_items') || [];
+            var items = this.get('_items') || []; // see https://github.com/adaptlearning/adapt_framework/issues/2480
             items.forEach(function(item, index) {
                 item._index = index;
             });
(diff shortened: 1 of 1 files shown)


src/core/js/models/itemsComponentModel.js
```diff
@@ -21,7 +21,7 @@ define([
         },
 
         setUpItems: function() {
-            var items = this.get('_items');
+            var items = this.get('_items') || [];
```
moloko:
```suggestion
            var items = this.get('_items') || []; // see https://github.com/adaptlearning/adapt_framework/issues/2480
```

PR #2499, opened 2019-08-08
Resolves #2498, move styles from drawer to drawer-holder

### LINKED ISSUE #2498: Drawer: move scrolling from "drawer" to "draw-holder"
Currently, when the drawer contains a long list of items that extends beyond the viewport, the drawer-toolbar scrolls with the items. Result: Close button (and Back button) scroll out of sight.

I'd like to see the drawer-toolbar better imitate the behavior of a navbar and remain fixed in position. Currently the entire drawer scrolls by default. Can this behavior be moved/restricted to drawer-holder since that is the div that receives the various drawer extensions e.g., resources, search, glossary.

### My environment
* framework v4.2

### Steps to reproduce
1 Add a lot of items to Resources
2 Scroll
3 Observe drawer-toolbar

### Expected behaviour
I'd like to see the drawer-toolbar remain fixed while the Resource items scroll behind it.

### Actual behaviour
When scrolling a long list of Resources or Glossary items, the entire drawer scrolls.

chucklorenz:
#2498 
This change prevents the drawer-toolbar from scrolling with long lists. It remains in position similar to a navbar.

Currently, drawer-toolbar has a transparent background. Because I saw no Less variables in base.less for the drawer, I added a necessary default background in Vanilla. Reference the [associated PR #234](https://github.com/adaptlearning/adapt-contrib-vanilla/pull/234)


src/core/less/base.less
```diff
@@ -65,10 +65,13 @@
         right:inherit;
         left:-320px;
     }
+    z-index:501;
+    .drawer-holder {
+        height: 100%;
```
Link2Twenty: When moving height into `drawer-holder` we lose the 100% height on the drawer.

moloko:
@chucklorenz you've accidentally committed package-lock.json and course.json

src/core/less/base.less
chucklorenz: Not sure that's correct. `drawer` maintains a rule for height: 100% at line 61. And shows up at run time in my dev tools. Am I misunderstanding?

Link2Twenty:
@chucklorenz I've removed my review you're right, I miss read the code when I was looking over it 🙂

chucklorenz:
@Link2Twenty no worries. Still, appreciate you taking time to look at it. I work alone too often so I need others to keep me sharp (or to grow sharper!).

guywillis:
v5 version of this PR has been merged into master. There were a couple of bug fixes for that PR that should make their way into the v4 version. They are:

* The bug in Windows where the scrollbar disappears beneath the `.drawer-holder` - fixed by moving the `padding-top: @navigation-height;` onto the `.drawer-inner`
* The bug with this current implementation where the drawer will open part way down where you left off - fixed by adding a scrollTop to the `.drawer-holder` in the `showDrawer` function

oliverfoster:
is this still needed?

oliverfoster:
@guywillis ?

guywillis:
This PR is an enhancement to the drawer scrolling functionality. 

It is not specifically needed.

moloko:
sure, but is it _wanted_? it's been stuck like this now for more than a year so we either need to carry it forward to completion or close it...

guywillis:
I'm happy to leave as is without merging. The requested amends are implemented in v5

PR #2552, opened 2019-10-18
Amends #2392: _htmlClassName

chucklorenz:
~This PR changes `_htmlClassName` to `_htmlClasses`.~ It clarifies some Titles in various schemas and adds some Help texts. 

Reference https://github.com/adaptlearning/adapt_framework/issues/2392


chucklorenz:
Updated this PR to return to use of `_htmlClassName` so that the amends to Titles and Help could continue toward master. I'll create another PR for `_htmlClasses` against release/v5.

Link2Twenty:
One thing to mention, developers used to JSX and React will already be used to classes being called className. If we go down the path of using React for the front end it may not be necessary to change the property name.

PR #2555, opened 2019-10-18
change _htmlClassName to _htmlClasses

chucklorenz:
The content of this PR was originally part of https://github.com/adaptlearning/adapt_framework/pull/2552. It was separated out so that it could be included as a part of v5 rather than master.
 
Reference #2392


src/course/en/course.json
```diff
@@ -1,7 +1,7 @@
 {
     "_id": "course",
     "_type": "course",
-    "_htmlClassName": "",
+    "_htmlClasses": "",
```
moloko: I think you added `"_classes": "",` here (as well) in your v4 PR

moloko: ah and of course you shouldn't need to here as well cause the v4 master branch should get merged into this one sorry

chucklorenz: yes I did; but I left that in https://github.com/adaptlearning/adapt_framework/pull/2552. I did not duplicate those additions here.

PR #2556, opened 2019-10-21
remove deprecated functions

moloko:
* findDescendants
* getAvailableChildren
* getParents

As have all been deprecated since v2.2


moloko:
> 
> 
> Didn't fancy getting rid of `setOptional()` at the same time?

@tomgreenfield I did consider it but decided that, since it has only been deprecated since v3.2.3 it would be more appropriate to leave it in place until v6. It's also only a tiny bit of code!

What I might do is update it to warn via the console that is deprecated as I notice it doesn't have that

Issue #2558, opened 2019-10-22
Loading enhancements


oliverfoster:
### Subject of the issue
1. Allowing extensions to halt execution until loaded
2. Stopping inview from processing until content object is ready

PR #2559, opened 2019-10-22
issue/2558 Added contentObject ready wait queue and inview ready lock

oliverfoster:
#2558 
* Added a wait queue around content object ready functions so that extensions can hold the loading screen until they are rendered correctly
* Lock inview until content object is ready so that inview doesn't trigger on initial text blocks until ready

### DIFF
--- a/src/core/js/views/menuView.js
+++ b/src/core/js/views/menuView.js
@@ -19,6 +19,7 @@ define([
     	},
 
         preRender: function() {
+            $.inview.lock('menuView');
             this.disableAnimation = Adapt.config.has('_disableAnimation') ? Adapt.config.get('_disableAnimation') : false;
             this.$el.css('opacity', 0);
             this.listenTo(this.model, 'change:_isReady', this.isReady);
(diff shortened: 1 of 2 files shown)


src/core/js/views/menuView.js
```diff
@@ -28,26 +29,33 @@ define([
...
+
+            var performIsReady = function() {
+                $('.loading').hide();
+                $(window).scrollTop(0);
+                Adapt.trigger('menuView:ready', this);
+                $.inview.unlock('menuView');
+                var styleOptions = { opacity: 1 };
+                if (this.disableAnimation) {
+                    this.$el.css(styleOptions);
+                    $.inview.unlock('menuView');
```
cahirodoherty-learningpool: You are calling this twice

oliverfoster: I'd misplaced it originally. Good catch.

src/core/js/views/pageView.js
```diff
@@ -22,32 +22,39 @@ define([
...
+
+            var performIsReady = function() {
+                $('.loading').hide();
+                $(window).scrollTop(0);
+                Adapt.trigger('pageView:ready', this);
+                $.inview.unlock('pageView');
+                var styleOptions = { opacity: 1 };
+                if (this.disableAnimation) {
+                    this.$el.css(styleOptions);
+                    $.inview();
```
moloko:
```suggestion
                    $.inview();
                    $(window).scroll();
                    return;
```
then you can lose the `else` block

src/core/js/views/menuView.js
```diff
@@ -28,26 +29,32 @@ define([
...
+
+            var performIsReady = function() {
+                $('.loading').hide();
+                $(window).scrollTop(0);
+                Adapt.trigger('menuView:ready', this);
+                $.inview.unlock('menuView');
+                var styleOptions = { opacity: 1 };
+                if (this.disableAnimation) {
+                    this.$el.css(styleOptions);
+                    $.inview();
```
moloko:
```suggestion
                    $.inview();
                    $(window).scroll();
                    return;
```
and drop the `else`?

Issue #2563, opened 2019-10-28
findRelativeModel: isn't working correctly when search for descendants


oliverfoster:
### Subject of the issue
Finding the next block from an article currently finds the first block in the article and not the first block in the next article as it should.

PR #2564, opened 2019-10-28
issue/2563 fixed findRelativeModel for descendent searches

oliverfoster:
#2563 
* findRelativeModel finds in the next parent on descendent type searches instead of finding in parent

This is an issue when using trickle on article with @block +1

### DIFF
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -448,6 +448,11 @@ define([
 
             var findDescendantType = (!findSiblingType && !findAncestorType);
 
+            if (findDescendantType) {
+                // move by one less as first found is considered next
+                moveBy-=1;
+            }
+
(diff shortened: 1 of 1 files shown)


brian-learningpool:
Does this require backporting?

oliverfoster:
Yes @brian-learningpool, i'll do it.

src/core/js/models/adaptModel.js
```diff
@@ -448,6 +448,11 @@ define([
 
             var findDescendantType = (!findSiblingType && !findAncestorType);
 
+            if (findDescendantType) {
+                // move by one less as first found is considered next
+                moveBy-=1;
```
moloko:
```suggestion
                moveBy -= 1;
```
or perhaps?
```suggestion
                moveBy--;
```

Issue #2544, opened 2019-10-09
new a11y API breaks adapt-contrib-languagePicker


moloko:
The new a11y API is causing a rendering issue and a script error when the languagePicker 'drawer' view is used

Using FW master branch and languagePicker v3.0.1

![image](https://user-images.githubusercontent.com/1249597/66500387-5789aa80-eab9-11e9-920e-6508a1e3e2f7.png)

Script error is triggered when attempting to select a language via the drawer view

PR #2566, opened 2019-10-28
issue/2544 Added method to allow open popup return focus to be set

oliverfoster:
#2544 
* Added a formalised way to change the return focus of a popup

### DIFF
--- a/src/core/js/a11y.js
+++ b/src/core/js/a11y.js
@@ -731,6 +731,18 @@ define([
         popupClosed: function($focusElement) {
             this._popup.closed($focusElement);
             return this;
+        },
+
+
+        /**
+         * When a popup is open, this function make it possible to swap the element
+         * which should receive focus on popup close.
+         *
+         * @param {Object} $focusElement Set a new element to focus on.
+         * @returns {Object} Returns previously set focus element.
+         */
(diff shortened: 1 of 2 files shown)


src/core/js/a11y.js
```diff
@@ -731,6 +731,17 @@ define([
         popupClosed: function($focusElement) {
             this._popup.closed($focusElement);
             return this;
+        },
+
+        /**
+         * When a popup is open, this function make it possible to swap the element
+         * which should receive focus on popup close.
```
tomgreenfield: `make → makes` & `which` → `that`.

src/core/js/a11y/popup.js
```diff
@@ -203,6 +203,19 @@ define([
                 }
             }.bind(this));
             return (document.activeElement = this._focusStack.pop());
+        },
+
+        /**
+         * When a popup is open, this function make it possible to swap the element
+         * which should receive focus on popup close.
```
tomgreenfield: `make → makes` & `which` → `that`.

moloko: grammar 👮

moloko: ;-)

Issue #2593, opened 2019-11-20
Incorrect Spoor dependency


tomgreenfield:
https://github.com/adaptlearning/adapt_framework/blob/2e920997591dceaed2f74b07a415511ddfa9c0f5/adapt.json#L23

This version doesn't exist and so the authoring tool gives this output:
```console
Failed to install adapt-contrib-spoor, Error: No tag found that was able to satisfy >=4
```
I would suggest that all dependencies can be switched back to `"*"` – the CLI more than adequately handles installing the latest compatible versions of plugins. This would irradicate the overhead of manually updating the ranges in `adapt.json`, which are frequently incorrect/incomplete.

moloko:
ah, this one's my bad - I realised the other day that spoor didn't need a major version bump as all that had really changed was indentation/whitespace... so changed the version number to 3.3.2 instead of 4.0.0 - and completely forgot that there was a reference to it in adapt.json!

> I would suggest that all dependencies can be switched back to `"*"`

yep if that works then it's definitely a better way to go

PR #2594, opened 2019-11-21
Updated dependencies to use an asterisk

guywillis:
Fixes https://github.com/adaptlearning/adapt_framework/issues/2593


moloko:
do we need to do something about the v4 branch as well?

moloko:
> do we need to do something about the v4 branch as well?

wrt to this: https://community.adaptlearning.org/mod/forum/discuss.php?d=3742

tomgreenfield:
I proposed patching the authoring tool [here](https://github.com/adaptlearning/adapt_authoring/issues/2479) but happy to do either/both.

PR #2616, opened 2019-12-19
added 'returnToStartLocation' to startController

### LINKED ISSUE #2541: startController: add 'return to start location' feature
Most of the courses that we build which include an 'intro page' (or pages) have some kind of 'back to intro' button contained in them.

It would be handy if the `startController` module offered some way of handling much of the logic for you, something like:
```js
Adapt.startController.returnToStartLocation(alwaysForce:Boolean, dontSkipCompleted:Boolean);
```

moloko:
Also added an event handler so that you can easily add a 'return to start' button with attribute `data-event="returnToStart"` to nav.hbs to get 'return to start' functionality

(As a next step, it'd be a good idea to include a 'show return to start button' as a configurable option in the `_start` configuration )

resolves #2541

(also changed double quotes to single)

### DIFF
--- a/src/core/js/startController.js
+++ b/src/core/js/startController.js
@@ -7,12 +7,23 @@ define([
     model: null,
 
     loadCourseData: function() {
-      this.model = new Backbone.Model(Adapt.course.get("_start"));
+      this.model = new Backbone.Model(Adapt.course.get('_start'));
     },
 
     setStartLocation: function() {
       if (!this.isEnabled()) return;
       window.history.replaceState('', '', this.getStartHash());
+    },
+
+    returnToStartLocation: function() {
+      var startIds = this.model.get('_startIds');
+      if (startIds) {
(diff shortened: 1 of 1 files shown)


moloko:
@cahirodoherty-learningpool @oliverfoster sorry couldn't help myself, would you mind reviewing again!

src/core/js/startController.js
```diff
@@ -7,74 +7,75 @@ define([
...
-          }
-
-          if (!className || $html.is(className) || $html.hasClass(className)) {
-            startId = item._id;
-            break;
-          }
+      var startId = this.model.get('_id');
+      var startIds = this.model.get('_startIds');
+
+      if (!startIds) return startId;
```
oliverfoster:
```suggestion
      if (!startIds ||  !startIds.length) return startId;
```
It may be declared as an empty array.

PR #2624, opened 2020-01-13
Added an event handler to easily add a 'return to start' button

cahirodoherty-learningpool:
Patching changes from https://github.com/adaptlearning/adapt_framework/pull/2616

resolves [#2541](https://github.com/adaptlearning/adapt_framework/issues/2541)

### DIFF
--- a/src/core/js/startController.js
+++ b/src/core/js/startController.js
@@ -1,93 +1,102 @@
 define([
-    'core/js/adapt'
+  'core/js/adapt'
 ], function(Adapt) {
-    
-    var StartController = Backbone.Controller.extend({
 
-        model: null,
+  var StartController = Backbone.Controller.extend({
 
-        initialize: function() {
-            this.model = new Backbone.Model(Adapt.course.get("_start"));
-        },
+    model: null,
 
-        setStartLocation: function() {
-            if (!this.isEnabled()) return;
-            window.history.replaceState('', '', this.getStartHash());
-        },
+    initialize: function() {
+      this.model = new Backbone.Model(Adapt.course.get('_start'));
+    },
 
-        getStartHash: function(alwaysForce) {
-            var startId = this.getStartId();
+    setStartLocation: function() {
+      if (!this.isEnabled()) return;
+      window.history.replaceState('', '', this.getStartHash());
+    },
 
-            var hasStartId = (startId)
-                ? true
-                : false;
+    returnToStartLocation: function() {
+      var startIds = this.model.get('_startIds');
+      if (startIds) {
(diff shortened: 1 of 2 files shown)


lc-alexanderbenesch:
Tested

It didn't work on my machine, since it ran into a TypeError on ```Adapt.course.get('_start')``` in the initialize function, saying Adapt.course is undefined.

I had to change the following code to make it work
```
Adapt.once('adapt:start', function() {
    Adapt.startController.initialize();
    Adapt.startController.setStartLocation();
});
```
to
```
Adapt.once('adapt:start', function() {
    Adapt.startController = new StartController();
    Adapt.startController.setStartLocation();
});
```
as well as

```
// ...
return (Adapt.startcontroller = new StartController());
```
to
```
// ...
return StartController;
```

I realize my changes are basically reverting something that you got from #2616 so maybe there's something else I'm missing?

moloko:
@lc-alexanderbenesch thank you for taking the time to test this. There were a lot of changes made to the 'entry point' code in Adapt v4.4.1 that may mean this code needs a little bit more 'finessing' to get it to work with the older code in Adapt v3. 

@cahirodoherty-learningpool could you take a look at this when you have a moment?

src/core/js/startController.js
```diff
@@ -1,93 +1,102 @@
...
 
-    Adapt.once("adapt:start", function() {
-        var startController = new StartController();
-        startController.setStartLocation();
-    });
+  /*
+  * allows you to call returnToStartLocation either by calling `Adapt.trigger('navigation:returnToStart')`
+  * or by including in the top navigation bar a button that has the attribute `data-event="returnToStart"`
+  */
+  Adapt.on('navigation:returnToStart', function() {
```
oliverfoster: We should not use events like this any more. As this is not a state change but an action it should just have a defined API. `Adapt.startController.returnToStartLocation()` should suffice here.

moloko: @oliverfoster this was actually my doing originally - see #2616 

appreciate what you're saying about how this would be better done in navigationView.js - might be worth also moving some of what's in router.js to navigationView.js for consistency? Specifically the 'navigation' events here:
https://github.com/adaptlearning/adapt_framework/blob/6e1c7e4939fe25c9114be2f9af3b09c69b6d24dd/src/core/js/router.js#L26-L32

oliverfoster: Absolutely.

oliverfoster: done in https://github.com/adaptlearning/adapt_framework/pull/2708

cahirodoherty-learningpool:
Closing

Issue #2625, opened 2020-01-13
$.imageReady error in IE 11 + SVG


aut0poietic:
### Subject of the issue/enhancement/features
imageReady fails when it encounters an SVG on IE11.

### Your environment
* 4.3
* InternetExplorer 11
* Windows 10

### Steps to reproduce
This error can only be encountered when a Custom Plugin/Menu/Theme has an inline SVG.

### Expected behaviour
imageReady should skip the SVG as it does not have a `currentStyle` property.

### Actual behaviour
imageReady throws an error on [line 44](https://github.com/adaptlearning/jquery.imageready/blob/a4a320f3ce8a43c3967f2dd1a98df426f81cff86/jquery.imageready.js#L44) "Unable to get property 'backgroundImage' of undefined or null reference.'

imageReady never completes for the component, and the course never finishes loading the page.

### [Proposed Fix](https://github.com/adaptlearning/jquery.imageready/blob/a4a320f3ce8a43c3967f2dd1a98df426f81cff86/jquery.imageready.js#L40-L49)

The issue appears to be that SVG's on IE11 do not have a `currentStyle` property. I check to ensure `el.currentStyle` is valid should solve the problem, right? 

```
...
            for (var i = 0, len = tags.length; i < len; i++) {
                el = tags[i];  
                if (!el.currentStyle) continue;
                var hasNoValue = (el.currentStyle[scriptName] == 'none');
                if (hasNoValue) continue;
                rtn.push(el);
            }
...
```

oliverfoster:
Sounds like a good solution. Could you PR?

aut0poietic:
Sure can. I'll take it up tonight (day job's in the way today). Should that PR be for the [jquery.imageReady](https://github.com/adaptlearning/jquery.imageready) repo or this one?

oliverfoster:
It's alright, I did it already.

oliverfoster:
The [other repo](https://github.com/adaptlearning/jquery.imageready) is a bit defunct / out of date / unmaintained unfortunately.

moloko:
@aut0poietic if you have a moment could you review #2626 and a comment of '+1' if you're happy with it?

aut0poietic:
Done! Thanks @oliverfoster, @moloko!

PR #2626, opened 2020-01-14
issue/2625 Fix for ie11 svg

oliverfoster:
#2625 
* Prevented issue where no 'currentStyle' property exists on SVGs


aut0poietic:
👍

Issue #2642, opened 2020-02-10
tslib not found


oliverfoster:
### Subject of the issue
Current framework isn't compiling LESS properly due to upstream bug with less 3.11.0

https://github.com/less/less.js/issues/3469

![image](https://user-images.githubusercontent.com/7974663/74155797-0e836780-4c0d-11ea-8da3-bbe9f654020c.png)

Confirmed working with 3.10.3

brian-learningpool:
In the meantime should the `less` version in `package.json` be set to 3.10.3 exactly?  We should probably consider using package.lock going forward

oliverfoster:
Seems like a sensible suggestion.

References: 
* https://docs.npmjs.com/configuring-npm/package-locks.html
* https://docs.npmjs.com/configuring/npm/package-lock-json

oliverfoster:
The above PR is only going to work with master atm and the package-lock will need to be remade after every version number change just before release.

moloko:
@oliverfoster that second link should be https://docs.npmjs.com/configuring-npm/package-lock-json.html I think

moloko:
> the package-lock will need to be remade after every version number change just before release

How's that done?

oliverfoster:
`$ npm install --package-lock-only` will ensure the lock file reflects the `package.json`

PR #2643, opened 2020-02-10
fixes issue/2642 adds package-lock with less 3.10.3

oliverfoster:
#2642 
* Add package-lock with fixed LESS version to 3.10.3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,3 @@ bower_components
 node_modules
 .DS_Store
 .idea
-package-lock.json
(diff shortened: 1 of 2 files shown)


cahirodoherty-learningpool:
@oliverfoster Should this include the package.json now that LESS has been updated there?

oliverfoster:
Nah, this package-lock file will need to be overwritten before the next release anyway. You can see this pr more as an acceptance of the package-lock prinicple rather than approval for what's in the package-lock file.

PR #2648, opened 2020-02-17
issue/2647 add es6-8 support

### LINKED ISSUE #2647: Add es6 support
* Add babel es6+ to es5 transpile
* Add polyfill for IE11

oliverfoster:
fixes #2647 

This pr allows us to use a large amount more ES6, ES7 and ES8, such as: classes to replace and extend from backbone ones, arrow functions, spread operators, async/await, Object.assign, Array.filter etc. all of which should work in IE11 given these changes.

### Limitations
We're still using requirejs to bundle modules so we **won't** be able to use ES6 module import/export directives yet.
As requirejs uses [Esprima](https://github.com/jquery/esprima/blob/master/README.md) to turn JavaScript into AST, the compilation process is limited to whatever [Esprima can parse](https://github.com/jquery/esprima/issues), it claims full support for ES8. This discounts some newer features of ECMAScript, such as [ES9 object property spreading](https://nitayneeman.com/posts/object-rest-and-spread-properties-in-ecmascript-2018/):
```js
var object1 = { a: 1 };
var object2 = { b: 2, ...object1 };
```
requirejs will error at compile time with warnings for unsupported syntax, like so:
image

Function argument spreading works as expected.
```js
function test(a, b, c) {
   debugger;
}
test(...[1, 2, 3])
```

### Changes
* created a compilation temp folder 
  * to provide a space for interim asset production
* changed requirejs javascript compiler
  * to output to temp folder
  * don't minify (would use uglify which isn't es6 compatible)
* added babel transpiler task to grunt
  * take bundled javascript from temp folder 
  * import and modify sourcemaps if available (so that the es6 sourcemaps map to minified es5 output)
  * transpile to es5
  * minify output
  * put output in outputdir/adapt/js/ as normal
* added clean task to remove temp folder after compilation
* updated backbone to v1.4.0
  * this adds a `preinitialize` function to model, view, collection, etc
  * updated backbone.controller to add `preinitialize` function
  * `preinitialize` allows es6 classes to setup events, defaults etc before the backbone `initialize` function calls
* added polyfill for ie11
* removed redundant polyfill for promises
* merged AdaptModel ES6 PR #2652 

### Notes
I was thinking we could add this to version 5 to provide some backwards compatibility for plugins which will be updated in v6+.
If we convert backbone classes to es6 classes in the core (we definitely should do this), then we will need to define [constructor properties](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/componentView.js#L89) as static getters:
```js
static get type() {
  return 'component';
}
```

### Testing
The PR https://github.com/adaptlearning/adapt_framework/pull/2651 has been accepted and merged into this one such that this PR contains client-side ES6-8 and updated build tools.


chucklorenz:
tested various (easy) features from ES6-ES8: arrow function, array.filter, Object.assign, Object.values, padStart, async/await. All successful.
👍

guywillis:
👀

moloko:
can we hold off on merging this and #2652 until I've had a chance to do a release of what's currently in master (plus maybe one or two other things)?

oliverfoster:
of course, we still don't have the requisite +1s, so you should be safe regardless

oliverfoster:
Awaiting a framework release and a few more reviews - just to be safe.

brian-learningpool:
Hi @oliverfoster, why exactly does `CourseModel` inherit `MenuModel`?

oliverfoster:
@brian-learningpool because it's a menu and a contentobject, as well as the root container of the course content, the home of the _globals namespace and the target for plugin attributes, etc. It'll make simplifying the router a little easier later. The router should just manage contentObjects + plugin routes.

oliverfoster:
This pull request is ready for final +1s

cahirodoherty-learningpool:
Hi Oliver,
Thanks for checking in, but I think we are holding you up on this unfairly, so I'd say fire away with your implementation. Brian may get in touch to discuss your changes from a curiosity perspective, but we don't want to hold you up any further.
Apologies for that.

PR #2652, opened 2020-02-19
issue/2647 Converted AdaptModel to ES Class

oliverfoster:
#2647 and #2648
* conversion of core AdaptModel to an ES6 Class


src/core/js/models/adaptModel.js
```diff
@@ -4,45 +4,51 @@ define([
...
+    }
+
+    bubblingEvents() {
+      return [
+        'change:_isComplete',
+        'change:_isInteractionComplete'
+      ];
+    }
+
+    initialize () {
```
brian-learningpool: Minor: space after function name.

src/core/js/models/adaptModel.js
```diff
@@ -555,38 +558,32 @@ define([
...
+      const siblingsCollection = new Backbone.Collection(siblings);
       this.set('_siblings', siblingsCollection);
       return siblingsCollection;
-    },
-
-    setOnChildren: function (key, value, options) {
+    }
 
-      var args = arguments;
+    setOnChildren(key, value, options) {
```
tomgreenfield: Suggest `setOnChildren(...args)` or similar? Would avoid the linter complaining about variables that are never used.

src/core/js/models/adaptModel.js
```diff
@@ -555,38 +558,32 @@ define([
...
-    },
+    }
 
     /**
      * @deprecated since v3.2.3 - please use `model.set('_isOptional', value)` instead
      */
-    setOptional: function(value) {
-      Adapt.log.warn("DEPRECATED - Use model.set('_isOptional', value) as setOptional() may be removed in the future");
+    setOptional(value) {
+      Adapt.log.warn('DEPRECATED - Use model.set(\'_isOptional\', value) as setOptional() may be removed in the future');
```
tomgreenfield: Suggest template literal to irradicate need for escaping:
```suggestion
      Adapt.log.warn(`DEPRECATED - Use model.set('_isOptional', value) as setOptional() may be removed in the future`);
```

src/core/js/models/adaptModel.js
oliverfoster: Hmmm...

PR #2655, opened 2020-02-20
#1912 expose _isResetOnRevisit to authoring tool.

### LINKED ISSUE #1912: core/schema/component.model: _isResetOnRevisit missing
The `_isResetOnRevisit` attribute doesn't seem to be exposed within the Authoring Tool. Is there a reason for this?

The need to set this at a component level is fairly uncommon, but it can be required. For example, when question responses are restored across sessions via `_shouldStoreResponses`, you may wish to disable this for particular questions or only restore responses for passed assessments and therefore need to override this via `_isResetOnRevisit` on the relevant components.

We should allow `true`, `false`, `"soft"` and `"hard"` values.

jamesrea83:
resolves #1912

### DIFF
--- a/src/core/schema/component.model.schema
+++ b/src/core/schema/component.model.schema
@@ -67,6 +67,23 @@
       "title": "Is this visible?",
       "help": "Disabling this is equivalent to applying the CSS 'visibility:hidden' to the component."
     },
+    "_isResetOnRevisit": {
+      "type": "string",
+      "default": "false",
+      "isSetting": true,
+      "inputType": {
+        "type": "Select",
+        "options": [
+          "hard",
+          "soft",
+          true,
+          false
+        ]
+      },
+      "title": "Reset on revisit",
+      "help": "Causes component to be reset if you leave the page and return to it. Soft will allow the user to do the component again, but will not require them to do. Hard / true will reset but will require the user to complete the component again.",
+      "validators": []
+    },
     "_ariaLevel": {
       "type": "number",
       "default": 0,
--- a/src/core/schema/component.model.schema
+++ b/src/core/schema/component.model.schema
@@ -74,10 +74,9 @@
       "inputType": {
         "type": "Select",
         "options": [
-          "hard",
+          "false",
           "soft",
-          true,
-          false
+          "hard"
         ]
       },
       "title": "Reset when revisited?",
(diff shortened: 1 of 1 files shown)


src/core/schema/component.model.schema
```diff
@@ -67,6 +67,23 @@
...
+    "_isResetOnRevisit": {
+      "type": "string",
+      "default": "false",
+      "isSetting": true,
+      "inputType": {
+        "type": "Select",
+        "options": [
+          "hard",
+          "soft",
+          true,
```
moloko: not sure `true` is really needed since it's equivalent to `"hard"`

src/core/schema/component.model.schema
```diff
@@ -67,6 +67,23 @@
...
+          false
+        ]
+      },
+      "title": "Reset on revisit",
```
moloko:
```suggestion
      "title": "Reset when revisited?",
```

src/core/schema/component.model.schema
```diff
@@ -67,6 +67,23 @@
...
+      "help": "Causes component to be reset if you leave the page and return to it. Soft will allow the user to do the component again, but will not require them to do. Hard / true will reset but will require the user to complete the component again.",
```
moloko:
```suggestion
      "help": "Controls whether this component will be reset when the learner leaves the page then returns to it. The 'soft' setting will reset the component to allow the learner to attempt it again, but will not require them to do so; whereas the 'hard' setting will require them to do so. The default (false) will not cause the component to be reset.",
```

tomgreenfield:
> 
> 
> @tomgreenfield is it valid to have `false` as a value here?

Doubt you can specify non-string values in a type `string` field. Equally happy for someone to prove me wrong by testing in a local instance.

moloko:
> > @tomgreenfield is it valid to have `false` as a value here?
> 
> Doubt you can specify non-string values in a type `string` field. Equally happy for someone to prove me wrong by testing in a local instance.

Well, since `"hard"` and `true` are synonymous let's just remove `true`

olivermartinfoster:
https://github.com/adaptlearning/adapt_framework/blob/b6ec1d15c6d14e7f1623f9fb5dc2632fdc023a9a/src/core/js/models/adaptModel.js#L167-L187 "false" would behave as intended here

Issue #2623, opened 2020-01-10
adapt-contrib-trickle: "_isAvailable": false on an article causes error


RobertPeek:
If an article is set to "_isAvailable": false, trickle breaks and you cannot trickle through a page.

### Environment
Framework v4.3.0
Latest Chrome
Windows 10

### Steps to reproduce
Set up a page with 3 articles, each with trickle configured on the article with "_scrollTo": "@article +1"

Set the second article to "_isAvailable": false

Run course and view the page.

### Expected behaviour
Trickle should move through the 2 'available' articles.

### Actual behaviour
Clicking Trickle on the 1st article does nothing and the console error is shown

![image](https://user-images.githubusercontent.com/7251995/72151341-df749e80-339f-11ea-9f02-0787178b8c45.png)

jamesrea83:
The quick fix for this one is to make sure the child blocks are also set to `_isAvailable: false`.

Would it be worth having a feature that, upon finding an `_isAvailable: false` flag, automatically sets all applicable child elements as well? Is there a scenario where you would want an available block contained in an unavailable article?

moloko:
> Is there a scenario where you would want an available block contained in an unavailable article?

@jamesrea83 we have found previously that cascading changes from parent to children can be problematic in unexpected ways. I think @oliverfoster has an alternative technique for addressing this particular issue from within trickle.

oliverfoster:
Add something like this to `adaptModel.js`:
```js
  /**
   * Derives isAvailable from parent stack
   */
  getIsAvailable: function() {
    var isAvailable = this.get('_isAvailable');
    if (!isAvailable) {
      return false;
    }
    var parent = this;
    while (parent = parent.getParent()) {
      isAvailable = parent.get('_isAvailable');
      if (!isAvailable) {
        return false;
      }
    }
    return true;
  }
```
Then use in trickle.

akshay-bhavsar:
Please neglect above comment and refer PR: https://github.com/adaptlearning/adapt_framework/pull/2665 for this issue

moloko:
this will need fixing in Adapt v5.x/Trickle v4.x as well I think?

moloko:
> 
> 
> Add something like this to `adaptModel.js`:
> 
> ```js
>   /**
>    * Derives isAvailable from parent stack
>    */
>   getIsAvailable: function() {
>     var isAvailable = this.get('_isAvailable');
>     if (!isAvailable) {
>       return false;
>     }
[...]

PR #2665, opened 2020-02-28
issue/2623 - fixes for scrolling in trickle

akshay-bhavsar:
#2623

### DIFF
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -58,8 +58,6 @@ define([
             _.defer(function() {
                 if (this._children) {
                     this.checkCompletionStatus();
-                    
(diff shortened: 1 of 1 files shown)


oliverfoster:
We did something similar for property cascading to children once for _isOptional and it did go very well. This is a bad idea.

If someone wants to hide a page from a menu they will tend to use `_isAvailable: false` but they won't want the children to be `_isAvailable: false` because they're probably rendering the page, but not from the menu.

A better solution would be to if something is `_isAvailable` in its hierarchy by checking its parents recursively, rather than cascading the properties.

akshay-bhavsar:
PR: https://github.com/adaptlearning/adapt-contrib-trickle/pull/105

For changes need in Trickle

PR #2674, opened 2020-03-04
Converted child models of AdaptModel to ES6 Class

NayanKhedkar:
#2647 
* Converted CourseModel, ContentObjectModel, ArticleModel, BlockModel and ComponentModel to ES6
* Added MenuModel, PageModel which extend ContentObjectModel
* adapt.contentObjects collection creates MenuModel or PageModel when instantiating
* CourseModel now extends ContentObjectModel


oliverfoster:
Do you mind if I modify this slightly? It will help us later if we do this in a slightly different way.

NayanKhedkar:
> Do you mind if I modify this slightly? It will help us later if we do this in a slightly different way.

Yup

oliverfoster:
Thanks for doing that :+1:

Issue #2688, opened 2020-03-12
Framework - Refresh icon font set


guywillis:
### Subject of the issue
The current icon font set doesn't look like it has been updated in a while. It would be nice to have a little refresh of the current icon suite so they look a little fresher and more modern. 

### Your environment
* v5 Framework

guywillis:
Old:
<img width="1453" alt="screen-01" src="https://user-images.githubusercontent.com/11569678/76624326-fb75f700-652c-11ea-8a9c-7eddfadf24e7.png">

New:
<img width="1441" alt="screen-02" src="https://user-images.githubusercontent.com/11569678/76624329-fd3fba80-652c-11ea-89f3-ed5f9fb7032f.png">

Additional:
![screen-03](https://user-images.githubusercontent.com/11569678/76632765-9e357200-653b-11ea-80d9-a30e967af272.png)

* accessibility-2
* alert
* arrow-down-2
* arrow-down-3
* arrow-left-2
* arrow-left-3
* arrow-right-2
* arrow-right-3
* arrow-up-2
* arrow-up-3
* attachment
* block
* camera-off
* camera-on
* cloud
* compass
* copy-file
* copyright
* crop
* devices
* down-up
* ellipsis-vertical
* exclamation
* expand_1
* extension
* favourite
* google-translate
* hand
* lightning
* list-2
* microphone-off
* move
* mute
* ordered-list
* question-circle
* quote-close
* quote-open
* refresh
* send
* sound-down
* sync
* tablet
* thumb-down
* thumb-up
* timer-empty
* timer-full
* video-captions-off-2
* video-captions-off
* video-captions
* video-exit-fullscreen
* video-fast-forward
* video-fullscreen
* video-loop
* video-pause-circle
* video-pause
* video-play-circle
* video-play
* video-replay
* video-rewind
* video-skip-back
* video-skip-forward
* video-stop

guywillis:
Further additions to the font set based on PR comments:
<img width="1441" alt="screen-04" src="https://user-images.githubusercontent.com/11569678/76978508-b2f47a00-692e-11ea-9500-ef5c63419e6c.png">

* quote-close-2
* quote-open-2
* emotion-happy
* emotion-not-ok
* emotion-ok
* emotion-sad
* emotion-very-happy
* emotion-very-unhappy
* blogger-2
* blogger
* deviantart-2
* deviantart
* doc-excel
* doc-pdf
* doc-word
* facebook-2
* facebook
[...]

PR #2691, opened 2020-03-13
Refreshed existing and added new icons

guywillis:
Resolves https://github.com/adaptlearning/adapt_framework/issues/2688

* Refreshed visual appearance of existing icons - see issue for screenshots
* Added new icons - see issue for list + screenshot
* Updated icon references


oliverfoster:
image love it, image nice refresh image a necessary piece of the puzzle image sets a nice... direction image

chucklorenz:
Thanks for the refresh, Guy.
Two proposals:
1) a second set of block quotes. The current are quite round and serif-y. Would be nice to have an option that has a more minimalist and sans-serif look.
2) social media/networking logos. Both positive and negative images per each would help design. Which ones? The popular ones are obvious candidates. How many? Hard to say, but I'd encourage erring on the side of more than less. Many are available for free on IcoMoon. These may not seem too important to e-learning, but Adapt is being used in a broad manner. I used them in [adapt-aboutUs](https://github.com/chucklorenz/adapt-aboutUs) and even received a PR from samumist (Zheng Xu) to include some that are popular in China. My [selection.json](https://github.com/chucklorenz/adapt-aboutUs/blob/master/selection.json) in case it helps discussion.

chucklorenz:
3. Wondering if we could get a series of 5 face emojis ranging from frown to smiling to use with slider/feedback situations

chucklorenz:
Thanks for the additions! 
:+1:

PR #2698, opened 2020-03-19
issue/2697 Added Backbone ES6 static property inheritance

### LINKED ISSUE #2697: Add ES6 class static property inheritance to Backbone
### Subject of the issue
Now that we have ES6 class support, there is an issue with Backbone class static properties that needs fixing.

### Generalised Issue

ES6 static property definition
```js
class ClassName {

  get prototypePropertyName() {
    return 'value';
  }

  static get staticPropertyName() {
    return 'value';
  }

}
```
Backbone static property definition
```js
var ClassName = Backbone.View({ 

  // prototype/class-instance properties
  prototypePropertyName: 'value'

}, {

  // class/static properties
  staticPropertyName: 'value'

});
```

Both styles produce the static property `ClassName.staticPropertyName === 'value` and the prototype property `var instance = new ClassName(); instance.prototypePropertyName === 'value'`.

With backbone only enumerable static property values are copied to each new sub class `var SubClassName = ClassName.extend({})`, whereas in ES6 `class SubClassName extends ClassName {} ` all static properties are inherited from the super class. Static properties in ES6 are not enumerable by default and so if used with Backbone.extend they won't get copied to the new sub class.

In Backbone and ES6 prototype/class-instance properties have an inheritance chain. In Backbone class/static-properties don't  have an inheritance chain but ES6 class/static properties do have an inheritance chain.

### Specific issue
Adapt views, such as [AdaptView](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/adaptView.js#L210), [ArticleView](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/articleView.js#L21-L24) and [ComponentView](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/componentView.js#L89), use static properties to define a variety of inheritable values. When mixing ES6 and Backbone classes, static property inheritance will fail here.

### Solution
Backbone behaves identically if it inherits super class static properties rather than copying them, so  [this PR](https://github.com/jashkenas/backbone/pull/4235) fixes the issue quite nicely whilst making backbone classes align with ES6 behaviour. We can create a polyfill library for Adapt to import this behaviour.

### Best template for class static properties
Do not use the static keyword for existing properties. This is as getters and setters change the way newly defined property values which override existing inherited property values are applied, the values will be applied through the getters and setters rather than being copied over the top of them.

Bad Example:
```js
/** 1. define getter */
class ClassName extends Backbone.View {
  static get staticPropertyName() {
    return 'value';
  }
}

/** 2. to Backbone legacy */
const ClassName2 = ClassName.extend({
}, {
  /**
   * This property value assignment will try to use the undefined setter.
  */
  staticPropertyName: 'value3'
});
```

Working Example 1:
```js
/** 1. from Backbone legacy */
const ClassName = Backbone.View.extends({
}, {
  staticPropertyName: 'value1'
});

/** 2. through ES6 */
class ClassName2 extends ClassName {
}
/**
 * Define the static property override in a way that allows further
 * Backbone extends to overwrite the value.
 */
ClassName2.staticPropertyName = 'value2';

/** 3. to Backbone legacy */
const ClassName3 = ClassName2.extend({
}, {
  staticPropertyName: 'value3'
});
```
Working Example 2:
```js
/** 1. from Backbone using ES6 */
class ClassName extends Backbone.View {
}
/**
 * Define the static property override in a way that allows further
 * Backbone extends to overwrite the value.
 */
ClassName.staticPropertyName = 'value1';

/** 2. through ES6 */
class ClassName2 extends ClassName {
}
/**
 * Define the static property override in a way that allows further
 * Backbone extends to overwrite the value.
 */
ClassName2.staticPropertyName = 'value2';

/** 3. to Backbone */
const ClassName3 = ClassName2.extend({
}, {
  staticPropertyName: 'value3'
});
```

oliverfoster:
fixes #2697 
* Added ES6 class static property inheritance to Backbone

Also merged PR #2700 Converted AdaptView to ES6

Inspired by https://github.com/jashkenas/backbone/pull/4235/files but without <IE11 support.


src/core/js/views/adaptView.js
```diff
@@ -30,185 +30,165 @@ define([
...
-      var onscreen = this.model.get('_onScreen');
+    setupOnScreenHandler() {
+      const onscreen = this.model.get('_onScreen');
 
       if (!onscreen || !onscreen._isEnabled) return;
 
       this.$el
         .addClass('has-animation')
-        .addClass(onscreen._classes + '-before');
+        .addClass(`${onscreen._classes}-before`);
```
tomgreenfield: Duplicate call to `addClass()` – can space-separate here:
```suggestion
      this.$el.addClass(`has-animation ${onscreen._classes}-before`);
```

src/core/js/views/adaptView.js
```diff
@@ -30,185 +30,165 @@ define([
...
-        });
-      }
-    },
+    }
+
+    setCompletionStatus() {
+      if (!this.model.get('_isVisible')) return;
+      this.model.set({
+        '_isComplete': true,
+        '_isInteractionComplete': true
```
tomgreenfield: Are quotes necessary here?

src/core/js/views/adaptView.js
```diff
@@ -30,185 +30,165 @@ define([
...
-    },
+      });
+    }
 
-    findDescendantViews: function(isParentFirst) {
-      var descendants = [];
-      this.childViews && _.each(this.childViews, function(view) {
+    findDescendantViews(isParentFirst) {
+      const descendants = [];
+      this.childViews && this.childViews.forEach(view => {
```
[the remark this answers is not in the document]
oliverfoster: So it is.

src/core/js/views/adaptView.js
oliverfoster: Nope

src/core/js/views/adaptView.js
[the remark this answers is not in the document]
oliverfoster: Reverted

PR #2741, opened 2020-05-04
Added menuGroup aria level support

guywillis:
Related to feature request in this issue: adaptlearning/adapt_framework#2702

* Adds menuGroup aria label level


moloko:
is it worth adding a note to the help text for `_menuItem` stating that it should be set to `3` if 'menu groups' are being used (and a11y support is required)?

Issue #2749, opened 2020-05-07
ItemsComponentModel: setActiveItem should disallow invalid values


moloko:
Currently you can set the active item to `-1` or a value greater than the number of items; this should not be allowed.

PR #2751, opened 2020-05-07
prevent invalid value from being processed by setActiveItem

moloko:
see #2749

### DIFF
--- a/src/core/js/models/itemsComponentModel.js
+++ b/src/core/js/models/itemsComponentModel.js
@@ -68,7 +68,9 @@ define([
       this.get('_children').each(item => item.toggleActive(false));
     }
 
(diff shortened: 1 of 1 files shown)


src/core/js/models/itemsComponentModel.js
```diff
@@ -68,10 +68,13 @@ define([
       this.get('_children').each(item => item.toggleActive(false));
     }
 
-    setActiveItem(index) {
+    setActiveItem() {
```
tomgreenfield, line 71: Please restore the parameter.

moloko: oops! thanks @tomgreenfield

PR #2752, opened 2020-05-07
prevent invalid value from being processed by setActiveItem (v4 version)

moloko:
see #2749

### DIFF
--- a/src/core/js/models/itemsComponentModel.js
+++ b/src/core/js/models/itemsComponentModel.js
@@ -66,6 +66,8 @@ define([
         },
 
         setActiveItem: function(index) {
+            if (index === null || index === undefined || index < 0 || index >= this.get('_children').length) return;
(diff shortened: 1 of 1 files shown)


src/core/js/models/itemsComponentModel.js
```diff
@@ -66,6 +66,8 @@ define([
         },
 
         setActiveItem: function(index) {
+            if (index === null || index === undefined || index < 0 || index >= this.get('_children').length) return;
```
tomgreenfield, line 69: Could just attempt to get the item (needed to toggle it anyway):
```js
var item = this.getItem(index);
if (!item) return;
...
item.toggleActive(true);
```

moloko: derp yes that would be easier wouldn't it

Issue #2758, opened 2020-05-09
adapt_framework v5.7.0: Add model deepClone 


olivermartinfoster:
### Subject of the issue
Deep cloning is useful for creating branching scenarios as looping models need to be cloned for render, but it also ensures that the data layer is able to receive new models after initialisation and utilise them as though they were any other model.

PR #2759, opened 2020-05-09
issue/2758 Added deepClone API (v5.7.0)

olivermartinfoster:
#2758 

#### Added
* `AdaptModel.deepClone` to clone model hierarchies

#### Changed
* `Adapt.findById` Moved indexing to add/remove events in data collection rather than indexing after file download
* `AdaptModel.setupModel` Stopped  running on the adapt:initialize event but instead is called after cloned or on instantiation after file download
* Allowed `QuestionModel` to `restoreUserAnswers` after clone without having to listen to the `adapt:initialize` event
* Changed the order of `init` execution in `ItemsQuestionModel` such that `ItemsComponentModel` can setup its children before `QuestionModel` tries to `restoreUserAnswers`


src/core/js/models/questionModel.js
```diff
@@ -45,7 +45,11 @@ define([
 
     init() {
       this.setupDefaultSettings();
-      this.listenToOnce(Adapt, 'adapt:initialize', this.onAdaptInitialize);
+      if (Adapt.get('_isStarted')) {
+        this.onAdaptInitialize();
```
moloko, line 49: `return` from the `if` and remove the `else`?

Issue #2760, opened 2020-05-09
adapt_framework v5.5.0: Add ChildViews API for rendering control


olivermartinfoster:
### Subject of the issue
Add an API which allows any plugin to control the render of child views. 

Adding child views is part of the Adapt Framework rendering process, as each container view starts to render its own content it also instantiates child views from its model's children and appends the views to the document object model.

Currently a page view renders itself, then calls addChildren to render all its immediate children, those children then proceed on the same path. The views are added parent-first, each column of the hierarchy, articles, blocks and components are rendered one at a time until the whole page is constructed.

For extensions such as trickle, which need to stop the rendering process at a given child, this method of rendering doesn’t work.

A better way of implementing this is to extend the addChildren function to render first-child-first, to have a control mechanism which informs plugins that is about to add a child, to say that it has added a child and to request any additional children and allow plugins to control that process. It should label views as having been rendered or not and provide mechanism to wait for or force children to be render.

PR #2761, opened 2020-05-09
issue/2760 Added ChildViews API for rendering control (v5.5.0)

olivermartinfoster:
#2760 

... documentation in progress...

### DIFF
--- a/src/core/js/adapt.js
+++ b/src/core/js/adapt.js
@@ -343,7 +343,9 @@ define([
       }
 
       const foundView = idPathToView.reduce((view, currentId) => {
-        return view && view.childViews && view.childViews[currentId];
+        if (!view) return;
+        const childViews = view.getChildViews();
+        return childViews && childViews.find(view => view.model.get('_id') === currentId);
       }, this.parentView);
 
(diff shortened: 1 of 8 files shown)


src/core/js/adapt.js
```diff
@@ -398,8 +402,14 @@ define([
...
-        currentView.model.setOnChildren('_isReady', false);
-        currentView.model.set('_isReady', false);
+        currentView.model.setOnChildren({
+          '_isReady': false,
+          '_isRendered': false
+        });
+        currentView.model.set({
+          '_isReady': false,
+          '_isRendered': false
+        });
```
moloko, line 412: I don't think this is needed as `setOnChildren` applies the settings to the model **and** its children

moloko:
this PR should probably be changed to merge into master rather than release/v5.6.0 I think?

src/core/js/adapt.js
oliverfoster: that's true

moloko:
@olivermartinfoster conflict needs resolving...

Issue #2782, opened 2020-05-26
adapt_framework: jquery.resize loop not suspending properly


oliverfoster:
### Subject of the issue
The loop which checks for dom resize events should stall after it expires. The `requestAnimationFrame` callback timestamp and loose boolean comparison is causing it not to stop.

https://github.com/adaptlearning/adapt_framework/blob/master/src/core/libraries/jquery.resize.js#L171-L178

`force` will be the `requestAnimationFrame` timestamp when called from raf

PR #2783, opened 2020-05-26
issue/2782 Fixed bug in jquery.resize

oliverfoster:
#2782 
* Modified `force` check to account for type


lc-alexanderbenesch:
👍

Issue #2796, opened 2020-06-15
Question component - Marking should only be displayed on final question attempt


cahirodoherty-learningpool:
### Subject of the enhancement
The Question component could benefit from a 'Display Final Marking Only' type setting

### Your environment
* FW5

### Steps to reproduce
Create a course with Matching Question components with multiple question stems; each question having only 2 possible answers and the user has 2 attempts.

### Expected behaviour
The component could be set to only display marking on the final attempt

### Actual behaviour
After the first attempt, the marking is displayed, making it quite obvious to the user what the correct answers are for each Matching Question

moloko:
I don't seem to be able to replicate this, am I missing something obvious?
![image](https://user-images.githubusercontent.com/1249597/84805465-3855fc00-affc-11ea-9a2e-db4a8d4592c4.png)

Using FW v5.3.0 and Matching v4.0.0 it displays marking only on the final attempt, as expected... is it specific to a more recent FW version perhaps?

oliverfoster:
Weird. From the code it definitely adds the classes each time, which it definitely shouldn't.

QuestionView:onSubmitClicked
https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/questionView.js#L189

MatchingView:showMarking
https://github.com/adaptlearning/adapt-contrib-matching/blob/master/js/matchingView.js#L86

moloko:
Ah OK yes IIRC it was adding the classes - but not displaying ticks/crosses. There must be something in the theme preventing ticks/crosses from being displayed until all attempts used up. We should definitely stop the classes from being added in the first place though.

oliverfoster:
I assume something about the styling / version is a reason for it. I do remember a few clients telling us that their users were using the debugger to figure out from the marking classes what the correct / incorrect options are. This will at least help tackle that one at the same time.

cahirodoherty-learningpool:
Yeh I have just confirmed that with FW5 it is behaving as you say @moloko 
Based on the conversation it probably does make sense for the code changes to go in to address any inspection cheating by end users

oliverfoster:
There was one more on found in this vein. 
![image](https://user-images.githubusercontent.com/7974663/85271700-0d8ffb80-b473-11ea-9a7a-3f4cdbd36a1f.png)

moloko:
This change has created couple of issues for MCQs with more than one attempt:

[...]

PR #2797, opened 2020-06-15
issue/2796 Limit display of marking until final question attempt

cahirodoherty-learningpool:
#2796 
* Limiting display of marking until final question attempt has been made

### DIFF
--- a/src/core/js/views/buttonsView.js
+++ b/src/core/js/views/buttonsView.js
@@ -148,7 +148,7 @@ define([
     },
 
     showMarking: function() {
-      if (!this.canShowMarking) return;
+      if (!this.model.canShowMarking) return;
 
       var isCorrect = this.model.get('_isCorrect');
       var ariaLabels = Adapt.course.get('_globals')._accessibility._ariaLabels;
--- a/src/core/js/views/buttonsView.js
+++ b/src/core/js/views/buttonsView.js
@@ -150,6 +150,10 @@ define([
     showMarking: function() {
       if (!this.model.get('_canShowMarking')) return;
 
+      var attemptsLeft = (this.model.get('_attemptsLeft')) ? this.model.get('_attemptsLeft') : this.model.get('_attempts');
+
+      if (this.model.get('_onlyShowFinalMarking') && attemptsLeft !== 0) return;
+
       var isCorrect = this.model.get('_isCorrect');
       var ariaLabels = Adapt.course.get('_globals')._accessibility._ariaLabels;
 
--- a/src/core/js/views/buttonsView.js
+++ b/src/core/js/views/buttonsView.js
@@ -148,11 +148,7 @@ define([
     },
 
     showMarking: function() {
-      if (!this.model.get('_canShowMarking')) return;
-
-      var attemptsLeft = (this.model.get('_attemptsLeft')) ? this.model.get('_attemptsLeft') : this.model.get('_attempts');
-
-      if (this.model.get('_onlyShowFinalMarking') && attemptsLeft !== 0) return;
+      if (!this.canShowMarking) return;
 
       var isCorrect = this.model.get('_isCorrect');
       var ariaLabels = Adapt.course.get('_globals')._accessibility._ariaLabels;
--- a/src/core/js/views/questionView.js
+++ b/src/core/js/views/questionView.js
@@ -23,7 +23,7 @@ define([
         (this.model.get('_isOptional') ? 'is-optional' : ''),
         (this.model.get('_canShowModelAnswer') ? 'can-show-model-answer' : ''),
         (this.model.get('_canShowFeedback') ? 'can-show-feedback' : ''),
-        (this.canShowMarking ? 'can-show-marking' : '')
+        (this.model.canShowMarking ? 'can-show-marking' : '')
       ].join(' ');
     }
 
@@ -39,19 +39,6 @@ define([
       this._runModelCompatibleFunction('setupDefaultSettings');
       // Blank method for setting up questions before rendering
       this.setupQuestion();
-    }
-
-    /** @type {boolean} */
-    get canShowMarking() {
-      if (!this.model.get('_onlyShowFinalMarking')) {
-        return this.model.get('_canShowMarking');
-      }
-
-      const attemptsLeft = this.model.get('_attemptsLeft');
-      const hasMultipleAttempts = this.model.get('_attempts') > 1;
-      const isFinalAttempt = !hasMultipleAttempts || (attemptsLeft === 0);
-
-      return isFinalAttempt;
     }
 
     // Used in the question view to disabled the question when _isEnabled has been set to false
--- a/src/core/js/views/questionView.js
+++ b/src/core/js/views/questionView.js
@@ -185,7 +185,9 @@ define([
       this._runModelCompatibleFunction('setScore');
 
       // Used by the question to display markings on the component
-      this.showMarking();
+      if (this.model.canShowMarking) {
+        this.showMarking();
+      }
 
       // Used to check if the question is complete
       // Triggers setCompletionStatus and adds class to widget
--- a/src/core/js/views/questionView.js
+++ b/src/core/js/views/questionView.js
@@ -23,7 +23,7 @@ define([
         (this.model.get('_isOptional') ? 'is-optional' : ''),
         (this.model.get('_canShowModelAnswer') ? 'can-show-model-answer' : ''),
         (this.model.get('_canShowFeedback') ? 'can-show-feedback' : ''),
-        (this.getCanShowMarking() ? 'can-show-marking' : '')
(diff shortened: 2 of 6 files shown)


src/core/js/views/questionView.js
```diff
@@ -39,7 +39,15 @@ define([
       this._runModelCompatibleFunction('setupDefaultSettings');
       // Blank method for setting up questions before rendering
       this.setupQuestion();
+    }
+
+    getCanShowMarking() {
```
oliverfoster, line 44: You can probably use an es6 property here unless you don't fancy it?

src/core/js/views/questionView.js
```diff
@@ -39,7 +39,15 @@ define([
...
+      if (this.model.get('_onlyShowFinalMarking')) {
+        var attemptsLeft = (this.model.get('_attemptsLeft')) ? this.model.get('_attemptsLeft') : this.model.get('_attempts');
```
oliverfoster, line 46: Logic is sound, confused me for a bit though, perhaps something like:
```js
const attemptsLeft = this.model.get('_attemptsLeft');
const hasMultipleAttempts = Boolean(this.model.get('_attempts'));
const isFinalAttempt = !hasMultipleAttempts ||(attemptsLeft === 0);
```

What I've written is obviously all about face, but making the implied more explicit would be cool.

src/core/js/views/questionView.js
```diff
@@ -307,7 +319,7 @@ define([
     refresh() {
       this.model.set('_buttonState', this.model.getButtonState());
 
-      if (this.model.get('_canShowMarking') && this.model.get('_isInteractionComplete') && this.model.get('_isSubmitted')) {
+      if (this.canShowMarking() && this.model.get('_isInteractionComplete') && this.model.get('_isSubmitted')) {
```
oliverfoster, line 322: Brackets can be removed as not a function any more
```suggestion
      if (this.canShowMarking && this.model.get('_isInteractionComplete') && this.model.get('_isSubmitted')) {
```

src/core/js/views/questionView.js
```diff
@@ -39,7 +39,19 @@ define([
...
+    }
+
+    /** @type {boolean} */
+    get canShowMarking() {
+      if (!this.model.get('_onlyShowFinalMarking')) {
+        return this.model.get('_canShowMarking');
+      }
+
+      const attemptsLeft = this.model.get('_attemptsLeft');
+      const hasMultipleAttempts = Boolean(this.model.get('_attempts') > 1);
```
oliverfoster, line 51: With:
```js
Boolean(this.model.get('_attempts') > 1);
```

Use either:
```js
(this.model.get('_attempts') > 1) // comparison operator gets evaluated to a boolean
```

or

```js
Boolean(this.model.get('_attempts')) // boolean coercion, false when value is 0, null, undefined, "", etc, true otherwise
```

They do the same thing. I think I like the comparison operator better.
```js
(this.model.get('_attempts') > 1)
```

src/core/js/views/buttonsView.js
```diff
@@ -150,6 +150,10 @@ define([
     showMarking: function() {
       if (!this.model.get('_canShowMarking')) return;
 
+      var attemptsLeft = (this.model.get('_attemptsLeft')) ? this.model.get('_attemptsLeft') : this.model.get('_attempts');
+
+      if (this.model.get('_onlyShowFinalMarking') && attemptsLeft !== 0) return;
```
oliverfoster, line 155: All of the above an be removed for:

```js
if (!this.canShowMarking) return;
```

src/core/js/views/questionView.js
```diff
@@ -39,7 +39,19 @@ define([
       this._runModelCompatibleFunction('setupDefaultSettings');
       // Blank method for setting up questions before rendering
       this.setupQuestion();
+    }
+
+    /** @type {boolean} */
+    get canShowMarking() {
```
oliverfoster, line 45: Move to `QuestionModel` as not accessible in `ButtonsView` my mistake. I'm clearly not thinking properly.

src/core/js/views/buttonsView.js
```diff
@@ -148,7 +148,7 @@ define([
     },
 
     showMarking: function() {
-      if (!this.model.get('_canShowMarking')) return;
+      if (!this.canShowMarking) return;
```
oliverfoster, line 151: `this.model.canShowMarking`

src/core/js/views/questionView.js
```diff
@@ -23,7 +23,7 @@ define([
         (this.model.get('_isOptional') ? 'is-optional' : ''),
         (this.model.get('_canShowModelAnswer') ? 'can-show-model-answer' : ''),
         (this.model.get('_canShowFeedback') ? 'can-show-feedback' : ''),
-        (this.model.get('_canShowMarking') ? 'can-show-marking' : '')
+        (this.canShowMarking ? 'can-show-marking' : '')
```
oliverfoster, line 26: `this.model.canShowMarking`

src/core/js/views/questionView.js
```diff
@@ -307,7 +319,7 @@ define([
     refresh() {
       this.model.set('_buttonState', this.model.getButtonState());
 
-      if (this.model.get('_canShowMarking') && this.model.get('_isInteractionComplete') && this.model.get('_isSubmitted')) {
+      if (this.canShowMarking && this.model.get('_isInteractionComplete') && this.model.get('_isSubmitted')) {
```
oliverfoster, line 322: `this.model.canShowMarking`

cahirodoherty-learningpool:
> I'm happy if you're happy?
> 
> Some of the components might need updating to do the inverse, where they show marking on all attempts when this property isn't set.
> 
> Should `_onlyShowFinalMarking` default to `true`? I think this might be the expected default behaviour.

If we keep the default to false then behaviour will remain consistent for existing question components. This addresses both your points raised

oliverfoster:
Lovely.

danielstorey:
I take it all question component schemas, examples and READMEs will need updating with the new property?

cahirodoherty-learningpool:
> I take it all question component schemas, examples and READMEs will need updating with the new property?

Yes, I will take responsibility for that and link them to the issue

moloko:
> Should `_onlyShowFinalMarking` default to true? I think this might be the expected default behaviour.

I'd even make a case for this to be the default behaviour and instead have a setting called something like `_showMarkingForAllAttempts` for anyone wanting the existing behaviour. However, appreciate I'm very late to the party on this one so if you want to go ahead and continue down this route that's also fine!

cahirodoherty-learningpool:
> > Should `_onlyShowFinalMarking` default to true? I think this might be the expected default behaviour.
> 
> I'd even make a case for this to be the default behaviour and instead have a setting called something like `_showMarkingForAllAttempts` for anyone wanting the existing behaviour. However, appreciate I'm very late to the party on this one so if you want to go ahead and continue down this route that's also fine!

Upon reflection I think the case for the marking to only display on the last attempt is probably fine as being defaulted to true. I will amend the code accordingly

oliverfoster:
Do you mean to remove the property `_onlyShowFinalMarking` but keep the latter half of `canShowMarking` ? @moloko 

That does sound much more sensible.

```js
    get canShowMarking() {
      if (!this.get('_canShowMarking')) {
        return false
      }

      const attemptsLeft = this.get('_attemptsLeft');
      const hasMultipleAttempts = this.get('_attempts') > 1;
      const isFinalAttempt = !hasMultipleAttempts || (attemptsLeft === 0);

      return isFinalAttempt;
    }

**Comment by cahirodoherty-learningpool** [comment]
> Do you mean to remove the property `_onlyShowFinalMarking` but keep the latter half of `canShowMarking` ? @moloko
> 
> That does sound much more sensible. I thought this was how the framework behaved by default, but it clearly doesn't. You wouldn't have to update all the schemas then @cahirodoherty-learningpool
> 
> ```js
>     get canShowMarking() {
>       if (!this.get('_canShowMarking')) {
>         return false;
>       }
> 
>       const attemptsLeft = this.get('_attemptsLeft');
>       const hasMultipleAttempts = this.get('_attempts') > 1;
>       const isFinalAttempt = !hasMultipleAttempts || (attemptsLeft === 0);
> 
>       return isFinalAttempt;
>     }
> ```

This alone would not provide the ability for display of marking after every attempt. If any user has currently set up a course with this intention, it would no longer behave as expected.

**Comment by oliverfoster** [comment]
I think only the matching behaved like that anyway? It could be considered a bugfix.

**Comment by oliverfoster** [comment]
To stop the classes appearing you'll need to wrap [this line](https://github.com/adaptlearning/adapt_framework/blob/f34208953151c95e88f0a8df0985c04e1b7aa1aa/src/core/js/views/questionView.js#L188) with:

```js
if (this.model.canShowMarking) {
  this.showMarking();
}
```

That is as the `showMarking` function is implemented by the component view and can't be relied upon to check `canShowMarking`.

Issue #2812, opened 2020-06-22
Error message when no input provided for question


simondate:
### Subject of the issue/enhancement/features

Questions (MCQ, GMCQ) don't provide any input when submit is pressed with no answer submitted.
The Matching component does prompt users but only does so by highlighting the dropdown boxes that haven't been selected. Text Input and Slider just takes the default value (empty or 1).

### Your environment
* version (AT/Framework) v5.6.0

Note: [Proposed solution](https://github.com/adaptlearning/adapt_framework/issues/2812#issuecomment-648017142) below, comments welcome.

moloko:
IIRC it is supposed to highlight the instruction text in red...

tomgreenfield:
Reference before Vanilla refactor: https://github.com/adaptlearning/adapt-contrib-vanilla/blob/157b98cc45d0febadfa6f5f3a4e07bfded586676/less/core/component.less#L45-L49.

oliverfoster:
We also have the option to do something different. Like a notify with a warning in it or similar.

simondate:
I agree it should be more than just colour as this wouldn't be accessible - https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-without-color.html

moloko:
@oliverfoster switch to having the submit button disabled until the learner has met the `canSubmit` criteria?

oliverfoster:
Disabling the button sounds like a plan. The instruction text would have to make it pretty clear each time that an option needs to be selected. Slider will be an issue with anything global as the first option is always a valid selection.

oliverfoster:
What do we think ^ ?

moloko:
looks good to me. increasingly this seems to be the desired behaviour for many of the clients I look after so it would be really helpful for this to be the default...

oliverfoster:
Default with optional disable or default unchangeable?

moloko:
it's a good question, I'll ask it on the forums and chat and see what people think...

oliverfoster:
I'll get the prs for matching and textinput done in the meantime.

chucklorenz:
👍 for default unchangeable

oliverfoster:
Have added this as an alternative:

This is with course.json: `_buttons._incompleteWarning._isEnabled = true`
![incompleteWarning](https://user-images.githubusercontent.com/7974663/85453715-10364200-b594-11ea-9db2-8c8be84e0ea1.gif)

[...]

PR #2814, opened 2020-06-23
issue/2812 Improved question canSubmit: false behaviour

oliverfoster:
#2812 

This pr adds an optional incomplete question notification when no options are selected and the submit button is click that is disabled by default. It also modifies the submit button so that when the incomplete question notification is disabled, the submit button is instead disabled until the instructions have been satisfied.

### Added
* `_canSubmit` property to question model, default `true`
* `checkCanSubmit()` function to question model to set `_canSubmit` based upon value of `canSubmit()` from the question component
* `ButtonsView` listens to `_canSubmit` changes and updates the button state accordingly
* `ItemsQuestionModel` runs `checkCanSubmit` on the model when the active items change
* Optional "Incomplete Warning" notify configured from the `course.json`
```json
{
  "_buttons": {
    "_incompleteWarning": {
      "_isEnabled": false,
      "title": "Instructions",
      "body": "{{#if instruction}}{{instruction}}{{else}}Please complete the question and then click Submit.{{/if}}"
    }
  }
}
```

Notes:
* Might need a switch to revert the disabled submit button to original behaviour.
* Will only work on `ItemsQuestionModel` components (mcq and gmcq) out of the box, so matching and textinput - which have their own models - will need to run `checkCanSubmit()` on `init` and after selection. PRs referenced below.

### DIFF
--- a/src/core/js/models/itemsQuestionModel.js
+++ b/src/core/js/models/itemsQuestionModel.js
@@ -30,9 +30,7 @@ define([
     init() {
       super.init();
       this.set('_isRadio', this.isSingleSelect());
-      this.listenTo(this.getChildren(), {
-       'change:_isActive': this.checkCanSubmit
-      });
+      this.listenTo(this.getChildren(), 'change:_isActive': this.checkCanSubmit);
       this.checkCanSubmit();
     }
 
--- a/src/core/js/models/itemsQuestionModel.js
+++ b/src/core/js/models/itemsQuestionModel.js
@@ -30,7 +30,7 @@ define([
     init() {
       super.init();
       this.set('_isRadio', this.isSingleSelect());
-      this.listenTo(this.getChildren(), 'change:_isActive': this.checkCanSubmit);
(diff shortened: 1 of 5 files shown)


src/core/js/models/itemsQuestionModel.js
```diff
@@ -30,6 +30,8 @@ define([
     init() {
       super.init();
       this.set('_isRadio', this.isSingleSelect());
+      this.listenTo(this.getChildren(), 'change:_isActive': this.checkCanSubmit);
```
simondate, line 33: This line throws an error when trying to you try to build the course.

```
    Error: Line 33: Unexpected token :
    In module tree:
        core/js/app
          components/adapt-contrib-gmcq/js/adapt-contrib-gmcq
            components/adapt-contrib-mcq/js/adapt-contrib-mcq
```

oliverfoster: Apologies, try now.

oliverfoster:
prefer https://github.com/adaptlearning/adapt_framework/pull/2820

PR #2815, opened 2020-06-24
issue/2796 Minor additional fixes

### LINKED ISSUE #2796: Question component - Marking should only be displayed on final question attempt

oliverfoster:
#2796 
### Fixed
* buttons template so that correctness aria label doesn't appear immediately
* Question view `can-show-marking` class should appear when property is set
* `QuesitonModel.canShowMarking` return true if property set and component interaction is complete

### DIFF
--- a/src/core/js/models/questionModel.js
+++ b/src/core/js/models/questionModel.js
@@ -321,7 +321,7 @@ define([
     getResponseType() {}
 
     /** @type {boolean} */
-    get canShowMarking() {
+    get shouldShowMarking() {
       if (!this.get('_canShowMarking')) {
         return false;
       }
--- a/src/core/js/models/questionModel.js
+++ b/src/core/js/models/questionModel.js
@@ -326,7 +326,7 @@ define([
         return false;
       }
 
-      if (this.get('_isComplete')) return true;
+      if (this.get('_isInteractionComplete')) return true;
 
       const attemptsLeft = this.get('_attemptsLeft');
       const hasMultipleAttempts = this.get('_attempts') > 1;
--- a/src/core/js/models/questionModel.js
+++ b/src/core/js/models/questionModel.js
@@ -326,13 +326,7 @@ define([
         return false;
       }
 
-      if (this.get('_isInteractionComplete')) return true;
-
-      const attemptsLeft = this.get('_attemptsLeft');
-      const hasMultipleAttempts = this.get('_attempts') > 1;
-      const isFinalAttempt = !hasMultipleAttempts || (attemptsLeft === 0);
-
-      return isFinalAttempt;
+      return this.get('_isInteractionComplete');
     }
   }
 
--- a/src/core/js/models/questionModel.js
+++ b/src/core/js/models/questionModel.js
@@ -326,6 +326,8 @@ define([
         return false;
       }
 
+      if (this.get('_isComplete')) return true;
+
       const attemptsLeft = this.get('_attemptsLeft');
       const hasMultipleAttempts = this.get('_attempts') > 1;
       const isFinalAttempt = !hasMultipleAttempts || (attemptsLeft === 0);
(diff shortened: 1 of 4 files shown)


oliverfoster:
@moloko I think that might be about it?

moloko:
worth adding a comment to the `canShowMarking` getter to clarify that its purpose is determine when marking _should_ be shown rather than just returning the value of `_canShowMarking`...?

src/core/js/models/questionModel.js
```diff
@@ -321,7 +321,7 @@ define([
     getResponseType() {}
 
     /** @type {boolean} */
-    get canShowMarking() {
+    get shouldShowMarking() {
```
moloko, line 324: much better!

oliverfoster: I don't see why not, seeing as it's brand new.

src/core/js/models/questionModel.js
```diff
@@ -321,16 +321,12 @@ define([
...
       if (!this.get('_canShowMarking')) {
         return false;
       }
 
-      const attemptsLeft = this.get('_attemptsLeft');
-      const hasMultipleAttempts = this.get('_attempts') > 1;
-      const isFinalAttempt = !hasMultipleAttempts || (attemptsLeft === 0);
-
-      return isFinalAttempt;
+      return this.get('_isInteractionComplete');
```
oliverfoster, line 329: This works because all the attempts checking stuff has to happen before _isInteractionComplete is set so it become implied here.

src/core/js/views/questionView.js
```diff
@@ -308,7 +308,7 @@ define([
     refresh() {
       this.model.set('_buttonState', this.model.getButtonState());
 
-      if (this.model.canShowMarking && this.model.get('_isInteractionComplete') && this.model.get('_isSubmitted')) {
+      if (this.model.shouldShowMarking && this.model.get('_isInteractionComplete') && this.model.get('_isSubmitted')) {
```
moloko, line 311: seeing as `shouldShowMarking` now checks `_isInteractionComplete`...
```suggestion
      if (this.model.shouldShowMarking && this.model.get('_isSubmitted')) {
```

src/core/templates/buttons.hbs
```diff
@@ -1,6 +1,6 @@
 <div class="btn__response-container">
 
-  <div class="btn__marking js-btn-marking u-display-none" aria-label="{{#if _isCorrect}}{{@root/_globals._accessibility._ariaLabels.answeredCorrectly}}{{else}}{{@root/_globals._accessibility._ariaLabels.answeredIncorrectly}}{{/if}}">
+  <div class="btn__marking js-btn-marking u-display-none" aria-label="{{#all _isInteractionComplete _canShowMarking}}{{#if _isCorrect}}{{@root/_globals._accessibility._ariaLabels.answeredCorrectly}}{{else}}{{@root/_globals._accessibility._ariaLabels.answeredIncorrectly}}{{/if}}{{/all}}">
```
moloko, line 3: could the getter be used here instead?
```suggestion
  <div class="btn__marking js-btn-marking u-display-none" aria-label="{{#if canShowMarking}}{{#if _isCorrect}}{{@root/_globals._accessibility._ariaLabels.answeredCorrectly}}{{else}}{{@root/_globals._accessibility._ariaLabels.answeredIncorrectly}}{{/if}}{{/all}}">
```

oliverfoster: No. The data in the handlebars content is not the model it's just a json object from model.attributes / model.toJSON() so there is no reference to the canShowMarking property.

PR #2820, opened 2020-06-29
issue/2812-1 Disabled submit button by default, facilitate instruction error

oliverfoster:
#2812 

#### Added
* `questionView:showInstructionError` event to allow for instruction error popups
* `_canSubmit` with locking to allow extension to lock `_canSubmit` to true

#### Removed
* Old instruction error behaviour

Use extension [adapt-contrib-instructionError](https://github.com/adaptlearning/adapt-contrib-instructionError) as discussed for warnings on invalid submissions.

Requires prs in matching and textinput to disable submit button in both cases [matching#125](https://github.com/adaptlearning/adapt-contrib-matching/pull/125), [textinput#101](https://github.com/adaptlearning/adapt-contrib-textInput/pull/101)

### DIFF
--- a/src/core/js/models/itemsQuestionModel.js
+++ b/src/core/js/models/itemsQuestionModel.js
@@ -30,6 +30,8 @@ define([
     init() {
       super.init();
       this.set('_isRadio', this.isSingleSelect());
+      this.listenTo(this.getChildren(), 'change:_isActive', this.checkCanSubmit);
+      this.checkCanSubmit();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


oliverfoster:
I think you might be overthinking it? It'll become apparent very quickly that it's not clickable or that it changes when you choose an option. With every UI the user must learn the implied norms as they use it. There are obviously stylist things, like recessed edging, crossing them out, etc, but we already have the instruction text and an alternative implementation.

Issue #2825, opened 2020-07-02
Framework - questionView checkIfResetOnRevisit function expects boolean


guywillis:
### Subject of the issue
On [L72](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/questionView.js#L72) of questionView the checkIfResetOnRevisit function expects a boolean but due to this [commit](https://github.com/adaptlearning/adapt_framework/commit/b6ec1d15c6d14e7f1623f9fb5dc2632fdc023a9a) and this [PR](https://github.com/adaptlearning/adapt_framework/pull/2655) a string is returned instead. This issue causes components in Authoring Tools to enter an error state when revisiting a previously answered question.

### Your environment
* FW v5.5.1 +

### Expected behaviour
Components should restore correctly on revisit of page.

### Screenshots

Submitted, first visit.
![Screenshot 2020-07-02 at 11 58 17](https://user-images.githubusercontent.com/11569678/86350993-7178a800-bc5b-11ea-97e6-86bbf0b82171.png)

Still submitted, second visit. No item marking, incorrect submission CSS. 
![Screenshot 2020-07-02 at 11 58 29](https://user-images.githubusercontent.com/11569678/86350996-72113e80-bc5b-11ea-9b71-d533b741bf1c.png)

moloko:
reopening because I don't think #2826 completely solved this - the value needs to be converted in adaptModel.js rather than questionView.js - see [this forum post](https://community.adaptlearning.org/mod/forum/discuss.php?d=3958). I can replicate this bug in the latest master branch.

PR #2826, opened 2020-07-02
issues/2825 _isResetOnRevisit "false" string convertion

oliverfoster:
#2825 
* Convert `"false"` to `false`


muthuselvan86 (commented):
Ideally this issue is coming down from `adapt_framework/src/core/schema/component.model.schema` where the default value for `_isResetOnRevisit` is set as `"false"` with double quotes.

Issue #2824, opened 2020-07-02
adapt_framework: sourcemaps break with too many plugins


oliverfoster:
### Subject of the issue
It seems that bundling with requirejs and handing the code and maps to babel to transpile has some file-size limitations. The source maps seem to be silently rejected by babel when the bundled code reaches >~250KB.

On the up-side, there are outstanding compilation issues that need solving:
* import/export directives (remove requirejs)
* beyond ES8 (remove requirejs)

So I'm thinking about replacing requirejs as the bundler.

PR #2827, opened 2020-07-03
issue/2824 Replaced requirejs with rollup for module bundler

oliverfoster:
#2824 
Will provide an almost seamless migration away from build-side [requirejs](https://requirejs.org/docs/optimization.html) and [AMD](https://requirejs.org/docs/whyamd.html) modules to [ES6](https://hacks.mozilla.org/2015/08/es6-in-depth-modules/) modules

#### Facilitates
* future proof
  * beyond ES8 code
* development improvements
  * supports larger code bases
* backward-compatibility
  * ES6 modules are are reregistered in the browser as AMD modules, so `var Adapt = require('core/js/adapt');` etc will still work
  * still ie11 compatible

#### Downsides

#### Added
* [babel-plugin-transform-amd-to-es6](https://github.com/cgkineo/babel-plugin-transform-amd-to-es6) to convert [AMD](https://requirejs.org/docs/whyamd.html) modules to [ES6](https://hacks.mozilla.org/2015/08/es6-in-depth-modules/) modules

#### Removed

#### Fixed

#### Testing
* Change [scorm/logger.js](https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/scorm/logger.js#L1-L4) header to:
```js
var Logger = window.Logger = function() {
  this.logArr = [];
  this.registeredViews = [];
};
```

#### Example

rollup-es-modules2

### Other
See a variant with added typescript support https://github.com/adaptlearning/adapt_framework/tree/issue/2824-ts with added command `npm run check-types`, it's a fair bit slower than the babel-only variety because it has to get processed by two compilers.

See a variant with added react support https://github.com/adaptlearning/adapt_framework/pull/2829 - I am very keen to get this in, it would mean we could ditch jQuery for DOM modification and we'd get a lot more flexibility as views will automatically rerender themselves on model changes.


oliverfoster:
This is ready for a proper review.

moloko:
> See a variant with added typescript support https://github.com/adaptlearning/adapt_framework/tree/issue/2824-ts with added command npm run check-types, it's a fair bit slower than the babel-only variety because it has to get processed by two compilers.

Just to note that the TypeScript compiler should be able to transpile down to ES5 by itself, so there shouldn't be any need to run it through babel as well. Honestly though, don't worry too much about trying to include support for TypeScript - I think getting support for React in Adapt would be a far bigger win.

oliverfoster:
I also had to make two custom AST transformation which are really easy to do in babel, for amd to ES6 and for jsx templates. Hopefully by the time we've converted everything over to ES6, typescript will have matured to a level where making transformation plugins is as easy as it is in babel. If we limit the stack through rollup, babel and react at the moment it leaves Typescript as a future possibility as the sole compiler.

I couldn't get it to transpile from ES6 > ES5, the `--target` option set to `ES5` just told me that `class` isn't a thing when compile. I am probably missing some properties. Perhaps `--lib`? https://www.typescriptlang.org/docs/handbook/compiler-options.html

oliverfoster:
Yea, I was missing `--libs` https://github.com/Microsoft/TypeScript/issues/25853

oliverfoster:
This was a reasonably good article on the subject: https://iamturns.com/typescript-babel/

oliverfoster:
closed in favour of https://github.com/adaptlearning/adapt_framework/pull/2829

PR #2829, opened 2020-07-05
issue/2824-jsx Changed bundler and added React support

oliverfoster:
#2824 
Will provide an almost seamless migration away from build-side [requirejs](https://requirejs.org/docs/optimization.html) and [AMD](https://requirejs.org/docs/whyamd.html) modules to [ES6](https://hacks.mozilla.org/2015/08/es6-in-depth-modules/) modules and add react support to the framework.

#### Facilitates
* future proof
  * beyond ES8 code
  * react templates
* development improvements
  * supports larger code bases
* backward-compatibility
  * ES6 modules are reregistered in the browser as AMD modules, so `var Adapt = require('core/js/adapt');` etc will still work in the console and for other plugins
  * still ie11 compatible

#### Downsides
* Currently this will only work with transpilation back to ES5 as calling an ES6 `constructor` function without the `new` keyword is not allowed - which is an essential part of the way Backbone constructs classes.

#### Added
* [babel-plugin-transform-amd-to-es6](https://github.com/cgkineo/babel-plugin-transform-amd-to-es6) (custom made) to convert [AMD](https://requirejs.org/docs/whyamd.html) modules to [ES6](https://hacks.mozilla.org/2015/08/es6-in-depth-modules/) modules

These plugins achieve react templating and integration in the same way we have been doing with handlebars:
* [babel-plugin-transform-react-templates](https://github.com/cgkineo/babel-plugin-transform-react-templates/) (custom made) register templates from filenames
* [html-react-parser](https://www.npmjs.com/package/html-react-parser) parse html into react nodes, like triples braces in handlebars
* [plugin-transform-react-jsx](https://www.npmjs.com/package/@babel/plugin-transform-react-jsx) jsx compiler
* [react](https://www.npmjs.com/package/react) client-side react node handler
* [react-dom](https://www.npmjs.com/package/react-dom) client-side react node renderer

#### Removed

#### Fixed

#### Testing
* Use master of all plugins and accordion branch issue/2824-jsx by running:
```js
git clone https://github.com/adaptlearning/adapt_framework 2824
cd 2824
git checkout issue/2824-jsx
cd src/components/adapt-contrib-accordion
git checkout issue/2824-jsx
cd ../../../
grunt dev
```

#### Example

rollup-es-modules2

#### Templates
Supports react 'templates' as near as I could get it to our current handlebars templates - separate, override-able etc. This change does not require a wholesale switch to React views but instead uses our current Backbone views and React as a templating engine. This means that some of the usual functionality of React views, such as the state object, aren't used to cause view re-render, instead the `this.changed()` function should be called to re-render a Backbone View with a React template. 

Here are a few reasons for using React as a templating engine only:
1. Our entire architecture uses Backbone events heavily - which React views don't support. 
2. Many of our bugs arise from jQuery DOM manipulation and using React to manage the DOM manipulation removes those bugs whilst introducing much more flexibility in the view <> state updates, this is as each manipulation is now automated and no longer needs to be coded by hand. 
3. Introducting jsx as a templating engine allows the upskilling requirement to be confined to templates only, which should be easier and less daunting for developers to confront. 
4. We will be able to convert to React at our leisure by not mandating a wholesale view layer rewrite. 
5. As React will be a part of the architecture we can better experiment with dropping Backbone as a large part of the architectural work is already complete.

#### Usage

A React version of the accordion lives here [pr](https://github.com/adaptlearning/adapt-contrib-accordion/pull/97) / [branch](https://github.com/adaptlearning/adapt-contrib-accordion/tree/issue/2824-jsx)

```js
// This is an example template file in **/templates/**/name.jsx
// It will get registered in Adapt by filename as handlebars templates currently do
// Templates will override from theme, menu, extension, component as usual
// There is no partials distinction with the react templates, they are all just templates which can be called from one another

export default function (model, view) {
  return <div></div> // React component
};
```

```js
// This is a template file in components/adapt-contrib-accordion/templates/accordion.jsx

// These functions help handlebars integration
import Adapt from 'core/js/adapt';
import { 
  compile, // Compile handlebars into html: compile(template, data);
  classes, // Helper for joining and filtering classes
  templates, // Lookup other react templates: templates[name](model, view);
  html, // Render html as react html:  html(`<div>new div</div>`); or html(compile('{{#if _prop}}<div>new div</div>{{/if}'));
  partial // Use a handlebars partial: partial(name, data);
} from 'core/js/reactHelpers';

export default function (model, view) {
  const data = model.toJSON();
  data._globals = Adapt.course.get('_globals');
  return <div className="component__inner accordion__inner">
    {templates.component(model, view)}
    <div className="component__widget accordion__widget">
      {data._items.map(({ _graphic, _classes, title, body, _index, _isVisited, _isActive }, index) => {
        return <div key={_index} className={classes([
              "accordion__item",
              _graphic.src && 'has-image',
              _classes && _classes
          ])} data-index={_index}>
          <button onClick={view.onClick.bind(view)} id={`${data._id}-${index}-accordion-button`} className={classes([
            'accordion__item-btn',
            'js-toggle-item',
            _isVisited && 'is-visited',
            _isActive ? 'is-open is-selected' : 'is-closed'
          ])} aria-expanded={_isActive ? 'true' : 'false'}>
            <div className="accordion__item-btn-inner">
              <div className="accordion__item-icon">
                <div className="icon"></div>
              </div>
              <div className="accordion__item-title">
                <div className="accordion__item-title-inner">
                  {html(title)}
                </div>
              </div>
            </div>
          </button>
          <div className="accordion__item-content" role="region" aria-labelledby={`${data._id}-${index}-accordion-button`}>
            <div className="accordion__item-content-inner">
              {body &&
              <div className="accordion__item-body">
                <div className="accordion__item-body-inner">
                  {html(compile(body))}
                </div>
              </div>
              }
              {_graphic.src &&
              <div className={classes([
                  'accordion__item-image-container',
                  _graphic.attribution && 'has-attribution'
              ])}>
                <img className="accordion__item-image" src={_graphic.src} aria-label={_graphic.alt} aria-hidden={_graphic.alt ? 'true' : 'false'} />
                {_graphic.attribution &&
                <div className="component__attribution accordion__attribution">
                  <div className="component__attribution-inner accordion__attribution-inner">
                    {html(_graphic.attribution)}
                  </div>
                </div>
                }
              </div>
              }
            </div>
          </div>
        </div>
      })}
    </div>
  </div>
}
```

```js
// Example from accordion.
// adaptView is extended to support react templates, so all Adapt views can use react template as below.

import ComponentView from 'core/js/views/componentView';

class AccordionView extends ComponentView {

  preRender() {
    this.listenTo(this.model.getChildren(), {
      'all': this.changed // call changed to rerender
    });
  }

  /** missing code */

}

AccordionView.template = 'accordion.jsx'; // Just update the template name to use the react template

export default AccordionView;
```

You can see the core templates are using react versions https://github.com/adaptlearning/adapt_framework/tree/issue/2824-jsx/src/core/templates which means you can update the models for page, article and block on the fly and the views will update on the page:
```js
var Adapt = require('core/js/adapt'); 
Adapt.findById('co-05').set('displayTitle', 'new title');
Adapt.findById('a-05').set('displayTitle', 'testing');
Adapt.findById('b-35').set('body', 'testing');
```

#### Release note
Remove `page.jsx`, `article.jsx` and `block.jsx` before merging as these are likely to have larger impacts on course content that should be avoided in the short-term. We should aim for component level react views to start.


grunt/config/javascript.js
```diff
@@ -18,6 +18,19 @@ module.exports = function(grunt, options) {
...
+          '../node_modules/react/umd/react.production.min.js',
+          '../node_modules/react-dom/umd/react-dom.production.min.js',
+          '../node_modules/html-react-parser/dist/html-react-parser.min.js'
+        ],
+        reactTemplates: [
+          '<%= sourcedir %>core/templates/**/*.jsx',
+          '<%= sourcedir %>components/*/templates/**/*.jsx',
+          '<%= sourcedir %>extensions/*/templates/**/*.jsx',
+          '<%= sourcedir %>menu/*/templates/**/*.jsx',
+          '<%= sourcedir %>theme/*/templates/**/*.jsx'
```
brian-learningpool, line 32: Just checking, should these be conditionally set for the plugins on use on the course rather than a wildcard?

grunt/config/javascript.js
```diff
@@ -71,6 +84,18 @@ module.exports = function(grunt, options) {
...
+          '../node_modules/object.assign/dist/browser.js',
+          '../node_modules/react/umd/react.production.min.js',
+          '../node_modules/react-dom/umd/react-dom.production.min.js'
+        ],
+        reactTemplates: [
+          '<%= sourcedir %>core/templates/**/*.jsx',
+          '<%= sourcedir %>components/*/templates/**/*.jsx',
+          '<%= sourcedir %>extensions/*/templates/**/*.jsx',
+          '<%= sourcedir %>menu/*/templates/**/*.jsx',
+          '<%= sourcedir %>theme/*/templates/**/*.jsx'
```
brian-learningpool, line 97: Same here?

grunt/config/javascript.js
oliverfoster: That is a good find. Will sort. Ta.

grunt/tasks/javascript.js
```diff
@@ -157,6 +157,16 @@ module.exports = function(grunt) {
       });
     }
 
+    // Collect react templates
+    const reactTemplatePaths = [];
+    options.reactTemplates.forEach(pattern => {
+      grunt.file.expand({
+        filter: options.pluginsFilter
```
oliverfoster, line 164: @brian-learningpool It already does the filter identical to the way the plugins currently work

brian-learningpool: Awesome.  👍

brian-learningpool:
Great work!

src/core/js/reactHelpers.js
```diff
@@ -0,0 +1,71 @@
...
+ * Storage for react templates
+ */
+export const templates = {};
+
+/**
+ * Convert html strings to react dom, equivalent to handlebars {{{html}}}
+ * @param {string} html
+ */
+export function html(html, ref = null) {
+  let node = html ? HTMLReactParser(html) : '';
```
tomgreenfield, line 18: As discussed, please fix when `undefined` is passed.

src/core/js/reactHelpers.js
```diff
@@ -0,0 +1,71 @@
...
+  const output = Handlebars.helpers[name].call(args[0]);
+  return (output && output.string) || output;
+};
+
+/**
+ * Helper for a list of classes, filtering out falsies and joining with spaces
+ * @param  {...any} args List of classes
+ */
+export function classes(...args) {
+  return args.filter(Boolean).join(' ');
```
tomgreenfield, line 70: Could this accept an array?

src/core/js/reactHelpers.js
oliverfoster: done in https://github.com/adaptlearning/adapt_framework/pull/2829/commits/9b3a7b0dd4551f56532f31aa29456266482af450

src/core/js/reactHelpers.js
oliverfoster: done in https://github.com/adaptlearning/adapt_framework/pull/2829/commits/9b3a7b0dd4551f56532f31aa29456266482af450

oliverfoster: updated core classes to match in https://github.com/adaptlearning/adapt_framework/pull/2829/commits/327a7932aa51f57c2be6a89be0a1e456cbf0a697

lc-alexanderbenesch (approved):
code 👍
testing 👍

oliverfoster:
@tomgreenfield Just waiting on you before I remove the page, article and block react templates.

src/core/js/adapt.js
```diff
@@ -1,431 +1,428 @@
...
-     * - "course"
-     * - "contentObjects"
-     * - "blocks"
-     * - "articles"
-     * - "components"
-     */
-    mapById(id) {
-      return this.data.mapById(id);
+    throw new Error('Cannot derive model class name from input');
+
```
tomgreenfield, line 291: This function isn't closed so the branch doesn't currently compile:
```shell
>> Unexpected token, expected ";" (297:33)
>> Line: 297, Col: 33, File: src/core/js/adapt.js
```

tomgreenfield (changes requested):

chucklorenz:
@olivermartinfoster per our 16/09 chat,
Console error after running `grunt build` and `grunt server`:
HTMLReactParser is not defined

oliverfoster:
Will rebase this against https://github.com/adaptlearning/adapt_framework/pull/2923 once that pr is accepted

oliverfoster:
Replaced with https://github.com/adaptlearning/adapt_framework/pull/2963

Issue #2836, opened 2020-07-16
Developer build vs final build request


robertmausser:
### Subject of the issue/enhancement/features

We use both Dev-Tools extension and Shadow-ID's extension to speed up development process of the framework

https://github.com/cgkineo/adapt-devtools

https://github.com/darylhedley/adapt-shadowIds

 

However every so often someone makes the mistake of not turning them off for the final build, which is obviously not a good thing!

Is there a way to have a line of code in JS or something that says "only build this if grunt dev command is used, not grunt build" So when we are working on the framework  with the Grunt Dev command, they show up, but on final build with Grunt Build they dont?

Any other similar method or solution is greatly appreciated, this is just how I would forsee it working. If you have a different idea that works too, im all ears

Thank you!!!

### Your environment
Adapt framework 4.4.0 and 5.0. 

Browsers etc irrelevant 

### Steps to reproduce
Not an issue, feature request.

### Expected behaviour
When the developer user Grunt Dev, devtools is available. When using Grunt Build, devtools is removed for final build. 

### Actual behaviour
Devtools is enabled even at build, causing learners to complete the course by using the devtools.

olivermartinfoster:
With the latest framework master you can now do this:

#### config.json
```json
    "build": {
        "productionExcludes": [
            "adapt-devtools"
        ]
    },
```

Issue #843, opened 2016-01-04
facilitate "development only" extensions


moloko:
a short while ago we discussed the possibility of being able to tag extensions as being 'helper' extensions i.e. they are of a kind you would not normally want to include in a production-ready build. there was some discussion of how the authoring tool could help warn the user about including such extensions in the output, ideally giving them the option to exclude some/all from the build.

moved from [ABU-693](https://adaptlearning.atlassian.net/browse/ABU-693)

PR #2837, opened 2020-07-16
issue/2836 Added ability to exclude plugin(s) from 'production' build

oliverfoster:
#2836 

To remove a plugin on `grunt build` use:

```json
    "build": {
        "productionExcludes": [
            "adapt-devtools"
        ]
    },
```

### Added 
* `Adapt.build.get('type') === 'production' || 'development';`
* config.json:`build.productionExcludes`

### DIFF
--- a/grunt/helpers.js
+++ b/grunt/helpers.js
@@ -36,7 +36,7 @@ module.exports = function(grunt) {
       grunt.log.writeln('');
     }
     if (productionExcludes) {
-      let count = productionExcludes.length;
+      const count = productionExcludes.length;
       grunt.log.writeln('The following will be excluded from the build in production:');
       for (let i = 0; i < count; i++) { grunt.log.writeln('- ' + productionExcludes[i]); }
       grunt.log.writeln('');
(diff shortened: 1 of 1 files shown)


tomgreenfield:
We already the config JSON `build.excludes` – could just add another object and leave it completely up to the course author to tinker with their build:
```json
"build": {
  "productionExcludes": [
    "adapt-devtools"
  ]
}
```

oliverfoster:
> could just add another attribute

I'm not following.

oliverfoster:
Added as requested @tomgreenfield

grunt/helpers.js
```diff
@@ -34,6 +35,12 @@ module.exports = function(grunt) {
       for (let i = 0; i < count; i++) { grunt.log.writeln('- ' + excludes[i]); }
       grunt.log.writeln('');
     }
+    if (productionExcludes) {
+      let count = productionExcludes.length;
```
moloko, line 39: minor, but `const` would be more appropriate here since the value of `count` doesn't get changed

moloko:
FYI also address #843 which I logged more than 4 years ago!

Issue #2841, opened 2020-07-23
default FW -> AT import issues - PLP settings


simondate:
### Subject of the issue/enhancement/features

When you create the framework using `adapt create course` the JSON for Page Level Progress isn't complete and breaks what should be a direct import into the Authoring Tool.

I propose updating the JSON so that a direct import is possible.

This could also be fixed by changing the PLP extension code to be more defensive (accommodating of incomplete JSON).

### Your environment
* version (Framework) v5.6.1

### Steps to reproduce
Tell us how to reproduce this issue.

1) create a course using `adapt create course`
2) Import the course into the Authoring Tool
3) Try to run the course

moloko:
> This could also be fixed by changing the PLP extension code to be more defensive (accommodating of incomplete JSON).

I agree and I think we should do this as well

olivermartinfoster:
Could we just do the PLP defensive code and not remove the indicators from the default course?

moloko:
> Could we just do the PLP defensive code and not remove the indicators from the default course?

We should definitely do the defensive code but @simondate has not actually removed the indicators from [the default course](https://adaptlearning.github.io/v5demo/#/id/co-05)

simondate:
As Matt said I made the JSON reflect what is currently in the pre-installed course.

I've never used this PLP functionality myself. I like having it half on and half off to show that there is a choice.

oliverfoster:
Sweat, I misunderstood then. Ta :+1:

oliverfoster:
do the defensive code bit

oliverfoster:
@simondate what was the original error / problem / point of failure?

simondate:
The course wouldn't run when you imported it. There was an error in the developer console that mentioned `_isCompletionIndicatorEnabled`

oliverfoster:
As you can see from this search https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/search?q=_isCompletionIndicatorEnabled&unscoped_q=_isCompletionIndicatorEnabled

https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/blob/master/js/adapt-contrib-pageLevelProgress.js#L76 This is the only place `_isCompletionIndicatorEnabled` is referenced which would never cause and error. I cannot proceed without more info.

simondate:
[...]

PR #2842, opened 2020-07-23
#2841

simondate:
Added `"_isCompletionIndicatorEnabled": false`  to every component to stop it breaking on AT import


olivermartinfoster:
Why `false`? The master course should showcase this feature. Would true not work?

moloko:
> Why `false`? The master course should showcase this feature. Would true not work?

in the master course this properly is not defined on non-interactive components and questions within the assessment topics, so all @simondate has done here is make explicit what was already set implicitly.

An argument could be made for enabling it for assessment questions - but IMO I'm not sure it should be enabled for non-interactive components seeing as they mark as completed as soon as you view them. but feel free to persuade me otherwise ;-) but all @simondate is doing here is explicitly defining the current master course settings, which seems fair enough in terms of fixing #2841

Issue #2851, opened 2020-07-29
DrawerView passes Backbone view to Adapt.a11y.focusFirst by mistake


chris-steele:
### Subject of the issue/enhancement/features
Since v5.5.0 Adapt passes a Backbone view when triggering the global `remove` event. `DrawerView` mistakenly passes this view to `Adapt.a11y.focusFirst` in its `remove` handler causing a runtime error.

### Your environment
Adapt v5.5+
all browsers
all devices

PR #2852, opened 2020-07-29
issue/2851 Fixed DrawerView passing Backbone view to Adapt.a11y.focusFirst

chris-steele:
See #2851 

### Fixed
* Added handler for external remove events fired when the drawer is open (originating from url changes, custom extensions etc)


chris-steele:
Thanks all!

Issue #2861, opened 2020-08-04
adapt_framework: underscore umd header breaks loading


oliverfoster:
### Subject of the issue
underscore loads as a require module rather than a global one.

PR #2862, opened 2020-08-04
issue/2861 Support umd underscorejs

oliverfoster:
#2861 
* Changed scriptLoader.js to support underscore as a(n) umd module


chris-steele:
:+1:

Issue #2589, opened 2019-11-18
navigation: screen reader read order and dom order bad


oliverfoster:
### Subject of the issue
Because nav buttons are floated their visual appearance is inverse to the dom order

### Environment
* Framework: >=4

kirsty-hames:
It is possible to display nav elements in their dom order if we use display flex instead of floats. 

See core changes below.

```
.nav {
  &__inner {
    display: flex;
    align-items: center;
  }

  &__back-btn {
    //.u-float-left;
  }

  &__drawer-btn {
    //.u-float-right;
    margin-left: auto;
    .dir-rtl & {
      margin-left: 0;
      margin-right: auto;
    }
  }
}

```
Instead of using floats to create the navigation split we can use margin. Both `.nav__back-btn` and `.nav__drawer-btn` are rendered even if they're not visually displayed and we can assume anything floated right would display after `.nav__drawer-btn`.

<img width="1357" alt="nav_elements_using_display_flex" src="https://user-images.githubusercontent.com/7045330/89319107-95bf0e80-d677-11ea-9fa2-82d7226ab314.png">


In addition we would no longer need the float properties on the following contrib extensions:
adapt-contrib-languagePicker
adapt-contrib-pageLevelProgress

oliverfoster:
navigation buttons now sortable with `data-order` attribute on the button node, see:
https://github.com/adaptlearning/adapt-contrib-core/blob/f4a589902c565d5749053ea3d091bda5f8aa4ed5/js/views/navigationView.js#L86-L92

needs fixing: 
https://github.com/adaptlearning/adapt-contrib-core/blob/f4a589902c565d5749053ea3d091bda5f8aa4ed5/schema/course.schema.json#L316

https://github.com/adaptlearning/adapt-contrib-core/blob/f4a589902c565d5749053ea3d091bda5f8aa4ed5/templates/nav.hbs#L17

PR #2864, opened 2020-08-05
issue/2589 Added navigation item management

oliverfoster:
#2589 

### Added
* API for creating navigation items
  * Left, middle, right containers
  * Sort order

### DIFF
--- a/src/core/js/models/navigationItemModel.js
+++ b/src/core/js/models/navigationItemModel.js
@@ -1,0 +1,17 @@
+define(function () {
+
+  class NavigationItemModel extends Backbone.Model {
+
+    defaults() {
+      return {
+        _name: '',
+        _order: -1,
+        _template: null
+      };
+    }
+
+  }
+
+  return NavigationItemModel;
+
+});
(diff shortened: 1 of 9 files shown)


chris-steele:
:+1:

tomgreenfield:
Looking good, @oliverfoster.

My feeling at this point in time for config would be to get the course author to hard-code the sequence of navbar items in a global location such as course.json. This would take the guess work out of where a plugin's functionality might be appended with the z-index style order system.

```json
"_navigation": {
  "_left": [
    "back"
  ],
  "_middle": [
    "navigationLogo"
  ],
  "_right": [
    "drawer",
    "close"
  ]
}
```

You mentioned possible further complexity where one single plugin might want to add several different icons to the navbar in different places – we might think about addressing this by getting a plugin's schema to define IDs or labels for each nav item for which the author can add to the global nav config. I assume the default label would be a property from the bower JSON for simplicity.

All up for debate and formalisation, of course.

oliverfoster:
closed temporarily to make way for https://github.com/adaptlearning/adapt_framework/issues/3169

Issue #2865, opened 2020-08-05
Core + Vanilla - Inconsistent responsive triggers


guywillis:
### Subject of the issue
There is an inconsistent trigger between the JS and the CSS responsively. The inconsistency appears to trigger at the medium break point exactly (though this could also be true for the small and large break points too). 

This is easiest to see when looking at the narrative component in the screenshots below. At 759 width the single column, mobile view is visible with the mobile instruction and strapline. At 760 width the narrative changes to the desktop view but still has the mobile instruction. At 761 the mobile instruction text swaps out for the desktop instruction.

### Your environment
* FW v5.6.1
* Chrome Version 84.0.4147.105 (Official Build) (64-bit)
* Mac OSX v10.15.3

### Steps to reproduce
Scale a browser between 759 and 761 px width.

### Expected behaviour
The JS and CSS responsive triggers should be aligned and trigger together or converted to one method.

### Screenshots (if you can)
![759](https://user-images.githubusercontent.com/11569678/89409443-facb4080-d719-11ea-8143-c61ee8b60f70.png)
![760](https://user-images.githubusercontent.com/11569678/89409444-fb63d700-d719-11ea-91de-9135dd6a63bc.png)
![761](https://user-images.githubusercontent.com/11569678/89409446-fb63d700-d719-11ea-9f3b-0e1ed534197e.png)

moloko:
from what I can figure out from my investigations into this, I think it boils down to the media queries considering anything >= 760 to be 'large' whereas the JS is set to anything > 760

So question is, which one do we change?

oliverfoster:
whichever requires fewest changes?

moloko:
The JavaScript I assume, surely just a case of changing `>` to `>=` in the `if...else if` conditions [here](https://github.com/adaptlearning/adapt_framework/blob/b6882dba282625b5100f9962e26802cdd331a04d/src/core/js/device.js#L73-L82)

oliverfoster:
sounds about right

oliverfoster:
@guywillis ?

guywillis:
Quick test locally seems to confirm it'll work but I'll give it a more thorough test tomorrow

kirsty-hames:
Issue resolved with both PRs now being merged.

PR #2874, opened 2020-08-11
Amend js responsive trigger

guywillis:
Related issue: https://github.com/adaptlearning/adapt_framework/issues/2865

* Amended js responsive screen size trigger


guywillis:
The introduction of the equals in the if statement treats the screen sizes as follows:

519 and below = small
520 - 759 = medium
760 and above = large

These values can then be synced with the min width medias queries for a uniformed responsive trigger.

PR #2881, opened 2020-08-16
Added Start controller

simondate:


src/course/en/course.json
```diff
@@ -116,6 +116,18 @@
             }
         ]
     },
+    "_start": {
+        "_isEnabled": false,
+        "_startIds": [
+            {
+                "_id": "co-05",
+                "_skipIfComplete": false,
```
moloko, line 124: I think this might be the more common use-case?
```suggestion
                "_skipIfComplete": true,
```

Issue #2884, opened 2020-08-18
adapt_framework: bug view-only question components use incorrect model


oliverfoster:
### Subject of the issue
The new model/view registration process has a hole whereby when registering a named legacy component with a `QuestionView` only (deprecated) the `ComponentModel` is chosen rather than automatically assigning a `QuestionModel`.

For backward compatibility this needs to be fixed.

PR #2885, opened 2020-08-18
issue/2884 Automatic detection of view-only question models

oliverfoster:
#2884 

### Added
* Automatic detection of view-only question models

### Changed
* Registered the abstract question model


moloko (approved):

cahirodoherty-learningpool (approved):
👀

lc-alexanderbenesch (approved):
👍

Issue #2886, opened 2020-08-19
Firefox - Triggering a Notify popup causes sudden page jump


joe-replin:
Using a fresh v5.6.1 Adapt install, triggering a notify panel causes a sudden jump to the top of the page. This happens with both presentation & question components.

### Your environment
* Adapt Framework - v5.6.1
* **Only present in Mozilla Firefox v79**
* Dell Inspiron + Windows 10

![Screen Shot 2020-08-19 at 15 47 30](https://user-images.githubusercontent.com/7740674/90688016-6b528100-e233-11ea-9ec8-4016a764db09.png)

oliverfoster:
@joe-replin could you try out the above pr please?

joe-replin:
Thank you, @oliverfoster for pushing this as well as for the reference links. I have tested in 2 separate projects and Firefox is not misinterpreting that focus change anymore.

oliverfoster:
Could you accept the changes in the pr? @joe-replin 
Or add a +1 if you don't have access to the "review changes" button on the "files changed" tab.
https://github.com/adaptlearning/adapt_framework/pull/2889/files

oliverfoster:
Will get closed when the pr goes through.

PR #2889, opened 2020-08-20
issue/2886 preventScroll fix now excludes firefox

oliverfoster:
#2886 

Firefox misreports `$(window).scrollTop()` frequently, returning 0 in unusual circumstances. The `preventScroll` fix for `Element.focus` - for browsers which don't support the `preventScroll` option, ie11 and safari at the time of writing - is necessary but should not be applied to Firefox where it often causes the scroll to bounce back to 0.

### Changed
* Stopped `preventScroll` fix applying in Firefox

References:
Browser compatibility - https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/focus#Browser_compatibility
Safari support ticket for preventDefault - https://bugs.webkit.org/show_bug.cgi?id=178583

Notes:
Have yet to identify why Firefox misreports the `$(window).scrollTop()`, it reads fine in the debugger console or watch window. The bug seems to appear at frequency in Firefox and then disappears again. The `scrollTop` overrides from `Adapt.scrolling` were not applied, so this is either a jQuery or Firefox bug.

### DIFF
--- a/src/core/js/a11y.js
+++ b/src/core/js/a11y.js
@@ -617,8 +617,8 @@ define([
             // Drop focus errors as only happens when the element
             // isn't attached to the DOM.
           }
-          if (Adapt.device.browser === 'internet explorer') {
-            // ie 11 doesn't support preventScroll but this breaks firefox 79.0
+          if (Adapt.device.browser !== 'firefox') {
+            // ie 11 and safari doesn't support preventScroll but this breaks firefox 79.0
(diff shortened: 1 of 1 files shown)


moloko (approved):

cahirodoherty-learningpool (approved):
👀

lc-alexanderbenesch (approved):
👍

Issue #2908, opened 2020-09-09
Adapt.mapById no longer works


moloko:
As it calls `this.data.mapById` - which does not exist

The only core plugin that used this was adapt-contrib-languagePicker

moloko:
will be removed in v5.8.0

PR #2923, opened 2020-09-23
issue/2824 Switch to rollup from requirejs

oliverfoster:
#2824 
This version does not have the react elements, it is merely to fix the currently broken build process in v5.  

Will provide an almost seamless migration away from build-side [requirejs](https://requirejs.org/docs/optimization.html) and [AMD](https://requirejs.org/docs/whyamd.html) modules to [ES6](https://hacks.mozilla.org/2015/08/es6-in-depth-modules/) modules.  

#### Facilitates
* future proof
  * beyond ES8 code
* development improvements
  * supports larger code bases
* backward-compatibility
  * ES6 modules are reregistered in the browser as AMD modules, so `var Adapt = require('core/js/adapt');` etc will still work in the console and for other plugins
  * still ie11 compatible

#### Downsides
* Currently this will only work with transpilation back to ES5 as calling an ES6 `constructor` function without the `new` keyword is not allowed - which is an essential part of the way Backbone constructs classes.

#### Added
* [babel-plugin-transform-amd-to-es6](https://github.com/cgkineo/babel-plugin-transform-amd-to-es6) (custom made) to convert [AMD](https://requirejs.org/docs/whyamd.html) modules to [ES6](https://hacks.mozilla.org/2015/08/es6-in-depth-modules/) modules

#### Removed

#### Fixed

#### Testing
* Use master of all plugins by running:
```js
git clone https://github.com/adaptlearning/adapt_framework 2824
cd 2824
git checkout issue/2824
grunt dev
```

#### Example

rollup-es-modules2


cahirodoherty-learningpool (approved):
👀

src/core/js/adapt.js
```diff
@@ -1,431 +1,429 @@
...
+   * @return {string} One of the following (or `undefined` if not found):
+   * - "course"
+   * - "contentObjects"
+   * - "blocks"
+   * - "articles"
+   * - "components"
+   */
+  mapById(id) {
+    return this.data.mapById(id);
+  }
```
moloko, line 324: suggest either removing this or at least adding a deprecated warning

https://github.com/adaptlearning/adapt_framework/issues/2908

oliverfoster: Can I do that in a different pr?

oliverfoster: Have removed.

cahirodoherty-learningpool (approved):
👀

chris-steele:
Looks fantastic, only issue I saw was a `childViews` reference runtime error with master branch of [adapt-contrib-boxMenu](https://github.com/adaptlearning/adapt-contrib-boxMenu), but works fine using latest release of said menu. :+1:

chris-steele (approved):

oliverfoster:
@moloko Could you +1 when you get a mo?

PR #2934, opened 2020-10-06
issue/2743 - remove the flush for Adapt.wait

### LINKED ISSUE #2743: Adapt.wait flush
Is the behaviour implemented via https://github.com/adaptlearning/adapt_framework/issues/2439 a good idea? Isn't it the responsibility of the plugin to ensure they call an `Adapt.wait.end()` if using an `Adapt.wait.begin()`.

Flushing any open waits after 7 seconds means you can't intentionally block Adapt from loading, if required. For example, you may want to not call an `end()` and display an uncancellable notification to indicate an issue which should prevent a user from continuing. Currently the flush will end this, continue with the loading and eventually close any notifications regardless of the plugin functionality.

danielghost:
resolves #2743

### DIFF
--- a/src/core/js/wait.js
+++ b/src/core/js/wait.js
@@ -2,63 +2,29 @@ define(function() {
 
   var Wait = Backbone.Controller.extend({
 
+    _waitCount: 0,
+    _callbackHandle: null,
+
     initialize: function() {
       _.bindAll(this, 'begin', 'end');
     },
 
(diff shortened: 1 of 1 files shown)


oliverfoster (approved):

moloko (approved):
code 👓

cahirodoherty-learningpool (approved):
👀

Issue #2938, opened 2020-10-09
adapt-contrib-narrative - Hotgraphic displayed twice after window resize.


deltanetdan:
### Subject of the issue
If two or more hotgraphics are used on the same page resizing the window will result in multiple hotgraphics being displayed

### Your environment
* version (AT/Framework) 5.7.0
* which browser and its version Firefox 81
* device(s) + operating system(s) macOS 10.14.6

### Steps to reproduce
1 - Create a course using adapt cli. 
2 - Edit the default course duplicating the hotgraphic on the first page.
3 - Preview the course in the browser with a large screen size 
4 - Reduce the screen size to small, hotgraphics are replaced with narratives
5 - Increase screen size to large, hotgraphics are re-rendered correctly but the second hotgraphic is duplicated.

### Expected behaviour
Hotgraphics should be rendered correctly on resize without being duplicated.

### Actual behaviour
Duplicate of the second hotgraphic could be due to the first narrative(hotgraphic) view triggering [replaceWithHotgraphic](https://github.com/adaptlearning/adapt-contrib-narrative/blob/master/js/narrativeView.js#L157) on the second.

Propose only running replaceWithHotgraphic on device:changed rather that device:changed and [device:resize](https://github.com/adaptlearning/adapt-contrib-narrative/blob/master/js/narrativeView.js#L25).

PR #2949, opened 2020-10-20
Added stop listener to remove function

guywillis:
Resolves https://github.com/adaptlearning/adapt_framework/issues/2938

* Stop listening to the event listeners in the remove function

### DIFF
--- a/src/core/js/views/adaptView.js
+++ b/src/core/js/views/adaptView.js
@@ -288,6 +288,7 @@ define([
       this.preRemove();
       Adapt.trigger(`${type}View:remove view:remove`, this);
       this._isRemoved = true;
+      this.stopListening();
 
       Adapt.wait.for(end => {
         this.$el.off('onscreen.adaptView');


cahirodoherty-learningpool (approved):
👀

moloko (commented):
that's odd, I would have thought that `super.remove` on line 291 would handle this (since [Backbone.View.remove](https://backbonejs.org/docs/backbone.html#section-175) calls `this.stopListening`... is it something to do with that being wrapped in `Adapt.wait`? Just checking we're not missing a wider issue here...

oliverfoster:
Yea. It's because the `Backbone.View.remove` `stopListening` now happens asynchronously. The `Backbone.View.remove` was deferred here to account for DOM removal. It seems that the hotgrpahic/narrative problem relies on an implied synchronous `stopListening` in the `View.remove` call but that by deferring the `Backbone.VIew.remove` the impact of moving the `stopListening` wasn't considered. The view listeners can be removed on the synchronous and the DOM on the asynchronous.

oliverfoster (approved):

moloko (approved):

chetan-hajare (approved):

PR #2951, opened 2020-10-22
issue/2950 Switch from disabled to aria-disabled for ButtonsView

### LINKED ISSUE #2950: adapt_framework: Disabled submit button should be available to screen reader
### Subject of the issue
The user is unaware of the existence of a button as `button[disabled]` elements are not available to the screen reader.

oliverfoster:
resolves #2950 

### Changed
* Made submit and feedback button available to the screen reader when disabled by switching from `disabled="disabled"` to `aria-disabled="true" tabindex="-1"`. The screen reader instead of reading nothing now reads "disabled button" or "dimmed button".

Reference:
https://a11y-101.com/development/aria-disabled

### DIFF
--- a/src/core/js/a11y.js
+++ b/src/core/js/a11y.js
@@ -235,9 +235,12 @@ define([
       }
       isEnabled = isEnabled === undefined ? true : isEnabled;
       if (!isEnabled) {
-        $elements.attr('disabled', 'disabled').addClass('is-disabled');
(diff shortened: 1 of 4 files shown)


moloko (approved):
code 👓

cahirodoherty-learningpool (approved):
👀

eleanor-heath (approved):

PR #2970, opened 2020-11-10
Convert port string to number

### LINKED ISSUE #2954: grunt server: 'port' switch doesn't work
According to the documentation in [grunt.md](https://github.com/adaptlearning/adapt_framework/blob/master/GRUNT.md) I should be able to change the port number used by the `grunt server` command by doing `grunt server --port=9002` - but when I run the command it errors with 'Fatal error: Port 9002 is already in use by another process'. It does this no matter what port number I choose, even if I use 9001 (which is what it defaults to normally)

FW: v5.7.1
Node: v10.13.0
grunt-cli: v1.2.0
grunt: v1.1.0

Tested on both Windows 10 and macOS 10.15

tomgreenfield:
Reference: https://github.com/gruntjs/grunt-contrib-connect/issues/254.

Resolves #2954.

### DIFF
--- a/grunt/config/connect.js
+++ b/grunt/config/connect.js
@@ -1,6 +1,6 @@
 module.exports = function(grunt, options) {
 
-  var port = grunt.option('port') || 9001;
+  var port = parseInt(grunt.option('port')) || 9001;
   var host = grunt.option('host') || 'localhost';
 
   return {


moloko (approved):

oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

PR #2978, opened 2020-11-20
move conversion of _isResetOnRevisit from questionView.js to adaptModel.js

### LINKED ISSUE #2825: Framework - questionView checkIfResetOnRevisit function expects boolean

moloko:
resolves #2825 

refactor to remove nested if..else statements from `checkIfResetOnRevisit` in questionView.js

### DIFF
--- a/src/core/js/data.js
+++ b/src/core/js/data.js
@@ -13,9 +13,9 @@ define([
     model(json) {
       const ModelClass = Adapt.getModelClass(json);
       if (!ModelClass) {
-        return new Backbone.Model(json);
+        return new Backbone.Model(json, { parse: true });
       }
-      return new ModelClass(json);
+      return new ModelClass(json, { parse: true });
     }
 
     initialize() {
--- a/src/core/js/data.js
+++ b/src/core/js/data.js
@@ -13,7 +13,7 @@ define([
     model(json) {
       const ModelClass = Adapt.getModelClass(json);
       if (!ModelClass) {
-        return new Backbone.Model(json, { parse: true });
+        return new Backbone.Model(json);
       }
       return new ModelClass(json, { parse: true });
     }
--- a/src/core/js/models/adaptModel.js
+++ b/src/core/js/models/adaptModel.js
@@ -755,7 +755,12 @@ define([
      * default values.
      */
     checkIfResetOnRevisit() {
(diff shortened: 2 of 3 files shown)


tomgreenfield (commented):
Should this instead be handled on model instantiation?
```js
parse(data) {
  if (data._isResetOnRevisit === 'false') {
    data._isResetOnRevisit = false;
  }
  return data;
}
```
Pass `{ parse: true }` as a second parameter [here](https://github.com/adaptlearning/adapt_framework/blob/5bca8daf160fc3ce52cf0dff1307a6fd1246f447/src/core/js/data.js#L18) to trigger the function.

moloko:
> Should this instead be handled on model instantiation?

Ooh that's a nice idea. I'll have a look at this on Monday.

src/core/js/data.js
```diff
@@ -13,9 +13,9 @@ define([
     model(json) {
       const ModelClass = Adapt.getModelClass(json);
       if (!ModelClass) {
-        return new Backbone.Model(json);
+        return new Backbone.Model(json, { parse: true });
```
tomgreenfield, line 16: This has no effect on the default `Backbone.Model`.

tomgreenfield (approved):
Works for me.

cahirodoherty-learningpool (approved):
👀

tomgreenfield (approved):

oliverfoster (approved):

PR #3000, opened 2020-12-17
issue/2999 Converted to ES6 classes and modules and let/const

### LINKED ISSUE #2999: adapt_framework: convert to es6 classes and modules
Remove all amd `define`/`return` and replace with `import`/`export default`.
Convert classic `var Class = Backbone.Class.extend({})` to ES6 `class Class extends Backbone.Class {}`

oliverfoster:
fixes #2999 

Please view with **whitespace off**.

This effectively enables intellisense in VSCode for the core framework classes and APIs (hopefully).

Tested in IE11 - working.

#### Changes
* Removed all amd `define`/`return` and replace with `import`/`export default`.
* Converted classic `var Class = Backbone.Class.extend({})` to ES6 `class Class extends Backbone.Class {}`
* Refactor `core/js/device.js`
* `var` to `let`/`const`
* Moved `LockingModel` into the inheritance chain rather than keep it as a duck-punch/money-patch

### DIFF
--- a/src/core/js/a11y.js
+++ b/src/core/js/a11y.js
@@ -1,23 +1,18 @@
-define([
-  'core/js/adapt',
-  './a11y/browserFocus',
-  './a11y/focusOptions',
-  './a11y/keyboardFocusOutline',
-  './a11y/log',
-  './a11y/scroll',
-  './a11y/wrapFocus',
-  './a11y/popup',
-  './a11y/deprecated'
-], function(Adapt, BrowserFocus, FocusOptions, KeyboardFocusOutline, Log, Scroll, WrapFocus, Popup) {
+import Adapt from 'core/js/adapt';
+import BrowserFocus from 'core/js/a11y/browserFocus';
+import FocusOptions from 'core/js/a11y/focusOptions';
+import KeyboardFocusOutline from 'core/js/a11y/keyboardFocusOutline';
+import Log from 'core/js/a11y/log';
+import Scroll from 'core/js/a11y/scroll';
+import WrapFocus from 'core/js/a11y/wrapFocus';
+import Popup from 'core/js/a11y/popup';
 
-  var A11y = Backbone.Controller.extend({
+import 'core/js/a11y/deprecated';
 
-    $html: $('html'),
-    _htmlCharRegex: /&.*;/g,
+class A11y extends Backbone.Controller {
 
-    config: null,
-    defaults: {
-
+  defaults() {
+    return {
       _isFocusOutlineKeyboardOnlyEnabled: true,
       /**
        * `_isFocusOutlineDisabled` ignores `_isEnabled` and can be used when all other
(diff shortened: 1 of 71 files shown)


src/core/js/a11y/popup.js
```diff
@@ -1,225 +1,218 @@
...
-        },
-        'popup:closed': function($target, ignoreInternalTrigger) {
-          if (ignoreInternalTrigger) {
-            return;
-          }
-          Adapt.a11y.log.deprecated('Adapt.trigger("popup:closed", $target) is replaced with Adapt.a11y.popupClosed($target);');
-          this.closed($target, true);
+    this._elementUIDIndex = 0;
+    this.listenTo(Adapt, {
+      'popup:opened'($element, ignoreInternalTrigger) {
```
cahirodoherty-learningpool, line 41: Is it okay to set a function like this?

cahirodoherty-learningpool (commented):
👀  Looks good. Just one question left on popup.js

src/core/js/a11y/popup.js
oliverfoster: Apparently so. No linting, nor parsing nor transpiring errors... So just a question of preference or conflicting standards at this point I think?

cahirodoherty-learningpool (approved):
👀

src/core/js/collections/notifyPushCollection.js
```diff
@@ -1,53 +1,49 @@
...
+      this.showPush(model);
     }
+  }
 
-  });
-
-  return NotifyPushCollection;
+  canShowPush() {
+    var availablePushNotifications = this.where({ _isActive: true });
+    if (availablePushNotifications.length >= 2) {
```
eleanor-heath, line 30: could do a ternary statement here

It might also be worth declaring all variables with const or let depending on the type of variable used to make them block scoped?

src/core/js/collections/notifyPushCollection.js
```diff
@@ -1,53 +1,49 @@
...
+    this.listenTo(this, 'add', this.onPushAdded);
+    this.listenTo(Adapt, 'notify:pushRemoved', this.onRemovePush);
+  }
+
+  onPushAdded(model) {
+    this.checkPushCanShow(model);
+  }
+
+  checkPushCanShow(model) {
+    if (this.canShowPush()) {
```
eleanor-heath, line 22: you could do an early return here with if (!this.canShowPush()) return

src/core/js/device.js
```diff
@@ -1,226 +1,223 @@
...
-    // larger than small em breakpoint. If so apply medium, otherwise apply small
-    if (screenSizeEmWidth >= mediumEmBreakpoint) {
-      screenSize = 'large';
-    } else if (screenSizeEmWidth >= smallEmBreakpoint) {
-      screenSize = 'medium';
-    } else {
-      screenSize = 'small';
-    }
+Adapt.once('app:dataReady', function() {
+  Adapt.device.screenSize = checkScreenSize();
```
eleanor-heath, line 34: the contents of this function and the below functions `checkScreenSize`, `getScreenWidth`, `getScreenWidth` etc could all be made into a class.

src/core/js/helpers.js
```diff
@@ -1,416 +1,411 @@
...
+  '_article': 2,
+  '_block': 3,
+  '_component': 4,
+  '_componentItem': 5,
+  '_notify': 1
+};
+
+const helpers = {
+
+  lowercase: function(text) {
```
eleanor-heath, line 17: could possibly put all these helper functions into a Helpers class?

src/core/js/models/adaptModel.js
```diff
@@ -1,859 +1,853 @@
...
-    }
+import Adapt from 'core/js/adapt';
+import ModelEvent from 'core/js/modelEvent';
+import 'core/js/logging';
+
+export default class AdaptModel extends Backbone.Model {
+
+  toJSON() {
+    // Perform shallow clone
+    const json = _.clone(this.attributes);
```
eleanor-heath, line 9: you could use an object Spread here to make a shallow copy of this.attributes

src/core/js/models/adaptModel.js
```diff
@@ -1,859 +1,853 @@
...
+      'all': this.onAll,
+      'bubble': this.bubble,
+      'change:_isReady': this.checkReadyStatus,
+      'change:_isComplete': this.onIsComplete,
+      'change:_isInteractionComplete': this.checkInteractionCompletionStatus
+    });
+  }
 
-      return _.pick(...args);
+  init() {}
```
eleanor-heath, line 166: I think there might need to be a space between the semicolons? `init() { }`

src/core/js/models/adaptModel.js
```diff
@@ -1,859 +1,853 @@
...
+  checkCompletionStatus() {
+    // defer to allow other change:_isComplete handlers to fire before cascading to parent
+    Adapt.checkingCompletion();
+    _.defer(this.checkCompletionStatusFor.bind(this), '_isComplete');
+  }
 
-      if (!options) {
-        return returnedDescendants;
-      }
+  checkInteractionCompletionStatus() {
```
eleanor-heath, line 264: This is not required but you could make a single function that carries out the tasks of `checkCompletionStatus` and `checkInteractionCompletionStatus` and pass through `_isComplete` and `_isInteractionComplete` as a parameter, e.g.:

```
checkCompletionStatus(competionType) {
    // defer to allow other change:_isComplete handlers to fire before cascading to parent
    Adapt.checkingCompletion();
    _.defer(this.checkCompletionStatusFor.bind(this), competionType);
  }
  ```

File: src/core/js/models/componentModel.js [hunk 9]
```diff
@@ -1,186 +1,182 @@
...
-      if (!this.get('_canReset') && !force) return;
-      this.resetUserAnswer();
-      super.reset(type, force);
-    }
+  /**
+   * Restore the user's answer from the _userAnswer property.
+   * The _userAnswer value must be in the form of arrays containing
+   * numbers, booleans or arrays only.
+   */
+  restoreUserAnswers() {}
```
eleanor-heath, line 67: could put a space in between the curly bracket for `restoreUserAnswers() { }` and `storeUserAnswer() { }`

src/core/js/models/lockingModel.js
```diff
@@ -1,138 +1,134 @@
...
-      } else {
-        delete this._lockedAttributesValues[attrName][options.pluginName];
-      }
-
-      return this;
+const set = Backbone.Model.prototype.set;
 
+Object.assign(Backbone.Model.prototype, {
+
+  set: function(attrName, attrVal, options = {}) {
```
eleanor-heath, line 5: could put these into a LockingModel class

src/core/js/models/questionModel.js
```diff
@@ -1,343 +1,339 @@
...
 
-    // Checks if the question should be set to complete
-    // Calls setCompletionStatus and adds complete classes
-    checkQuestionCompletion() {
+  // Should return a boolean based upon whether to question is correct or not
+  isCorrect() {}
 
-      const isComplete = (this.get('_isCorrect') || this.get('_attemptsLeft') === 0);
+  // Used to set the score based upon the _questionWeight
+  setScore() {}
```
eleanor-heath, line 147: add a space for curly brackets `isCorrect() { }` `setScore() { }`

src/core/js/views/drawerView.js
```diff
@@ -1,285 +1,288 @@
...
+    this.$('.drawer__holder').html(view);
+  }
+
+  checkIfDrawerIsAvailable() {
+    if (this.collection.length === 0) {
+      $('.js-nav-drawer-btn').addClass('u-display-none');
+      Adapt.trigger('drawer:noItems');
+      return;
+    }
+    $('.js-nav-drawer-btn').removeClass('u-display-none');
```
eleanor-heath, line 99: could do a toggleClass here 
```
const isEmptyDrawer = this.collection.length === 0;
$('.js-nav-drawer-btn').toggleClass('u-display-none', isEmptyDrawer);
if (isEmptyDrawer) {
   Adapt.trigger('drawer:noItems');
   return;
}
```

src/core/js/views/drawerView.js
```diff
@@ -1,285 +1,288 @@
...
-        }
-        Adapt.trigger('drawer:openedCustomView');
+        this.renderItems();
+        Adapt.trigger('drawer:openedItemView');
+      }
+    } else {
+      if (this._hasBackButton && this.collection.models.length > 1) {
+        this.$('.drawer__back').removeClass('u-display-none');
+      } else {
+        this.$('.drawer__back').addClass('u-display-none');
```
eleanor-heath, line 160: could do a toggleClass here `this.$('.drawer__back').toggleClass('u-display-none', this._hasBackButton && this.collection.models.length > 1)`

src/core/js/views/notifyView.js
```diff
@@ -1,268 +1,262 @@
...
-      event.preventDefault();
-      // tab index preservation, notify must close before subsequent callback is triggered
-      this.closeNotify();
-      Adapt.trigger($(event.currentTarget).attr('data-event'), this);
-    }
+  render() {
+    const data = this.model.toJSON();
+    const template = Handlebars.templates.notify;
+    // hide notify container
+    this.$el.css('visibility', 'hidden');
```
eleanor-heath, line 61: could add and remove the `u-visibility-hidden` class based on if the element should be displayed.

eleanor-heath:
Looks good to me. I left a few suggestions in the code, but these could be implemented at a later stage, though it might be worthwhile declaring all variables with const or let and putting classless functions into a class?

src/core/js/helpers.js
oliverfoster: It would make it somewhat difficult to do the `Handlebars.registerHelper(name, helpers[name]);` bit at the bottom because the functions wouldn't be enumerable on the instance.

src/core/js/models/adaptModel.js
oliverfoster: It seems fine in the linting and compiling. It's one of those functions designed for override.

src/core/js/models/adaptModel.js
oliverfoster: I'm not going to change the API here, but you do have a good point. I'd have to put deprecated warnings in etc for the missing function.

src/core/js/models/componentModel.js
oliverfoster: Same as above. I think this might be a personal preference thing?

src/core/js/models/lockingModel.js
oliverfoster: This is a bit ambiguous, it's actually a hack to override the set and get functions on a backbone model to implement property locking. I agree it shouldn't live here. I'll have a think.

src/core/js/models/questionModel.js
oliverfoster: Same as above.

src/core/js/views/notifyView.js
oliverfoster: This is rendering effect to produce DOM area and no flashes whilst rendering, the `visibility` value is only set and unset in this function.

eleanor-heath (approved):

src/core/js/models/lockingModel.js
oliverfoster: I'll just add a header comment to this file to clarify what it's for if that's ok for the moment?

eleanor-heath: of course! that would be helpful

oliverfoster: It was actually more self explanatory to add `LockingModel` into the inheritance chain where it's expected rather than try to quantify it in words.

eleanor-heath: Looks far better and more easily read.

oliverfoster:
@eleanor-heath could you rereview?

cahirodoherty-learningpool (approved):
👀

eleanor-heath (approved):

src/core/js/views/notifyPushView.js
```diff
@@ -1,124 +1,122 @@
...
+          this.updatePushPosition();
         }
       }
-    },
-
-    updatePushPosition: function () {
-      if (this.hasBeenRemoved) {
-        return;
-      }
+    }
```
tomgreenfield, line 99: Early returns for this function? Something along the lines of:
```js
if (this.hasBeenRemoved) return;
const models = this.model.collection.models;
models.forEach((model, index) => {
  if (!model.get('_isActive')) return;
  model.set('_index', index);
  this.updatePushPosition();
});
```

src/core/js/views/notifyPushView.js
```diff
@@ -1,124 +1,122 @@
-define([
-  'core/js/adapt'
-], function (Adapt) {
+import Adapt from 'core/js/adapt';
```
tomgreenfield, line 1: ⚠️ With this PR I can no longer seem to trigger push notifications e.g.
```js
require('core/js/adapt').notify.push({ title: 'Hello world' });
```

src/core/js/wait.js
```diff
@@ -1,102 +1,96 @@
...
+  /**
+   * Wait for this asynchronous function to execute before triggering ready event.
+   *
+   * @param  {Function} callback   [ Function to execute whilst holding queued callback. Once complete run first argument, done(). ]
+   * @return {Object}
+   */
+  for(callback) {
+    this.begin();
+
+    _.defer(function() {
```
tomgreenfield, line 89: Arrow function?

tomgreenfield (commented):
A sprinkling of ES5/6 suggestions.

Also a conflict to be sorted.

src/core/js/a11y.js
```diff
@@ -59,353 +54,389 @@ define([
...
+  }
+
+  _onNavigationStart() {
+    if (!this.isEnabled()) {
+      return;
+    }
+    // Stop document reading
+    _.defer(function() {
+      Adapt.a11y.toggleHidden('.contentobject', true);
+    });
```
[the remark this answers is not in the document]
moloko: We could probably do with setting up a bit in the style guide about use of arrow functions. I've seen a few that argue that you shouldn't use a one-liner arrow function in this situation since:
* there's no need to bind the scope of `this`
* there's no implicit `return`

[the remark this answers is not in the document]
tomgreenfield: Fair dos! Personally, I favour readability over anything else – find it quicker to scan without the keyword/brackets etc., but is just an opinion.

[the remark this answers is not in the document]
oliverfoster: I don't know about those points. It sounds like a really weird way of rationalising the 'I really don't like the look of this new syntax' stance.

* there's no need to bind the scope of this

Every function has a call time `this` binding. It's the difference between:
```js
var spartacus = function() { 
  console.log(this.name || "he's a very naughty boy!"); 
};
var a = { name: "i am spartacus", spartacus };
var b = { name: "no! i am spartacus", spartacus };
a.spartacus();
b.spartacus();
spartacus();
```
Where the same function has different `this` context based upon the implied calling context,

* there's no implicit return
All functions have an implied `return undefined;`, whether we use the return value or not.

src/core/js/views/notifyPushView.js
oliverfoster: `Collection.prototype.model` can't be a function, apparently.

src/core/js/device.js
```diff
@@ -110,11 +110,11 @@ define([
...
     if (os === '') {
       // Fall back to using navigator.platform in case Bowser can't detect the OS.
-      var platform = navigator.platform;
-      var platforms = ['Win', 'Mac', 'Linux'];
+      const platform = navigator.platform;
+      const platforms = ['Win', 'Mac', 'Linux'];
       os = 'PlatformUnknown';
 
-      for (var j = 0; j < platforms.length; j++) {
+      for (let j = 0; j < platforms.length; j++) {
```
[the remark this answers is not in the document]
tomgreenfield: Bump.

src/core/js/headings.js
```diff
@@ -1,26 +1,22 @@
...
-    onViewRender: function(view) {
-      var $headingSeats = view.$('.js-heading');
-      $headingSeats.each(function(index, el) {
-        new HeadingView({
-          el: el,
-          model: view.model
-        });
+  onViewRender(view) {
+    const $headingSeats = view.$('.js-heading');
+    $headingSeats.each(function(index, el) {
```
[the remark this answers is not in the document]
tomgreenfield: Suggest:
```js
$headingSeats.each((index, el) => new HeadingView({ el, model: view.model }));
```

src/core/js/a11y/browserFocus.js
```diff
@@ -1,89 +1,82 @@
...
-
-  return BrowserFocus;
+  /**
+   * Force focus when clicked on a tabbable element,
+   * making sure `document.activeElement` is updated.
+   * Stop event handling on aria-disabled elements.
+   *
+   * @param {JQuery.Event} event
+   */
+  _onClick: function(event) {
```
tomgreenfield, line 63: Botched merge – should be class function syntax.

src/core/js/device.js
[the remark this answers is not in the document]
oliverfoster: Its more this below innit?
```js
os =  (platforms.find(name => platform.includes(name)) || '').toLowerCase() || os;`
```
Platform will be a longer browser string which should contain one of the platforms sub strings to match?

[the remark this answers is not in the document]
oliverfoster:
```js
      // Fall back to using navigator.platform in case Bowser can't detect the OS.
      const platform = navigator.platform.toLowerCase();
      const match = platform.match(/win|mac|linux/);
      if (match) os = match[0];
      // Set consistency with the Bowser flags.
      if (os === 'win') os = 'windows';
      if (!os) os = 'PlatformUnknown';
```

tomgreenfield (approved):

PR #3016, opened 2021-01-13
issue/3014 Fixed reset on revisit completion cascade

### LINKED ISSUE #3014: adapt_framework: Reset on revisit isn't complete before page render
### Subject of the issue

The code to reset on revisit now lives in the preRender function of the PageView.
https://github.com/adaptlearning/adapt_framework/blob/657fea688ffbc2ded888f882a514e443be4574fd/src/core/js/views/pageView.js#L8-L11

It causes an asynchronous cascade of resets and completion check in AdaptModel from the components to the page.
https://github.com/adaptlearning/adapt_framework/blob/657fea688ffbc2ded888f882a514e443be4574fd/src/core/js/models/adaptModel.js#L259-L269

Which when started by a component are never finished before the page has started nor finished rendering. There are 4 layers of deferred completion checks with components, blocks, articles and pages, but only one defer holding the postRender function of the page. Meaning the postRender calls in sequence with the component completion checks.
https://github.com/adaptlearning/adapt_framework/blob/eed48b34b873e6e941c0258e68d35965ecc56b5a/src/core/js/views/adaptView.js#L50-L56

An entire page and its content are rendered before the whole set of descendent models are properly reset on revisit.
This causes issues if the reset state of a block should be derived after the page prerender and before the block is rendered and whilst a component is set to reset on revisit - such is the case in the new version of trickle where trickle will stop only if the block is incomplete, which it should be but isn't yet.

### Proposed solution
Have the router perform the reset on revisit and wait for the cascade to finish before rendering the page.

oliverfoster:
fixes #3014 

### Changed
* `Adapt.deferUntilCompletionChecked` is now an async +/ callback function
* Moved reset on revisit check from PageView preRender to router before render

### DIFF
--- a/src/core/js/adapt.js
+++ b/src/core/js/adapt.js
@@ -76,17 +76,17 @@ class AdaptSingleton extends LockingModel {
    * wait until there are no outstanding completion checks
    * @param {Function} callback Function to be called after all completion checks have been completed
    */
-  deferUntilCompletionChecked(callback) {
+  async deferUntilCompletionChecked(callback = () => {}) {
     if (this.get('_outstandingCompletionChecks') === 0) return callback();
-
-    const checkIfAnyChecksOutstanding = (model, outstandingChecks) => {
-      if (outstandingChecks !== 0) return;
-      this.off('change:_outstandingCompletionChecks', checkIfAnyChecksOutstanding);
-      callback();
-    };
-
-    this.on('change:_outstandingCompletionChecks', checkIfAnyChecksOutstanding);
-
+    return new Promise(resolve => {
+      const checkIfAnyChecksOutstanding = (model, outstandingChecks) => {
+        if (outstandingChecks !== 0) return;
+        this.off('change:_outstandingCompletionChecks', checkIfAnyChecksOutstanding);
+        callback();
+        resolve();
+      };
+      this.on('change:_outstandingCompletionChecks', checkIfAnyChecksOutstanding);
+    });
   }
 
   setupWait() {
--- a/src/core/js/adapt.js
+++ b/src/core/js/adapt.js
@@ -74,15 +74,17 @@ class AdaptSingleton extends LockingModel {
 
   /**
    * wait until there are no outstanding completion checks
-   * @param {Function} callback Function to be called after all completion checks have been completed
+   * @param {Function} [callback] Function to be called after all completion checks have been completed
    */
-  async deferUntilCompletionChecked(callback = () => {}) {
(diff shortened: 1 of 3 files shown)


cahirodoherty-learningpool (approved):
👀

src/core/js/adapt.js
```diff
@@ -74,19 +74,21 @@ class AdaptSingleton extends LockingModel {
...
-
-    const checkIfAnyChecksOutstanding = (model, outstandingChecks) => {
-      if (outstandingChecks !== 0) return;
-      this.off('change:_outstandingCompletionChecks', checkIfAnyChecksOutstanding);
-      callback();
-    };
-
-    this.on('change:_outstandingCompletionChecks', checkIfAnyChecksOutstanding);
-
+  async deferUntilCompletionChecked(callback) {
```
tomgreenfield, line 79: Could set default parameter to remove the need for existence checks:
```js
callback = () => {}
```

tomgreenfield (approved):
:eye:

src/core/js/adapt.js
oliverfoster: I did do that originally... But I'm not sure, feels a bit weird. Are you insistent or ambivalent?

src/core/js/router.js
```diff
@@ -193,6 +193,17 @@ class Router extends Backbone.Router {
       Adapt.log.deprecated(`Using event based menu view instantiation for '${Adapt.getViewName(model)}'`);
       return;
     }
+
+    if (!isMenu) {
+      // checkIfResetOnRevisit on descendant models before render
+      model.getAllDescendantModels().forEach(model => {
+        if (!model.checkIfResetOnRevisit) return;
+        model.checkIfResetOnRevisit();
+      });
```
moloko, line 202: Just a suggestion but you could shorten this to:
```suggestion
      _.invoke(model.getAllDescendantModels(), 'checkIfResetOnRevisit');
```
(I think `_.invoke` handles checking that the method specified in the 2nd argument actually exists)

oliverfoster: Well, sheeet. Ain't that pretty.

src/core/js/adapt.js
tomgreenfield: > Are you insistent or ambivalent?

My already-approved review would imply simple ambivalence.

oliverfoster: Duly noted.

tomgreenfield (approved):
:eye:

moloko (approved):
code 👓

PR #3018, opened 2021-01-14
issue/3017 Added config.json:build.strictMode

### LINKED ISSUE #3017: adapt_framework: Remove strict mode
### Subject of the issue
Some older code has missing `var` keywords and so variables are declared on the `window` namespace. 
When switching to the new version of Adapt with rollup, the output is in [`strict mode`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) changing the way these variables are declared and so breaking some code.

### Solution
1. Update the broken code.
2. Provide an option to disable strict mode enforcement.

oliverfoster:
fixes #3017 

### Added
* `config.json:build.strictMode` to turn on strict mode directives in javascript output
* Turned on in standard course

### DIFF
--- a/grunt/helpers.js
+++ b/grunt/helpers.js
@@ -234,7 +234,7 @@ module.exports = function(grunt) {
         menu: grunt.option('menu') || exports.defaults.menu,
         languages: languageFolders || exports.defaults.languages,
         scriptSafe: exports.defaults.scriptSafe,
-        strict: true
+        strictMode: false
       };
 
       if (buildConfig.jsonext) data.jsonext = buildConfig.jsonext;
@@ -246,7 +246,7 @@ module.exports = function(grunt) {
           return item.trim();
         });
       }
-      if (buildConfig.hasOwnProperty('strict')) data.strict = buildConfig.strict;
+      if (buildConfig.hasOwnProperty('strictMode')) data.strictMode = buildConfig.strictMode;
 
       return data;
     },
--- a/grunt/helpers.js
+++ b/grunt/helpers.js
@@ -213,7 +213,7 @@ module.exports = function(grunt) {
       var configPath = path.join(path.resolve(root, configDir), 'course', 'config.' + jsonext);
 
       try {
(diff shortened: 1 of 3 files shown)


moloko:
this will also need to be added to the schema for config.json I think? and also suggest adding it to the 'OOTB' course's config.json

I'm thinking that for FW courses strict mode should be on by default but maybe off by default for AAT ones? purely on the basis that an AAT instance is more likely to have a large collection of 3rd party plugins and it's unreasonable to require an another to amend this setting on every single course they have in the AAT just because one or two of the plugins are missing variable declarations

oliverfoster:
Yup, sounds like a plan. Need to make sure it fixes the issue before I do that stuff.

moloko:
@oliverfoster it's certainly fixing the problem of using a 'rollup' build of the FW with spoor v3.3.2 - not seeing any errors now if I set `strict: false`

oliverfoster:
There isn't any `build` property in the config schema atm. Is that possible / allowed? @taylortom  @tomgreenfield

oliverfoster:
@moloko it might just be easier to turn off the strict mode universally until some later breaking change date?

moloko:
sure, or just default it to off unless enabled via the setting in config.json - I think most people developing plugins do so in the FW first so that would give them the opportunity to get their code working to the `strict` requirement. I'd still prefer it to be enabled by default for the OOTB course (via config.json)

moloko:
could we change the name of the setting in config.json to `strictMode`?

oliverfoster:
So strict mode is off by default (AAT), enabled with `config.json:build.strictMode = true;` and enabled in the standard course.

oliverfoster:
@davetaz :
> @oliverfoster I can confirm that works and the course builds from the AAT with my broken plugin.

cahirodoherty-learningpool (approved):
👀

moloko (approved):

eleanor-heath (approved):

PR #2806, opened 2020-06-17
issue/2805 Introduced optional component tracking ids

oliverfoster:
#2805 
#### Added
* `$ grunt --trackingidtype='component'` option, set to `'block'` by default to retain the existing behaviour

#### Changed
* In grunt, made `_trackingIds` content `_type` agnostic

#### Fixed
* Tracking ids were put in the src/course after src/course > build copy so they only came out on a second run of grunt build/dev/diff. The tracking ids now go into src/course before copy.

#### Testing
Clone the course and switch to the correct branches
 ```js
git clone https://github.com/adaptlearning/adapt_framework 2805 
cd 2805 
git checkout issue/2805 
adapt devinstall && npm install
grunt dev
```

* With component tracking ids the course will not work properly unless you are using spoor pr [#197](https://github.com/adaptlearning/adapt-contrib-spoor/pull/197) (master branch). Use this pr to test trackingId placement only
* Assessments will not save and restore correctly without further amends, all components, blocks, articles, pages and menus should be fine

Delete tracking ids:
```bash
# remove block tracking ids
grunt tracking-remove
# remove component tracking ids
grunt tracking-remove --trackingidtype='component'
```

Add tracking ids:
```bash
# adds block tracking ids
grunt tracking-insert
# adds component tracking ids
grunt tracking-insert --trackingidtype='component'
```

Note: This does not yet mandate tracking ids on components, it only allows tracking ids on components for the meantime. Assessment and spoor need updates to allow for agnostic trackingId locations and attemptStates needs to go into the framework first. Assessment pr [#125](https://github.com/adaptlearning/adapt-contrib-assessment/pull/125) will need further reworking after merge in order to allow agnostic tracking ids.

The following extension need modifying to support component tracking ids:

- [x] [Spoor](https://github.com/adaptlearning/adapt-contrib-spoor/pull/197)
- [ ] Assessment

PR #3024, opened 2021-02-01
issue/2805 Added a mechanism for trackingId location agnosticism

### LINKED ISSUE #2805: adapt_framework: move trackingIds to components from blocks
### Subject of the issue
Blocks currently own the trackingIds for session state save and restore. This makes blocks mandatory as part of the course structure. It makes the component state save and restore more complicated than it needs to be.

By moving trackingIds to components we can remove the complexity and demote blocks from a necessary part of the course content to an optional part of the course content.

oliverfoster:
fixes #2805 

Provides a simple method of identifying models based upon the course's `_trackingId` properties, regardless of their location, be them on the `blocks` or `components`.

### Added
* `Adapt.findByTrackingPosition` similar to `Adapt.findById` but for `_trackingId` relative save/restore functions
* `AdaptModel.trackingPosition` which returns the model's trackingPosition relative to its nearest `_trackingId`

A tracking position is comprised of two numbers, the `_trackingId` and the relative offset from it. The relative offset is positive for a descendant or negative for an ancestor.

### Changed
* Comment in `Adapt.findById` was inaccurate

### Testing
```js
// Test all trackingPositions can be found for the course
Adapt.data.forEach((model) => {
  var id = model.get('_id');
  var trackingPosition = model.trackingPosition;
  console.log(trackingPosition, id);
  var found = Adapt.findByTrackingPosition(trackingPosition);
  if (id !== found.get('_id')) throw "Mismatch";
});
```

Try with `grunt dev` and `grunt dev --trackingidtype='component'`

https://github.com/adaptlearning/adapt_framework/pull/2806

### DIFF
--- a/src/core/js/adapt.js
+++ b/src/core/js/adapt.js
@@ -346,6 +346,15 @@ class AdaptSingleton extends LockingModel {
   }
 
   /**
+   * Returns the model represented by the trackingPosition.
+   * @param {Array<Number, Number>} trackingPosition Represents the relative location of a model to a _trackingId
+   * @returns {Backbone.Model}
+   */
+  findByTrackingPosition(trackingPosition) {
+    return this.data.findByTrackingPosition(trackingPosition);
+  }
+
+  /**
    * Relative strings describe the number and type of hops in the model hierarchy
    * @param {string} relativeString "@component +1" means to move one component forward from the current model
    * This function would return the following:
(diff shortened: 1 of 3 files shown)


moloko (approved):
code 👓

eleanor-heath (approved):

cahirodoherty-learningpool (approved):
👀

PR #3037, opened 2021-02-16
Update ESLint's ECMAScript syntax version

### LINKED ISSUE #3035: Syntax error when using ES2020 features
ESLint warns with an error in the code editor when using ES2020/ES11 features such as [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).

### Your environment
* Framework 5.10.1

### Steps to reproduce
Add code such as:
```js
if (foo?.bar) return;
```

### Expected behaviour
No syntax errors.

### Actual behaviour
```console
eslint      	error     	Parsing error: Unexpected token .
```

tomgreenfield:
Resolves #3035.

### DIFF
--- a/src/core/js/a11y.js
+++ b/src/core/js/a11y.js
@@ -99,7 +99,7 @@ class A11y extends Backbone.Controller {
   }
 
   _setupNoSelect() {
-    if (!this.config || !this.config._disableTextSelectOnClasses) {
+    if (!this.config?._disableTextSelectOnClasses) {
       return;
     }
     const classes = this.config._disableTextSelectOnClasses.split(' ');
@@ -135,7 +135,7 @@ class A11y extends Backbone.Controller {
 
   _onNavigationEnd(view) {
     // Prevent sub-menu items provoking behaviour
-    if ((view && view.model && view.model.get('_id') !== Adapt.location._currentId) || !this.isEnabled()) {
+    if ((view?.model?.get('_id') !== Adapt.location._currentId) || !this.isEnabled()) {
       return;
     }
     // Allow document to be read
@@ -148,7 +148,7 @@ class A11y extends Backbone.Controller {
   }
 
   isEnabled() {
-    return this.config && this.config._isEnabled;
+    return this.config?._isEnabled;
   }
 
   /**
@@ -380,7 +380,7 @@ class A11y extends Backbone.Controller {
 
     // check children by walking the tree
     let $found = this._findFirstForwardDescendant($element, iterator);
-    if ($found && $found.length) {
+    if ($found.length) {
       return $found;
     }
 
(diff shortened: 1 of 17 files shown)


src/core/js/adapt.js
```diff
@@ -30,7 +30,7 @@ class AdaptSingleton extends LockingModel {
    * @deprecated since v6.0.0 - please use `Adapt.store` instead
    */
   get componentStore() {
-    this.log && this.log.deprecated('Adapt.componentStore, please use Adapt.store instead');
+    this.log?.deprecated('Adapt.componentStore, please use Adapt.store instead');
```
moloko, line 33: I do wonder if the check for `this.log` is even necessary?

tomgreenfield: I also wonder. @oliverfoster?

oliverfoster: It's a quirk of the compilation order that makes `src/core/logging` load before that getter would ever be called. In practise it can be removed.

src/core/js/router.js
```diff
@@ -246,7 +246,7 @@ class Router extends Backbone.Router {
   setGlobalClasses() {
     const currentModel = Adapt.location._currentModel;
 
-    const htmlClasses = (currentModel && currentModel.get('_htmlClasses')) || '';
+    const htmlClasses = currentModel?.get('_htmlClasses') || '';
```
moloko, line 249: Possible argument for using nullish coalescing operator here - just on the off-chance someone decides they want to use "false" as an HTML class?

moloko: Ignore me, "false" wouldn't actually be considered 'falsy' would it?

src/core/js/adapt.js
moloko: as is, it's still causing an eslint error for some reason:
image

so yeah let's get rid of the check

moloko (approved):
lovely. such a useful new feature!

eleanor-heath (approved):

oliverfoster (approved):

src/core/js/a11y.js
```diff
@@ -380,7 +380,7 @@ class A11y extends Backbone.Controller {
 
     // check children by walking the tree
     let $found = this._findFirstForwardDescendant($element, iterator);
-    if ($found && $found.length) {
+    if ($found.length) {
```
cahirodoherty-learningpool, line 383: Won't these error out if $found is undefined?

cahirodoherty-learningpool:
Hi Tom, this cleanup is satisfying! Just a note on a few cases such as `$found && $found.length` turning to `$found.length` but I believe they should have the question mark `$found?.length`

src/core/js/a11y.js
oliverfoster: It always returns a jquery object.

tomgreenfield:
Morning @cahirodoherty-learningpool, that's entirely intentional; the jQuery object returned will never be falsy.

cahirodoherty-learningpool (approved):
👀

tomgreenfield:
> lovely. such a useful new feature!

Been available since version 5.8 in case you were wondering about including this syntax in your own plugins!

PR #3039, opened 2021-02-17
issue/3038 Click focus last parent

### LINKED ISSUE #3038: adapt_framework: focus on click focusing on last parent not first parent
### Subject of the issue
Because the [`$element.parents()`](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/a11y/browserFocus.js#L73) arrive in order from body to leaf, if more than one ancestor is tabbable then the ancestor closest to the body is chosen at [`$focusable[0]`](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/a11y/browserFocus.js#L79). Such that, with the example below, if `btn__inner` received a click event, `component__inner` would receive focus instead of `button`.

```html
<div class="component__inner" tabindex="-1">
  <button>
    <div class="btn__inner"></div>
  </button>
</div>

oliverfoster:
fixes #3038 

### Changed
* Made force click focus on the last parent, not the first

### DIFF
--- a/src/core/js/a11y/browserFocus.js
+++ b/src/core/js/a11y/browserFocus.js
@@ -76,7 +76,7 @@ export default class BrowserFocus extends Backbone.Controller {
       return;
     }
     // Force focus for screen reader enter / space press
-    $focusable[0].focus();
+    $focusable.last()[0].focus();
   }
 
 }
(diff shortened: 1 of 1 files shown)


moloko (approved):
code 👓

guywillis (approved):
👀

eleanor-heath (approved):

tomgreenfield (approved):

eleanor-heath (approved):

PR #2963, opened 2020-11-02
issue/2824-react Added react template support

oliverfoster:
#2824 #2944 

Added react support and handlebars-style react templates.

These plugins achieve react templating and integration in the same way we have been doing with handlebars:
* [babel-plugin-transform-react-templates](https://github.com/cgkineo/babel-plugin-transform-react-templates/) (custom made) register templates from filenames
* [html-react-parser](https://www.npmjs.com/package/html-react-parser) parse html into react nodes, like triples braces in handlebars
* [plugin-transform-react-jsx](https://www.npmjs.com/package/@babel/plugin-transform-react-jsx) jsx compiler
* [react](https://www.npmjs.com/package/react) client-side react node handler
* [react-dom](https://www.npmjs.com/package/react-dom) client-side react node renderer

#### Templates
Supports react 'templates' as near as I could get it to our current handlebars templates - separate, override-able etc. This change does not require a wholesale switch to React views but instead uses our current Backbone views and React as a templating engine. This means that some of the usual functionality of React views, such as the state object, aren't used to cause view re-render, instead the `this.changed()` function should be called to re-render a Backbone View with a React template. 

Here are a few reasons for using React as a templating engine only:
1. Our entire architecture uses Backbone events heavily - which React views don't support. 
2. Many of our bugs arise from jQuery DOM manipulation and using React to manage the DOM manipulation removes those bugs whilst introducing much more flexibility in the view <> state updates, this is as each manipulation is now automated and no longer needs to be coded by hand. 
3. Introducting jsx as a templating engine allows the upskilling requirement to be confined to templates only, which should be easier and less daunting for developers to confront. 
4. We will be able to convert to React at our leisure by not mandating a wholesale view layer rewrite. 
5. As React will be a part of the architecture we can better experiment with dropping Backbone as a large part of the architectural work is already complete.

#### Usage

A React version of the accordion lives here [pr](https://github.com/adaptlearning/adapt-contrib-accordion/pull/97) / [branch](https://github.com/adaptlearning/adapt-contrib-accordion/tree/issue/2824-jsx)

```js
// This is an example template file in **/templates/**/name.jsx
// It will get registered in Adapt by filename as handlebars templates currently do
// Templates will override from theme, menu, extension, component as usual
// There is no partials distinction with the react templates, they are all just templates which can be called from one another

export default function (model, view) {
  return <div></div> // React component
};
```

```js
// This is a template file in components/adapt-contrib-accordion/templates/accordion.jsx

// These functions help handlebars integration
[...]

PR #3050, opened 2021-03-01
Update view properties on model change

tomgreenfield:
Implements https://github.com/adaptlearning/adapt_framework/pull/2963#issuecomment-773214305.

### DIFF
--- a/src/core/js/views/adaptView.js
+++ b/src/core/js/views/adaptView.js
@@ -18,6 +18,7 @@ class AdaptView extends Backbone.View {
     });
     this.isReact = (this.constructor.template || '').includes('.jsx');
     if (this.isReact) {
+      this.classSet = new Set(_.result(this, 'className').trim().split(/\s+/));
       this.listenTo(this.model, 'all', this.changed);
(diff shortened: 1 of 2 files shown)


src/core/js/views/adaptView.js
```diff
@@ -18,6 +18,7 @@ class AdaptView extends Backbone.View {
     });
     this.isReact = (this.constructor.template || '').includes('.jsx');
     if (this.isReact) {
+      this.classSet = new Set(_.result(this, 'className').trim().split(/\s+/));
```
oliverfoster, line 21: Hi Tom,

Would you be able to make the instance property `this._classSet` private by prefixing it with an underscore please? 

It may also be prudent to change the names the variables throughout to give future readers a better indication as to their function?

`this.classSet`, `const classes` and `const difference` seems somewhat abstract. Could I perhaps suggest something along the lines of `this._renderedClasses`, `const currentClasses`, `const deleteClasses` or similar?

I look forward to hearing from you.

Warm regards,

Ollie.

oliverfoster (approved):

Issue #2961, opened 2020-10-29
Jaws cursor placement after submit in questions that _canShowFeedback: false & _canShowModelAnswer: true


moloko:
### Subject of the issue/enhancement/features
The placement of the Jaws virtual cursor goes a bit awry after submitting an incorrect (final attempt) answer to a question that no feedback but which has 'show model answer' enabled

### Your environment
* FW, v5.7.1 and master branch
* IE11 & Chrome with Jaws 2020

### Steps to reproduce
Use `adapt create course` to download the 'OOTB' course. Open articles.json and, in the assessment article (a-15), amend the assessment configuration so that `_questions._canShowModelAnswer: true`. Build and run the course. Launch Jaws, navigate to the assessment and answer a question incorrectly.

### Expected behaviour
I'm not sure TBH, it's hard to see where the cursor should go at this point. If you're using trickle with the button then the trickle button is the obvious solution, but what if you're using trickle with no button?

### Actual behaviour
In Chrome, moves back up to the answer options.

In IE11, focus remains on the Correct Answer button but Jaws reads it out as a Submit button

I did try adding in the changes from PR #2951 to see if that make any difference (it didn't).

moloko:
If we were to address #2942 and make the 'correct answer' functionality work for screen readers that would solve this problem as well as #2942 since focus could then be sent to the 'correct answer' button.

Issue #2942, opened 2020-10-14
'correct answer' button is not keyboard accessible


moloko:
Note: I'm specifically referring to navigating the course using the keyboard and NOT when using Jaws etc. Although it would be great if we could also look at getting that working, perhaps using an aria-live region to read out the correct answer(s)?

moloko:
tested using v5.6.1 and v5.7.0 but almost certainly still an issue in v5.7.1

oliverfoster:
As detailed previously and unless someone can think of an easier model: each question would need to be able to read its own correct answer as each question has a different format:

multi select mcq: you should have selected options 1, 3 and 5
single select mcq: you should have selected option 1
slider: you should have chosen between 1 and 4

etc

moloko:
> At that point it's almost sensible to make this feedback visible as it is much more useful than what we have at the moment.

I think if we're talking about reading out the labels (G/MCQ) - which I think you'd need to do as 'option 1' then it would all start to get a bit crowded on-screen. and unnecessary since you could see the options you should have selected more quickly than reading them out. Matching questions would be even worse. So I think just having them read by the screen reader is a better option.

And yes totally understand it's something each question type will need to implement.

as a short-term solution, is it possible to make the 'correct answer' button keyboard accessible but still keep it hidden from assistive tech?

oliverfoster:
> as a short-term solution, is it possible to make the 'correct answer' button keyboard accessible but still keep it hidden from assistive tech?

No, tab-index will allow focus regardless of the aria-hidden value. So it'd do nothing when focused / read anyway / mess up the screen reader.

PR #3051, opened 2021-03-03
allow correct answer/hide correct answer buttons to be keyboard/screen reader accessible

moloko:
part of #2942

~~Note: will also require updates to all the core question components, these will follow shortly~~ Now done - except for textInput

~~I'll also try and add a demo build so that anyone wanting to actually test this out can do so~~ done, see [this comment](#issuecomment-790584389)

### DIFF
--- a/src/core/js/models/itemsQuestionModel.js
+++ b/src/core/js/models/itemsQuestionModel.js
@@ -1,3 +1,4 @@
+import Adapt from 'core/js/adapt';
 import QuestionModel from 'core/js/models/questionModel';
 import ItemsComponentModel from 'core/js/models/itemsComponentModel';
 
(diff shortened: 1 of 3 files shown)


oliverfoster (approved):

moloko:
should also resolve #2961

moloko:
Sample course:
[correct-answer-a11y.zip](https://github.com/adaptlearning/adapt_framework/files/6083525/correct-answer-a11y.zip)

**Notes:** 

- doesn't yet include support for textInput. I'll sort this out at some point.
- everything should be backwardly-compatible i.e. the framework version should work with older versions of question components, though obviously the ARIA live region won't actually update if the learner clicks the correct answer button.

eleanor-heath (approved):

cahirodoherty-learningpool (approved):
👀

src/core/js/models/itemsQuestionModel.js
```diff
@@ -215,4 +216,51 @@ export default class ItemsQuestionModel extends BlendedItemsComponentQuestionMod
...
+    let correctAnswer;
+
+    if (this.isSingleSelect()) {
+      ariaAnswer = globals.ariaCorrectAnswer;
+      const correctOption = this.getChildren().findWhere({ _shouldBeSelected: true });
+      correctAnswer = correctOption.get('text');
+    } else {
+      ariaAnswer = globals.ariaCorrectAnswers;
+      const correctOptions = this.getChildren().where({ _shouldBeSelected: true });
+      correctAnswer = correctOptions.map(correctOption => correctOption.get('text')).join('<br>');
```
tomgreenfield, line 237: As discussed, could always pass through the correct answers as an array, giving flexibility to the plugin globals:
```hbs
The correct answers are {​{​#each correctAnswer}​}​{​{​{​this}​}​}​{​{​#unless @last}​}​<br>{{/unless}}{​{​/each}​}​
```
Don't know if it will ever be tweaked in practice.

tomgreenfield (approved):
:eye:

src/core/js/models/itemsQuestionModel.js
moloko: It's a nice suggestion but as you say, whether it will ever get used in practice is questionable.... can easily be changed in future if it does come up as a requirement.

PR #3060, opened 2021-03-10
amend labels of correct answer toggle button

### LINKED ISSUE #3059: Buttons labels: where possible the label should explain what action will be taken on click
For example, the button in question components that shows the correct answer is, by default, set to 'Correct answer' - it would be clearer if this were labelled 'Show correct answer'.

There may be other instances, I'll have a check

moloko:
fixes #3059


cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

PR #3063, opened 2021-03-10
issue/3061 Fixed router history

### LINKED ISSUE #3061: adapt_framework: router _canNavigate false state bug
### Subject of the issue
When routing is prevented using `router.set('_canNavigate', false)` on `router:navigate` after calling [`router.navigateBack`](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/router.js#L275), a history record always gets dropped and should be re-added to return the history to its proper state.

Currently the router calls [`router.navigateToCurrentRoute`](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/router.js#L113) which replaces the previous record rather than re-adding the current one.

When navigation is otherwise prevented, the history should always be moved back one to the proper state.

Essentially, if routing is prevented, the router needs to know if it has moved backwards to then go forward one or if it was going forward to then go backward one.

oliverfoster:
fixes #3061 

### Changed
* Router uses `history.back` and `history.forward` to correct url appropriately when routing is blocked going forwards and backwards

Test with [adapt-pageIncompletePrompt](https://github.com/cgkineo/adapt-pageIncompletePrompt) to see in action

### DIFF
--- a/src/core/js/router.js
+++ b/src/core/js/router.js
@@ -4,6 +4,7 @@ import CourseModel from 'core/js/models/courseModel';
 import ContentObjectModel from 'core/js/models/contentObjectModel';
 import MenuModel from 'core/js/models/menuModel';
 import 'core/js/startController';
+import Backbone from 'backbone';
 
 class Router extends Backbone.Router {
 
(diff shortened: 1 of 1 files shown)


moloko (approved):
Tested in FF

eleanor-heath (approved):

cahirodoherty-learningpool (approved):
👀

PR #3069, opened 2021-03-11
allow the tracking-remove grunt task to run even if spoor isn't installed

### LINKED ISSUE #3068: the tracking-remove task doesn't run if the spoor plugin isn't installed
I can't see any reason why not having spoor installed should prevent this from running. 

I might be converting a course from SCORM to non-SCORM and it would be convenient to be able to remove all tracking ids even though spoor has already been uninstalled.

moloko:
fixes #3068


oliverfoster (approved):

tomgreenfield (approved):
:eye:

cahirodoherty-learningpool (approved):
👀

Issue #3011, opened 2021-01-11
xAPI not restoring users language selection when resuming a course


kevindoherty30:
### Subject of the issue/enhancement/features
When viewing a multi-language course with xAPI enabled, the users selected language is not being restored when they revisit the course. Instead the user is prompted to select their language again.

### Your environment
* f/w version 5.7.0    xAPI version 0.8.3
* all browsers
* all devices + OS

### Steps to reproduce
Create a course
Add a page
Add a text component 
Duplicate the page a few times
Add the xAPI extension and ensure Track state is enabled
Add a few additional languages to the course 
Publish course and add it to platform
Open the course as a learner and select a language
Complete page 1
Exit the course
Re-open the course.

### Expected behaviour
On re-opening, the course should be in the language selected on the previous launch

### Actual behaviour
The language picker screen is presented for users to select their language again

Additional information:
The expected behaviour is currently how it works on a scorm course with spoor enabled.

moloko:
duplicate of https://github.com/adaptlearning/adapt-contrib-xapi/issues/88

cahirodoherty-learningpool:
Linked PR: https://github.com/adaptlearning/adapt_framework/pull/3086

PR #3086, opened 2021-03-18
Language set from param

cahirodoherty-learningpool:
Work arising from suggestion on https://github.com/adaptlearning/adapt-contrib-languagePicker/pull/46

### DIFF
--- a/src/core/js/models/configModel.js
+++ b/src/core/js/models/configModel.js
@@ -21,6 +21,7 @@ export default class ConfigModel extends LockingModel {
    */
   setValuesFromURLParams() {
     const paramMappings = {
+      dir: '_defaultDirection',
       lang: '_defaultLanguage'
     };
 
@@ -29,9 +30,9 @@ export default class ConfigModel extends LockingModel {
     for (const [key, value] of Object.entries(paramMappings)) {
       const passedVal = params.get(key);
 
-      if (passedVal) {
-        this.set(value, passedVal);
-      }
+      if (!passedVal) continue;
+
+      this.set(value, passedVal);
     }
   }
 
--- a/src/core/js/models/configModel.js
+++ b/src/core/js/models/configModel.js
@@ -16,6 +16,22 @@ export default class ConfigModel extends LockingModel {
     };
   }
 
+  /**
+   * Parse the incoming search queries for a 'lang' parameter value to return
+   */
+  getLanguageFromURL() {
+    const search = window.location.search.substring(1);
+    const queries = search.split('&');
+
+    queries.forEach((query) => {
+      const pair = query.split('=');
+
+      if (decodeURIComponent(pair[0]) !== 'lang') return;
+
+      return decodeURIComponent(pair[1]);
+    });
+  }
+
   initialize(attrs, options) {
     this.url = options.url;
     // Fetch data & if successful trigger event to enable plugins to stop course files loading
@@ -23,6 +39,9 @@ export default class ConfigModel extends LockingModel {
     // 'configModel:loadCourseData' event starts the core content collections and models being fetched
     this.fetch({
       success: () => {
+        const language = this.getLanguageFromURL();
+        if (language) this.set('_defaultLanguage', language);
+
         Adapt.trigger('offlineStorage:prepare');
         Adapt.wait.queue(() => {
           Adapt.trigger('configModel:dataLoaded');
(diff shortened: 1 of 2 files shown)


src/core/js/models/configModel.js
```diff
@@ -16,13 +16,36 @@ export default class ConfigModel extends LockingModel {
...
     // Fetch data & if successful trigger event to enable plugins to stop course files loading
     // Then check if course files can load
     // 'configModel:loadCourseData' event starts the core content collections and models being fetched
     this.fetch({
       success: () => {
+        const language = this.getLanguageFromURL();
+
+        if (language) {
+          this.set('_defaultLanguage', language);
+          this.set('_activeLanguage', language);
```
moloko, line 46: are you sure this line is required? thought it would handle this automatically elsewhere

src/core/js/models/configModel.js
```diff
@@ -16,13 +16,36 @@ export default class ConfigModel extends LockingModel {
     };
   }
 
+  /**
+   * Parse the incoming search queries for a 'lang' parameter value to return
+   */
+  getLanguageFromURL() {
```
moloko, line 22: I like [your idea of making this more generic](https://github.com/adaptlearning/adapt-contrib-languagePicker/pull/46#discussion_r596862681) - it'd just be a question of where best to put that code - this doesn't feel like quite the right place? @oliverfoster any suggestions?

moloko (commented):
are you sure about removing the `return` from `onLoadCourseData`? Pretty sure changing `_activeLanguage` automatically triggers a call to `loadCourseData`.

src/core/js/models/configModel.js
moloko: We might also consider adding a [polyfill for URLSearchParams](https://github.com/ungap/url-search-params)

eleanor-heath (approved):

moloko:
looks great - just one thought: might want to allow setting of `_defaultDirection` too for RTL languages? again, a nice short param would be good e.g. `"dir"` - so you'd end up with something like `?lang=ar&dir=rtl`

I'll do a PR for the `URLSearchParams` polyfill in a sec

src/core/js/models/configModel.js
```diff
@@ -16,13 +16,34 @@ export default class ConfigModel extends LockingModel {
...
+
+    const params = new URLSearchParams(window.location.search);
+
+    for (const [key, value] of Object.entries(paramMappings)) {
+      const passedVal = params.get(key);
+
+      if (passedVal) {
+        this.set(value, passedVal);
+      }
+    }
```
moloko, line 35:
```suggestion
    const paramMappings = {
      dir: '_defaultDirection',
      lang: '_defaultLanguage'
    };

    const params = new URLSearchParams(window.location.search);

    for (const [key, value] of Object.entries(paramMappings)) {
      const passedVal = params.get(key);
      if (!passedVal) continue;

      this.set(value, passedVal);
    }
```

moloko, line 35:
```suggestion
    const paramMappings = {
      dir: '_defaultDirection',
      lang: '_defaultLanguage'
    };

    const params = new URLSearchParams(window.location.search);

    for (const [key, value] of Object.entries(paramMappings)) {
      const passedVal = params.get(key);
      if (!passedVal) continue;
      
      this.set(value, passedVal);
    }
```

moloko: not quite sure why every single line is being flagged with a change...! but i'm sure you get the general idea anyway

moloko (approved):
the comment for `setValuesFromURLParams` needs updating a little now but otherwise this looks good

cahirodoherty-learningpool:
This will also resolve https://github.com/adaptlearning/adapt_framework/issues/3011

src/core/js/models/configModel.js
```diff
@@ -16,13 +16,35 @@ export default class ConfigModel extends LockingModel {
...
+    const params = new URLSearchParams(window.location.search);
+
+    for (const [key, value] of Object.entries(paramMappings)) {
+      const passedVal = params.get(key);
+
+      if (!passedVal) continue;
+
+      this.set(value, passedVal);
+    }
+  }
```
oliverfoster, line 37: This works just as well:
```suggestion
    Object.entries(paramMappings).forEach(([key, value]) => {
      const passedVal = params.get(key);
      if (!passedVal) return;
      this.set(value, passedVal);
    });
  }
```

oliverfoster (approved):

eleanor-heath (approved):

moloko (approved):

PR #3100, opened 2021-03-23
search element's parents for aria-disabled

### LINKED ISSUE #3097: adapt_framework: doesn't always prevent buttons with aria-disabled from being clicked
If the button contains an inner element - for example a `<div>` wrapping the text label - clicking that element will trigger the button's click event anyway.

Need to amend [this function](https://github.com/adaptlearning/adapt_framework/blob/833b4ef03760ac78fb5d53eb8b0e009380405bfd/src/core/js/a11y/browserFocus.js#L63-L68) to get it to check parents as well.

Additionally, that code doesn't seem to be initialising during the initial display of the language picker. Might be easiest just to change those to use the `disabled` attribute since I don't think it's really the end of the world if screenreader users are unaware of the existence of languages that aren't yet available.

moloko:
fixes #3097

### DIFF
--- a/src/core/js/a11y/browserFocus.js
+++ b/src/core/js/a11y/browserFocus.js
@@ -62,7 +62,9 @@ export default class BrowserFocus extends Backbone.Controller {
    */
   _onClick(event) {
     const $element = $(event.target);
-    if ($element.is('[aria-disabled=true]')) {
+    // search elemenent and parents for aria-disabled - see https://github.com/adaptlearning/adapt_framework/issues/3097
+    const isAriaDisabled = $element.closest('[aria-disabled=true]') > 0;
+    if (isAriaDisabled) {
       event.preventDefault();
       event.stopImmediatePropagation();
     }
--- a/src/core/js/a11y/browserFocus.js
+++ b/src/core/js/a11y/browserFocus.js
@@ -62,8 +62,8 @@ export default class BrowserFocus extends Backbone.Controller {
    */
   _onClick(event) {
     const $element = $(event.target);
-    // search elemenent and parents for aria-disabled - see https://github.com/adaptlearning/adapt_framework/issues/3097
-    const isAriaDisabled = $element.closest('[aria-disabled=true]').length > 0;
+    // search element and parents for aria-disabled - see https://github.com/adaptlearning/adapt_framework/issues/3097
+    const isAriaDisabled = $element.closest('[aria-disabled=true]').length === 1;
     if (isAriaDisabled) {
       event.preventDefault();
       event.stopImmediatePropagation();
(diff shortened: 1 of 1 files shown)


src/core/js/a11y/browserFocus.js
```diff
@@ -62,7 +62,9 @@ export default class BrowserFocus extends Backbone.Controller {
    */
   _onClick(event) {
     const $element = $(event.target);
-    if ($element.is('[aria-disabled=true]')) {
+    // search elemenent and parents for aria-disabled - see https://github.com/adaptlearning/adapt_framework/issues/3097
+    const isAriaDisabled = $element.closest('[aria-disabled=true]') > 0;
```
tomgreenfield, line 66: This will never resolve to a truthy.

`$element.closest('[aria-disabled=true]') > 0` = `false`
`$element.closest('[aria-disabled=true]').length` = `0` | `1`
`$element.closest('[aria-disabled=true]').length !== 0` = `true` | `false`
`Boolean($element.closest('[aria-disabled=true]').length)` = `true` | `false`

tomgreenfield (changes requested):

src/core/js/a11y/browserFocus.js
oliverfoster: `$element.closest('[aria-disabled=true]')` is a jQuery object, needs `.length` on the end.

moloko: oops - thanks both!

src/core/js/a11y/browserFocus.js
```diff
@@ -62,7 +62,9 @@ export default class BrowserFocus extends Backbone.Controller {
    */
   _onClick(event) {
     const $element = $(event.target);
-    if ($element.is('[aria-disabled=true]')) {
+    // search elemenent and parents for aria-disabled - see https://github.com/adaptlearning/adapt_framework/issues/3097
```
tomgreenfield, line 65: Typo.

oliverfoster (approved):

src/core/js/a11y/browserFocus.js
```diff
@@ -62,7 +62,9 @@ export default class BrowserFocus extends Backbone.Controller {
...
+    const isAriaDisabled = $element.closest('[aria-disabled=true]').length > 0;
```
tomgreenfield, line 66: Just an FYI re: `> 0`, in practice `.length` will either be `0` or `1` since `$.closest` only ever returns one element for each source element.

tomgreenfield (approved):

cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

PR #3112, opened 2021-04-06
don't force start page to redisplay

### LINKED ISSUE #3111: multilanguage course w/start page: switching language forces start page to be (re)displayed
In a multilanguage course where:
* there's a start page with the `_force` setting set to `false` and `_skipIfComplete: true`
* `_restoreStateOnLanguageChange: true` in the language picker configuration

changing language (after having completed the start page) causes the learner to be returned to the start page before being taken back to the location there were at when they changed language.

The learner should be taken straight back to the location they were at when they changed language.

moloko:
following a change in language (fixes #3111)

fix typo in comment

### DIFF
--- a/src/core/js/data.js
+++ b/src/core/js/data.js
@@ -100,7 +100,7 @@ class Data extends AdaptCollection {
   getJSON(path) {
     return new Promise((resolve, reject) => {
       $.getJSON(path, data => {
-        // Add path to data incase it's necessary later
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

tomgreenfield (approved):
:eye:

eleanor-heath (approved):

PR #3114, opened 2021-04-06
added URLSearchParams polyfill

### LINKED ISSUE #3113: add URLSearchParams polyfill
As IE11 doesn't have support for [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) and the [existing polyfill library](https://github.com/adaptlearning/adapt_framework/blob/63431509e471d2f414235934c9bb2826baac8f82/src/core/libraries/polyfill.min.js) doesn't seem to include it...

moloko:
from https://github.com/jerrybendy/url-search-params-polyfill

resolves #3113

### DIFF
--- a/src/core/libraries/urlsearchparams.js
+++ b/src/core/libraries/urlsearchparams.js
@@ -1,0 +1,355 @@
+/**
+ *
+ *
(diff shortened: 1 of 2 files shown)


moloko:
closing in favour of #3115

PR #3118, opened 2021-04-09
Drawer focus on first element in drawer

eleanor-heath:
Issue: https://github.com/adaptlearning/adapt_framework/issues/3119

### DIFF
--- a/src/core/js/views/drawerView.js
+++ b/src/core/js/views/drawerView.js
@@ -114,7 +114,8 @@ class DrawerView extends Backbone.View {
   }
 
   showDrawer(emptyDrawer) {
-    this.$el.removeClass('u-display-none').removeAttr('aria-hidden');
+    const $drawer = this.$el;
+    $drawer.removeClass('u-display-none').removeAttr('aria-hidden');
     // Only trigger popup:opened if drawer is visible, pass popup manager drawer element
     if (!this._isVisible) {
       Adapt.a11y.popupOpened(this.$el);
@@ -152,7 +153,7 @@ class DrawerView extends Backbone.View {
       Adapt.trigger('drawer:opened');
 
       // focus on first tabbable element in drawer
-      Adapt.a11y.focusFirst(this.$el, { defer: true });
+      Adapt.a11y.focusFirst($drawer, { defer: true });
     };
 
     // delay drawer animation until after background fadeout animation is complete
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

moloko:
are you absolutely sure something else is not going on here? unless I'm missing something, all you're doing is storing a reference to `this.$el` in a local variable. It's really hard to see how that's going to have any impact...

eleanor-heath:
@moloko I will double check now

eleanor-heath:
this.$el evaluates to undefined for me on my phone, but if I create a reference to this.$el outside the complete function scope I can access the element.

this evaluates to the following as shown in the screenshot when im inside the complete function:

Screenshot 2021-04-09 at 11 46 32

Screenshot 2021-04-09 at 11 46 46


But if I use the reference to this.$el that I created outside the scope I can retrieve the element see below screenshot

Screenshot 2021-04-09 at 11 47 03

moloko:
Interestingly I get exactly same result on FF. The `complete` function just doesn't appear to be picking up the scope of `this`. Either it's one of those weird ones which @oliverfoster will probably be able to explain, or possibly it's something the ES6 transpiler is screwing up.

If you change it back to:
```js
const complete = function () {
  this.addShadowEvent();
  Adapt.trigger('drawer:opened');

  // focus on first tabbable element in drawer
  Adapt.a11y.focusFirst(this.$el, { defer: true });
}.bind(this);
```
it works fine.

moloko:
as @tomgreenfield has pointed out, it could just be a quirk of the debugger... if I breakpoint the `complete` function, `this` is `undefined` - but if I include a `console.log`...

image

More importantly, if I check the value of `document.activeElement` after `complete` has run, the `.drawer` div does have focus...

moloko:
as @tomgreenfield has pointed out, it could just be a quirk of the debugger... if I breakpoint the `complete` function, `this` is `undefined` - but if I include a `console.log`...

image

More importantly, if I check the value of `document.activeElement` after `complete` has run, the `.drawer` div does have focus...

tomgreenfield:
Can't replicate this one on my flatmate's phone – the `div.drawer` gets focus.

eleanor-heath:
@moloko @tomgreenfield grand, am happy enough to close this one if it is just a quirk of the debugger.

PR #3120, opened 2021-04-12
initialise device size & orientation checking earlier

### LINKED ISSUE #3116: adapt_framework: incorrect device classes with adapt-contrib-languagePicker
If using https://github.com/adaptlearning/adapt-contrib-languagePicker, it is possible for the size and orientation classes to be incorrect on the html tag once the course has fully loaded/initialized.

### Your environment
* adapt_framework 5.12.1, but not specific to this version

### Steps to reproduce
* Ensure `adapt-contrib-languagePicker` is installed and enabled, with `"_showOnCourseLoad": true`
* Before selecting a language, resize the browser enough to enter a different breakpoint, i.e. large to small, portrait to landscape
* Pick a language and then inspect the html classes which will still be set to the original values when the course loaded 

### Expected behaviour
https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/device.js doesn't listen to `resize orientationchange` events until `Adapt#app:dataReady`, which occurs after the languagePicker, so any changes made before a language selection are redundant and not reflected in the html classes once the data is ready.

### Proposed change
The most obvious solution would be to change https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/device.js#L19 to listen to `configModel:dataLoaded` or `offlineStorage:ready`.

moloko:
fixes #3116

### DIFF
--- a/src/core/js/device.js
+++ b/src/core/js/device.js
@@ -16,7 +16,7 @@ class Device extends Backbone.Controller {
     this.renderingEngine = this.getRenderingEngine();
     this.onWindowResize = _.debounce(this.onWindowResize.bind(this), 100);
     this.listenTo(Adapt, {
-      'app:dataReady': this.onDataReady
+      'configModel:dataLoaded': this.onConfigDataLoaded
     });
     const browser = this.browser.toLowerCase();
     // Convert 'msie' and 'internet explorer' to 'ie'.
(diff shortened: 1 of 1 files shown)


oliverfoster (approved):

tomgreenfield (approved):
:eye:

cahirodoherty-learningpool (approved):
👀

Issue #3163, opened 2021-05-27
adapt_framework: scoring


oliverfoster:
### Subject of the issue
[Original forum comment](https://community.adaptlearning.org/mod/forum/discuss.php?d=4036#p14820)

There is no option specific scoring in Adapt at present. 

Currently, with an assessment, questions are scored 0 / `_questionWeight` according to their correctness, not an option specific score. Effectively, `_questionWeight` behaves as `_questionMaxScore` or `_questionCorrectScore` (for weird legacy reasons) so that we can calculate a percentage score from a collection of questions.

* [mcq/gmcq model](https://github.com/adaptlearning/adapt_framework/blob/deea638ca36553b03ba4c13fc719d9854364341d/src/core/js/models/itemsQuestionModel.js#L108-L113)
* [matching model](https://github.com/adaptlearning/adapt-contrib-matching/blob/949a7e2ae0765a9c69871d3d37b26d2f89e87457/js/matchingModel.js#L124-L138)
* [assessment scoring](https://github.com/adaptlearning/adapt-contrib-assessment/blob/8b37309e2dff5d9129e306b0d97dc7ef6fcb58e6/js/adapt-assessmentArticleModel.js#L440-L463)

You can see that the assessment ignores the `_score` value on the questions in favour of `_questionWeight` as questions have no `_maxScore` which is essential for proportionality (out of 100%).

@danielghost Suggests separating the `correctness` from the `score` calculation and handing that decision to the designer.

Proposal:
* QuestionModel
  * Add `score`, defaulted to `0` for incorrect/partly correct, `maxScore` for correct
  * Add `maxScore`, defaulted to `_questionWeight`
  * Add `minScore`, defaulted to `0`
* ItemsQuestionModel
  * Add `_hasItemScoring`, defaulted to `false`
  * With `_hasItemScoring === true`
    * Add `_items._score`, defaulted to `0`
    * Add `score` as the sum of the `_items._isActive`,  `_items._score`
    * Add `maxScore` as the sum of the `_selectable` highest positive `_items._score` items or 0
    * Add `minScore` as the sum of the `_selectable` lowest negative `_items._score` items or 0
  * Keep correctness as incorrect/partiallyCorrect/correct based on `_selectable`, `_items._shouldBeSelected`
  * Keep `_score` as a ratio of `score` to `maxScore` in a range of 0-1
* Assessment  
  * Change to scoring by `score` instead of [`_questionWeight` and `_isCorrect`] 
[...]

PR #3172, opened 2021-06-13
issue/3163 New scoring api

### LINKED ISSUE #893: Different answers weight rather than just a single question weight
I'm to build an assessment which requires a question component I'm not sure how to build with Adapt.
It's a 4-options MCQ of single answer. All 4 options are selectable, but different answer option has different weight point. For example, Option A - 4 points, Option B - 3 points, Option c - 2 points and Option d - 1 point.
Currently, the adapt-contrib-mcq only allow setting up the whole question weight rather than the answer weight.

+1 johnniezen

Moved from https://adaptlearning.atlassian.net/browse/ABU-1043

### LINKED ISSUE #1974: assessment: uses _questionWeight rather than _score to calculate each question's score
Given that questions have a `setScore` function that's there specifically to allow them to calculate their score based on `_questionWeight`, it seems a little bit odd that the `_score` is then completely ignored by the assessment's scoring calculations.

In the example I'm working to I have a question component that presents the user with multiple questions - therefore I'd like the assessment to take into account that the user got, say, 3 out of a possible 5 correct - but at the moment I can't see that this is possible.

What is clear is that if this is changed to use `_score` instead of `_questionWeight` it will be necessary to make some adjustments to matching and textInput since they currently calculate their score based on the number of correct answers given - so to maintain existing behaviour for these it will be necessary to do something like add a setting (defaulted to `true`) that says that a score should only be given for a completely correct answer.

oliverfoster:
#3163 
fixes #1974
fixes #893

### Added
* `QuestionModel.score` returns `0` for incorrect/partly correct or `maxScore` for correct
* `QuestionModel.maxScore` returns `_questionWeight`
* `QuestionModel.minScore` returns `0`
* `QuestionModel._hasItemScoring` to control item scoring
* `QuestionModel.updateRawScore()` adds `_rawScore = get score`, `_maxScore = get maxScore`, `_minScore = get minScore` on `init` and `markQuestion` to give access to the values in templates
* `ItemsQuestionModel.score` when `_hasItemScoring === true` return sum of `_items._isActive && _items._score`
* `ItemsQuestionModel.maxScore` when `_hasItemScoring=== true` return the sum of `_selectable` highest positive `_items._score`
* `ItemsQuestionModel.minScore` when `_hasItemScoring=== true` return the sum of `_selectable` lowest negative `_items._score`
* `ItemsQuestionModel._score` defaulted to 0

Test with https://github.com/adaptlearning/adapt-contrib-assessment/pull/132

### DIFF
--- a/src/core/js/models/itemsQuestionModel.js
+++ b/src/core/js/models/itemsQuestionModel.js
@@ -113,25 +113,25 @@ export default class ItemsQuestionModel extends BlendedItemsComponentQuestionMod
   }
 
   get score() {
-    if (!this.get('_isItemScoring')) return super.score;
+    if (!this.get('_hasItemScoring')) return super.score;
     const children = this.getChildren()?.toArray() || [];
     return children.reduce((score, child) => (score += child.get('_isActive') ? child.get('_score') || 0 : 0), 0);
   }
 
   get maxScore() {
-    if (!this.get('_isItemScoring')) return super.maxScore;
+    if (!this.get('_hasItemScoring')) return super.maxScore;
     const children = this.getChildren()?.toArray() || [];
     const scores = children.map(child => child.get('_score') || 0);
     scores.sort();
-    return scores.reverse().slice(0, this.get('_seletable')).filter(score => score > 0).reduce((maxScore, score) => (maxScore += score), 0);
+    return scores.reverse().slice(0, this.get('_selectable')).filter(score => score > 0).reduce((maxScore, score) => (maxScore += score), 0);
   }
 
   get minScore() {
-    if (!this.get('_isItemScoring')) return super.minScore;
+    if (!this.get('_hasItemScoring')) return super.minScore;
     const children = this.getChildren()?.toArray() || [];
     const scores = children.map(child => child.get('_score') || 0);
     scores.sort();
-    return scores.slice(0, this.get('_seletable')).filter(score => score < 0).reduce((minScore, score) => (minScore += score), 0);
+    return scores.slice(0, this.get('_selectable')).filter(score => score < 0).reduce((minScore, score) => (minScore += score), 0);
   }
 
   setupFeedback() {
--- a/src/core/js/models/itemsQuestionModel.js
+++ b/src/core/js/models/itemsQuestionModel.js
@@ -112,6 +112,28 @@ export default class ItemsQuestionModel extends BlendedItemsComponentQuestionMod
     this.set('_score', score);
   }
 
+  get score() {
+    if (!this.get('_isItemScoring')) return super.score;
+    const children = this.getChildren()?.toArray() || [];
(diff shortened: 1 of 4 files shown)


cahirodoherty-learningpool (approved):
👀

eleanor-heath (approved):

eleanor-heath (approved):

src/core/js/models/itemsQuestionModel.js
```diff
@@ -112,6 +112,28 @@ export default class ItemsQuestionModel extends BlendedItemsComponentQuestionMod
...
+    const children = this.getChildren()?.toArray() || [];
+    return children.reduce((score, child) => (score += child.get('_isActive') ? child.get('_score') || 0 : 0), 0);
+  }
+
+  get maxScore() {
+    if (!this.get('_isItemScoring')) return super.maxScore;
+    const children = this.getChildren()?.toArray() || [];
+    const scores = children.map(child => child.get('_score') || 0);
+    scores.sort();
+    return scores.reverse().slice(0, this.get('_seletable')).filter(score => score > 0).reduce((maxScore, score) => (maxScore += score), 0);
```
tomgreenfield, line 126: Typo here and on line 134 (`seletable`).

src/core/js/models/questionModel.js
```diff
@@ -20,6 +20,7 @@ class QuestionModel extends ComponentModel {
       _canSubmit: true,
       _isSubmitted: false,
       _questionWeight: Adapt.config.get('_questionWeight'),
+      _isItemScoring: false,
```
tomgreenfield, line 23: Suggest `_hasItemScoring`?

tomgreenfield (changes requested):

chrisgillison:
Using Framework 5.13.0 as a starting point, I've changed the Framework, Assessment & MCQ files to test this. Each correct question item in the assessment has had a "_score" attribute added under "_shouldBeSelected", with values of 4 for correct & 0 for incorrect, but I'm still getting awarded 1 point per question, and maxScore value is the same as the number of questions posed. Am I missing something?

oliverfoster:
Probably:
`
"_hasItemScoring": true
`

oliverfoster:
image

chrisgillison:
> 
> 
> image

Oh God! How incredibly embarrassing! Yep - works like a dream for me now!

chrisgillison:
Personally, I think this is a brilliant enhancement to the scoring API Ollie. Something to shout about. Sometimes there is no set of correct/incorrect answers, but rather degrees of correctness. I can think of a number instances where I could have used this in the past.

So, is adapt-contrib-matching next on your list?!! I tried to update it myself using your mods to itemsQuestionModel as a guide, but of course each item has multiple options. My efforts led to a hopelessly broken component ;D

oliverfoster:
Yup. Matching next. 

Glad you like it. It's been frustrating me for ages the lack of item specific scoring.

PR #3175, opened 2021-06-16
issue/1568 Allow parse of @menu @page @article without offsets

### LINKED ISSUE #1568: adapt-contrib-trickle - finish button should navigate back when page complete
Very minor enhancement but when you get to the end of a trickle-enabled page and the finish button pops up clicking it doesn't do anything. I suggest the default behaviour should be to navigate back.

Am happy to provide the code for this if the consensus is that this is a good idea - in which case please assign this to me. Otherwise just close the issue and I won't take offence :)

oliverfoster:
fixes #1568 

### Changed
* Allowed `Adapt.parseRelativeString` to return `{ type: 'menu', offset: 0 }` when receiving type groups without an offset like `@menu`, `@page`, `@article` rather than `@menu+0` which is how to find the current ancestor menu

### DIFF
--- a/src/core/js/adapt.js
+++ b/src/core/js/adapt.js
@@ -365,7 +365,8 @@ class AdaptSingleton extends LockingModel {
    * Trickle uses this function to determine where it should scrollTo after it unlocks
    */
   parseRelativeString(relativeString) {
-    const splitIndex = relativeString.search(/[ +\-\d]{1}/);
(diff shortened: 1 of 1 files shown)


tomgreenfield (approved):
:eye:

eleanor-heath (approved):

guywillis (approved):
👀

Issue #3193, opened 2021-07-06
adapt-contrib-vanilla: font weight not defined for <strong> and <b>


kirsty-hames:
Should we define a `font-weight` variable for `<strong>` and `<b> `tags in the theme?

I'd suggest the following for __font-mixins.less_:
```
b,
strong {
  font-weight: @font-weight-bold;
}

```

In _core/less/_browser/normalize.less_ we have the following:
```
b,
strong {
  font-weight: bolder;
}
```

This often causes inconsistencies in weights where we have specified a set weight for other bold items.

oliverfoster:
Yes, I think that's sensible.

guywillis:
Agree with the suggestion though do you think it should be defined in base.less in core like the below?

```
b,
strong {
  font-weight: bold;
}
```

PR #3201, opened 2021-07-08
Set bold font weight for b and strong tags

kirsty-hames:
It's expected that these tags are bold.

The current font-weight: bolder; inherited from normalize.less only increase the weight to the next weight value which can be less than the standard bold (700).
https://www.quirksmode.org/css/text/fontweight.html

resolves https://github.com/adaptlearning/adapt_framework/issues/3193

### DIFF
--- a/src/core/less/_defaults/base.less
+++ b/src/core/less/_defaults/base.less
@@ -74,3 +74,10 @@ zw {
 nb {
   white-space: nowrap;
 }
+
+// Set bold font weight for b and strong tags
+// --------------------------------------------------
+b,
+strong {
+  font-weight: bold;
+}


guywillis (approved):
👁️

cahirodoherty-learningpool (approved):
👀

guywillis (commented):
Upon reflection, I feel the normalize approach may be the correct approach for core and the override definition should remain in the vanilla pr.

If using a light font (300) for a body font, the normalize bolder option would increase the boldness of the font from light to regular (400) and this would be expected within CSS. I believe it's our expectation that using <strong> and <b> would change the boldness of the font to bold (700) which is where the issue resides.

eleanor-heath (approved):

kirsty-hames:
> Upon reflection, I feel the normalize approach may be the correct approach for core and the override definition should remain in the vanilla pr.
> 
> If using a light font (300) for a body font, the normalize bolder option would increase the boldness of the font from light to regular (400) and this would be expected within CSS. I believe it's our expectation that using `<strong>` and `<b>` would change the boldness of the font to bold (700) which is where the issue resides.

I agree with the above - I'm going to close the PR and keep the override in the theme only

PR #3343, opened 2022-11-18
New: Added grunt `scripts:adaptpostcopy` command

### LINKED ISSUE #3342: Add grunt `scripts:adaptpostcopy` command
The current `scripts:adaptpostbuild` command executes after all files have been copied and modified, meaning they won't benefit from commands such as `replace`. To allow plugins easier manipulation of build files, a `scripts:adaptpostcopy` command can be used to allow specific files to be included based on configuration settings, rather than adding all required files and then removing those not required post build.

This will help with https://github.com/adaptlearning/adapt-contrib-spoor/issues/249.

danielghost:
Fixes #3342

### New
* Added an additional script for plugins to hook into. Fires following the files being copied to the build folder, but before other helper commands such as `replace`.

### DIFF
--- a/grunt/tasks/build.js
+++ b/grunt/tasks/build.js
@@ -9,6 +9,7 @@ module.exports = function(grunt) {
     'build-config',
     'tracking-insert',
     'copy',
+    'scripts:adaptpostcopy',
     'schema-defaults',
     'language-data-manifests',
     'handlebars',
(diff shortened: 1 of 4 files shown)


cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):

oliverfoster (approved):

PR #3364, opened 2023-02-11
New: Cache management for subsequent compilation improvements (fixes #3363)

### LINKED ISSUE #3363: Potential performance compilation improvement
### Subject of the issue
Javascript compilation caches are currently deleted with a `grunt build` but retained with `grunt dev`. `grunt dev` therefore presents much faster subsequent recompilation times, sometimes in the realms of 3seconds for a subsequent compiled compared to 9seconds for a first pass/no cache compilation.

Caches are currently kept in the `build/` output folder so as to manage their appropriate erasure by project and not consume unnecessary disk space.

The key to improving subsequent compilation would be to retain the compilation caches for each project, aside from the `build/` folder, but doing so in a managed way to prevent unnecessary disk usage.

oliverfoster:
fixes #3363 

### New
* Allow adapt_framework to store javascript compilation cache files in the operating system temp folder, cleaning up files, once a day, if they are older than a week. Reduces subsequent compilation times sustantially. Initial compilation remains unchanged. Files are cached in `$TMPDIR/adapt_framework/`.

### DIFF
--- a/grunt/helpers/CacheManager.js
+++ b/grunt/helpers/CacheManager.js
@@ -28,9 +28,8 @@ module.exports = class CacheManager {
     return tempPath;
   }
 
-  get cachePath() {
-    const projectPath = process.cwd();
-    const projectHash = CacheManager.hash(projectPath);
+  cachePath(basePath, outputFilePath = process.cwd()) {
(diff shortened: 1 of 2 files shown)


eleanor-heath (approved):

cahirodoherty-learningpool (approved):
👀

Issue #2526, opened 2019-09-25
Future: transition from bower to npm / yarn


gorhgorh:
### transition from bower to npm / yarn
notes following a discussion on the gitter channel, where we discussed advantages / need to choose between npm and yarn.
here are some questions that @oliverfoster raised 

- is yarn going to be the defacto standard npm client at some point?
- are the features going to work their way back into npm?
- does yarn have it's own module listing infrastructure or does it rely on the npm catalogue?
- would i need to go to two places to understand adapt's plugin management architecture (the npm and yarn websites) and is that tooling ambiguity worth the extra features?
- if we use npm does that stop us using yarn in the future and vice-versa?

here is my own take on these, feel free to discuss
#### tldr
I don't think adapt should get bound to npm or yarn only, currently yarn is 100% compatible with npm so stick to what npm offer, as it allow 100% of the npm OR yarn only users to use the framework.
One maybe interesting thing in yarn would be [workspaces](https://yarnpkg.com/en/docs/workspaces)

As for namespacing, the adapt org is still up as I'm writing this issue.
so you can have `@adapt/mcq` like package names.

more detailed answers bellow

#### is yarn going to be the defacto standard npm client at some point?
i don't think so, but big things will happen soon in the registry world as the revolution against npm began (almost all of the historial employee are now out, some of them rolling their own infra)
#### are the features going to work their way back into npm?
no and none of them really wants that, as they want to captivate their users ;(
#### does yarn have it's own module listing infrastructure or does it rely on the npm catalogue?
it rely on npm infra, as most of the current replacement for npm, some alternative package registry will come soon, but will be drop in replacement of npm (entropic, github have something in mind, too)
#### would i need to go to two places to understand adapt's plugin management architecture (the npm and yarn websites) and is that tooling ambiguity worth the extra features?
for now, I don't think so, [workspaces](https://yarnpkg.com/en/docs/workspaces) seems like something adapt can use, but the price is too high imo

#### if we use npm does that stop us using yarn in the future and vice-versa?
As of today, only yarn have yarn specific feature that locks in.

oliverfoster:
duplicate

Issue #2363, opened 2019-02-25
Drop 'src/course' folder in favour of 'build/course'


oliverfoster:
### Advantages
* makes src folder "code only"
* unifies build pathways across AAT and Fwk
* deduplicates course assets when stored on disk
* allows developers to concentrate on build output
* simplifies grunt code

### Suggested timeline
Release with v6

tomgreenfield:
Please also include cleaning of build folder.

Reference: https://github.com/adaptlearning/adapt_authoring/issues/2235.

guywillis:
:+1:

PR #3324, opened 2022-08-18
Adapt Framework Version 6 Prototype

oliverfoster:
### What's changed?
1. We have NPM install support for plugins via our [normal registry server](https://github.com/adaptlearning/adapt-registry-server). The server is publically available for those who wish to host their own registries. Any plugins registered in the server should be installable via npm using the registry at https://adapt-bower-repository.herokuapp.com/npm/. With:
```sh
//.npmrc
registry=https://adapt-bower-repository.herokuapp.com/npm/
```
2. The core has been broken into the following pieces:
[adapt-contrib-a11y](https://github.com/adaptlearning/adapt-contrib-a11y) Accessibility layer
[adapt-contrib-core](https://github.com/adaptlearning/adapt-contrib-core) Model/view registry, tracking, offlineStorage, fixes, logging and wait queues
[adapt-contrib-data](https://github.com/adaptlearning/adapt-contrib-data) Data loader, router, scrollto, location, startcontroller
[adapt-contrib-device](https://github.com/adaptlearning/adapt-contrib-device) Device identification and breakpoints
[adapt-contrib-drawer](https://github.com/adaptlearning/adapt-contrib-drawer) Side drawer
[adapt-contrib-mpabc](https://github.com/adaptlearning/adapt-contrib-mpabc) Menu, page, article, block and component interfaces
[adapt-contrib-navigation](https://github.com/adaptlearning/adapt-contrib-navigation) Navigation
[adapt-contrib-notify](https://github.com/adaptlearning/adapt-contrib-notify) Notify
3. We've gotten rid of `src/course` in favour of `build/course`
4. React, bower and semver are all now installed node modules.
5. It will be possible to use (most/some) npm modules in courses moving forwards (tbc as to limitations).

### Testing
1. Run the following commands
```bash
git clone https://github.com/adaptlearning/adapt_framework v6-test
cd v6-test
git checkout issue/v6
cd src/
npm install
cd ../
npm install
grunt dev
```
2. Once built, cancel the dev task with `ctrl+c`, then run
```js
grunt server
```
3. Please have a look around and post any comments on this pr

### Todo
* LESS: Much of the styling needs to be moved into the right extensions
* SCHEMAS: The grunt tasks do not yet apply defaults from the new AAT schemas, the old AAT schemas will not work with v6 without sustantial work but the new AAT schema styles should be able to do what's necessary, they have yet to be cleaned up and trimmed
[...]

Issue #3044, opened 2021-02-24
v6 proposal


oliverfoster:
# v6 proposal

## npm modules
Prototype: [issue/2526](https://github.com/olivermartinfoster/adapt_framework/tree/issue/2526)

### How
1. Drop `src/core`, move code to `adapt-contrib-core`
2. Add `package.json` to each plugin
3. Drop `src/course` in favour or `build/course`
4. Add `src/package.json` and `src/index.js`

### Problems
1. Zip uploads to the AAT
   * Does not work with node_modules which must be installed
   * Need to develop a solution: local linked plugins / overrides folder / other
2. Should move from `libraries/` to installed npm dependencies (jquery, backbone, underscore etc)
3. Cannot allow scripts in the `src/node_modules` folder as will allow plugins to execute custom code on the AAT
4. adapt-cli needs modifying to accomodate `npm` instead of `bower`

## backbone + jsx
PR: [#2963](https://github.com/adaptlearning/adapt_framework/pull/2963)

1. Integrate with `npm` modules, move `react` and `react-dom` from manually injected dependencies to `package.json` dependencies
2. Allow react components as adapt components without backbone views, develop backbone.model <> react best practise

## navigation item management
PR: [#2864](https://github.com/adaptlearning/adapt_framework/pull/2864)

1. Develop json configuration standard
2. Migrate existing extensions to use new API

moloko:
v6 might also be a good opportunity to sort out the whole `_isA11yCompletionDescriptionEnabled` vs `_disableAccessibilityState` thing

moloko:
and maybe #1743 too?

oliverfoster:
I need to make issues for:
* npm cli rewrite
* additions of package.json

PR #3370, opened 2023-03-02
Breaking: v6 with backward compatibility

oliverfoster:
fixes https://github.com/adaptlearning/adapt_framework/issues/2526
fixes https://github.com/adaptlearning/adapt_framework/issues/2363

original issues https://github.com/adaptlearning/adapt_framework/issues/3044
original prototype https://github.com/adaptlearning/adapt_framework/pull/3324

### What's changed?
1. We have NPM install support for plugins via our [normal registry server](https://github.com/adaptlearning/adapt-registry-server). The server is publically available for those who wish to host their own registries. A registry server can be setup to fallback to any other public npm registry. The default server fallback registry is to the npm public registry but it is possible to chain to the adapt registry which then chains to the npm one. Any plugins registered in the server should be installable via npm using the registry at https://adapt-bower-repository.herokuapp.com/npm/. With:
```sh
//.npmrc
registry=https://adapt-bower-repository.herokuapp.com/npm/
```
2. We've gotten rid of `src/course` in favour of `build/course`
3. It will be possible to use (most/some) npm modules in courses moving forwards (tbc as to limitations).
4. Most existing plugins should be compatible (tbc, with the addition or injection of a package.json)

### Testing
1. Run the following commands
```bash
git clone https://github.com/adaptlearning/adapt_framework v6-test
cd v6-test
git checkout issue/v6
cd src/
npm install
cd ../
npm install
grunt dev
```
2. Once built, cancel the dev task with `ctrl+c`, then run
```js
grunt server
```
3. Please have a look around and post any comments on this pr

### Todo
* SCHEMAS: The grunt tasks do not yet apply defaults from the new AAT schemas, the old AAT schemas will not work with v6 without sustantial work but the new AAT schema styles should be able to do what's necessary, they have yet to be cleaned up and trimmed. We need to decide if v6 will support the classic AAT / vice-versa.
* [adapt-cli](https://github.com/adaptlearning/adapt-cli) is not yet compatible with v6 but the `adapt install` command etc will be made v6 compatible https://github.com/adaptlearning/adapt-cli/issues/175
* grunt build `includes` do not yet work with dependent extensions, such that if resources is included but drawer is not, drawer will not yet be added to the construction phase


grunt/config/javascript.js
```diff
@@ -2,136 +2,50 @@ module.exports = function(grunt, options) {
...
         plugins: [
-          '<%= sourcedir %>components/*/bower.json',
-          '<%= sourcedir %>extensions/*/bower.json',
-          '<%= sourcedir %>menu/<%= menu %>/bower.json',
-          '<%= sourcedir %>theme/<%= theme %>/bower.json'
+          '<%= sourcedir %>/node_modules/adapt-*/package.json',
+          '<%= sourcedir %>/node_modules/adapt-*/bower.json'
         ],
-        pluginsPath: '<%= sourcedir %>plugins.js',
+        pluginsPath: '<%= sourcedir %>/node_modules/plugins.js',
```
chris-steele, line 14: Might be redundant?

grunt/tasks/javascript.js
```diff
@@ -196,13 +262,32 @@ module.exports = function(grunt) {
...
             const mapPart = mapParts.find(part => moduleId.startsWith(part));
             if (mapPart) {
               // Remap module, usually coreJS/adapt to core/js/adapt etc
-              moduleId = moduleId.replace(mapPart, options.map[mapPart]);
+              moduleId = moduleId.replace(mapPart, mappings[mapPart]);
             }
             // Remap ../libraries/ or core/js/libraries/ to libraries/
             moduleId = Object.entries(externalMap).reduce((moduleId, [ match, replaceWith ]) => moduleId.replace((new RegExp(match, 'g')), replaceWith), moduleId);
+            const externalPart = externalParts.find(part => moduleId.startsWith(part));
+            const isExteral = Boolean(externals[externalPart]);
```
chris-steele, line 276: typo (sorry for the trivia!)

oliverfoster: not trivia. thanks

grunt/config/scripts.js
```diff
@@ -4,14 +4,12 @@ module.exports = function(grunt, options) {
       outputdir: '<%= outputdir %>',
       sourcedir: '<%= sourcedir %>',
       plugins: [
-        '<%= sourcedir %>components/*/bower.json',
-        '<%= sourcedir %>extensions/*/bower.json',
-        '<%= sourcedir %>menu/<%= menu %>/bower.json',
-        '<%= sourcedir %>theme/<%= theme %>/bower.json'
+        '<%= sourcedir %>node_modules/adapt-*/pacakge.json'
```
chris-steele, line 7: typo `pacakge.json`

grunt/tasks/javascript.js
```diff
@@ -131,46 +143,102 @@ module.exports = function(grunt) {
...
+        }, pattern).forEach(templatePath => {
+          const requireJSRootPath = templatePath.substr(options.baseUrl.length);
+          return reactTemplatePaths.push(requireJSRootPath.replace(convertSlashes, '/'));
+        });
       });
 
+      const isProduction = false;
+      // These will be overridden/added to in the package.json of any plugin
+      //  they are here for backward compatibility only
+      const v5toV6DefaultMappings = {
```
chris-steele, line 186: with the maps and paths now specified here can we remove the corresponding config from [scriptLoader](https://github.com/adaptlearning/adapt-contrib-core/blob/master/required/adapt/js/scriptLoader.js)?

oliverfoster: No. The ones in the scriptLoader are client-side and they happen before adapt.min.js is loaded, these mappings happen at the top of adapt.min.js.

We will be able to remove some of them as we move over to node_modules, specifically: `regenerator-runtime`, `core-js`, `react`, `react-dom`, `html-react-parser` and `semver`. The others will be more difficult.

chris-steele: ah, of course, though is the scriptLoader map used at all?

chris-steele: The only use I saw was spoor's offline_API_wrapper.js, but I think that has recently been moved into compilation, right?

oliverfoster: The `map` properties are used heavilly in `adapt-devtools`. The `paths` section is used further down the `scriptLoader.js` to load the foundational libraries+templates and some of those libraries are used by name at various places throughout the core and plugins.

oliverfoster: Effectively, when core is updated to switch to `node_modules`, we can also drop the `scriptLoader.js` directives, where reasonable to do so considering backward compatibility and load order. `jQuery`, `jqueryMobile`, `underscore`, `backbone`, `handlebars` and `Modernizer` will be most difficult I imagine. `imageReady`, `inview` and `jquery.resize` are libraries I made, so they should be quite easy to integrate into the core. `bowser`, `scrollTo`, `requirejs` and `velocity` I haven't analysed yet.

PR #3383, opened 2023-03-31
New: Introduce Jest component testing to the framework (fixes #3417)

### LINKED ISSUE #3417: Introduce Jest Unit Tests
Jest is a popular unit testing framework.
In the attached pull request I have unit tested the `Header` component.
You can run the test with the following command: `yarn jesttest`
Ensure that all packages are up to date when running the tests for the first time with `npm install`.

**IDE Considerations**
There is the option to add the `Jest` extension from within Visual Studio Code.
Screenshot 2023-05-17 at 14 14 59

This provides a visual view of all running tests. It automatically runs all tests in watch mode and can be used to debug tests.
Screenshot 2023-05-17 at 14 17 58

I have added in the `jest` env option to the eslinter to ensure the IDE recognises all `jest` references to avoid eslint errors appearing.

**Unit Test Approach**
To begin with we want to target all `core` jsx component templates to test that the component state gets updated correctly and the template renders as expected.
```
it('should display the header body', () => {
    const body = 'test body text';
    const props = {
      body
    };
    act(() => {
      render(<Header {...props} />, container);
    });
    expect(container.textContent).toBe(body);
  });
```
Here is an example component unit test above, it passes the `body` property to the `Header` component and then asserts that the `Header` templates' text content is equal to the `body` text that was passed into the template.

**Folder Structure**
All tests reside in a `__tests__` directory. The structure of adding tests adjacent to the file they are testing and adding a `test.js` or `spec.js` extension to the end of the test file name was not an option for us as all tests are to be added to the framework. We do not require the `spec.js` or `test.js` named extension to the end of the test file names we have in the `__tests__` directory. 
We are required to mock our ES6 class files. These will be contained within a `__mocks__` directory. The structure of this can be seen below. The example is relevant to `core/js/adapt.js` mock and Header component test.
```
-__mocks__
  -core
    -js
      -adapt.js
-__tests__
  -header.js
```

**Further Code Considerations**
Although we must have a base mock for our ES6 Classes of which are contained within the `__mocks__` directory, we can override some of the default properties/returned function calls within individual tests. I have included an example below for how to override a returned function call.
This is the device mock where we originally set the returned value of `isScreenSizeMin` to `true`.
```
const mockDeviceInstance = {
  isScreenSizeMin: jest.fn().mockReturnValue(true)
};

const mockIsScreenSizeMinMock = jest.fn().mockReturnValue(true);
export default {
  __esModule: true,
  default: mockDeviceInstance,
  get isScreenSizeMin() {
    return mockIsScreenSizeMinMock;
  }
};
```

Within the test we can override this with a spy, an example of this is given below.
`jest.spyOn(device, 'isScreenSizeMin').mockReturnValue(false);`
This will ensure the `device.isScreenSizeMin('medium')` call within the `Header.jsx` template returns false when the test renders the template.

We can override properties from classes within individual tests by first ensuring the property is set up as a default value within the mock. Then in the test file, import the class and modify the property, an example can be seen below.

This is a mock with a property `screenSize` set to `small`.
```
const mockDeviceInstance = {
};

const mockIsScreenSizeMinMock = jest.fn().mockReturnValue(true);
export default {
  __esModule: true,
  default: mockDeviceInstance,
  screenSize: 'small'
};
```

Within the test you can then import the Device class and set the `screenSize` property to a value of your choice.
```
import Device from '../src/core/adapt-contrib-core';
it('es6 class property setting example', () => {
  Device.screenSize = 'small';
}
```

Note that we must also provide a `default` property in our mock with a class instance for when the class is exported with `export default ClassName`.

**Question/Further Considerations**
Are there any plans to unit test `JS` files? For example we could unit test our accessibility file a11y and ensure the functions are returning the correct values.

eleanor-heath:
Please see details here https://github.com/adaptlearning/adapt_framework/issues/3417

Steps to run
Navigate to the adapt_framework repo in the terminal and run 

```sh
# to setup
npm install

# to see help screen
npm test help

# to execute
npm test
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
 
       - uses: actions/setup-node@v3
         with:
-          node-version: "14"
+          node-version: "16"
           cache: "npm"
 
       - name: Installing Environment
--- a/test.js
+++ b/test.js
@@ -1,4 +1,3 @@
-
 const os = require('os');
 const { spawn, exec } = require('child_process');
 const jest = require('jest');


cahirodoherty-learningpool (approved):
👀

coverage/coverage-final.json
cahirodoherty-learningpool, line 1: This looks like it shouldn't have been committed. You might have to delete it, commit that change and then add it to the .gitignore file and make a further commit

__tests__/header.js
```diff
@@ -0,0 +1,161 @@
...
+
+  it('should be able to set the displayTitle', () => {
+    const displayTitle = 'test body text';
+    const props = {
+      displayTitle
+    };
+    act(() => {
+      render(<Header {...props} />, container);
+    });
+    expect(container.textContent).toBe('');
```
cahirodoherty-learningpool, line 54: Should this be:
```suggestion
    expect(container.textContent).toBe(displayTitle);
```

eleanor-heath: The displayTitle is not compiled in the same way the other props are using `dangerouslySetInnerHTML` so Im not sure there is a way to assert the displayTitle is shown.

cahirodoherty-learningpool: ah right, thanks. In that case it may be best to remove this and perhaps add a TODO
We don't want to mislead that we have coverage to check displayTitle when we aren't really doing that

oliverfoster:
### Yarn
We don't recommend using yarn anywhere else, we should have the instructions in the npm equivalent:
```js
npm run jesttest
npm run cleanjest
```

### Test suite
As we have two testing suites now, e2e and unit, it might be wise to have a facade on the testing stuff. That would include 1 root folder `test` containing the e2e and unit test files, minus their configs which will need to stay in the root. It might also be sensible to have a single command, the `npm test` command, able to run both `npm test unit` and `npm test e2e`.

### Config issues
It's running `./test.js`, kicking off the e2e tests.
image
Causing an `adapt install` to render last.
image
That can be fixed by enabling and using only the first line of this config:
https://github.com/adaptlearning/adapt_framework/blob/e3743b438a8df95e38260e23f03d9f8dcdb36a7e/jest.config.js#L156-L159

### ReactDOM.render
image
We should probably use the same version of React and ReactDOM as the version we're using in the core, v17.0.2. Is there any specific other reason for using v18?

### Mocks
I couldn't understand how these were being used instead of the normal core/js/adapt etc files, which are imported from within the header template. So I deleted them to assess the consequences.

Mocks are automatically loaded. Cool.
Without mocking the APIs, jest could not load the original files.
image
This is as jest is running from `/` and not `/src/`, that can be changed by enabling and adding `src/` to to this config:
image
That allows jest to load the original APIs where any test does not have a mock but requires something from the `src/` folder.

Irritatingly, because the core `libraries/` are not loaded using `import` directives, they will not be loaded into the jest environment when the tests are running. Giving rise to errors like this:
image
This currently necessitates mocking for anything which uses `Backbone` or `Underscore` or any other `libraries/` file. Solving this problem will only be sensibly done, when we use those libraries via import directives, hopefully in v6.

oliverfoster:
# Next steps

## This pr
### Files and folders
* Move all relevant testing files to a root `test/` folder for the framework and plugins (including core and including the files for e2e testing)
* Ensure that the tests and mocks can be executed from the plugins

### Commands
* Run e2e from `npm test e2e` and `npm test`
* Run jest from `npm test unit` and `npm test`
* Where running any variation of `npm test [type]`, ensure that all tests are executed and that a final exit code is sent on pass or fail for continuous integration (CI) systems
* Test execution needs to be broken into two phases, the preparation and the execution, such that the test results from e2e and unit tests appear at the bottom of the console output together when running `npm test`

### Jest Environment
* The [environment options](https://jestjs.io/docs/configuration#testenvironmentoptions-object) object needs to be passed through with:
  * An `--outputdir=` value, identical to the `grunt build --outputdir=` option, so that we can perform tests where the `build/` folder is detached as it is in the AAT
  * The framework `__mock__` folder final location

## Other plugins
Libraries from the `libraries/` folders are loaded into a `use strict` environment (I have yet to figure out how to stop this, although I'm not sure it will be possible to do so), which causes some of the code in those libraries to fail compilation. It can be remediated by:
* Removing some mediaelement-and-player.js global delete statements
* Using `window.require` instead of `require` in adapt-contrib-spoor.js
* Return early in wrapper.js if Adapt.course does not exist

## Jest Environment
* The library [jest-environment-adapt](https://github.com/cgkineo/jest-environment-adapt) needs moving into adaptlearning org 
* The [environment options](https://jestjs.io/docs/configuration#testenvironmentoptions-object) object needs to be passed through with:
  * An `--outputdir=` value, identical to the `grunt build --outputdir=` option, so that we can perform tests where the `build/` folder is detached as it is in the AAT. The option should be used [here](https://github.com/cgkineo/jest-environment-adapt/blob/b3bef48cbcab4858dee8175b526a6fa21768e773/index.js#L12), [here](https://github.com/cgkineo/jest-environment-adapt/blob/b3bef48cbcab4858dee8175b526a6fa21768e773/index.js#L24), [here](https://github.com/cgkineo/jest-environment-adapt/blob/b3bef48cbcab4858dee8175b526a6fa21768e773/index.js#L36), [here](https://github.com/cgkineo/jest-environment-adapt/blob/b3bef48cbcab4858dee8175b526a6fa21768e773/index.js#L61) and [here](https://github.com/cgkineo/jest-environment-adapt/blob/b3bef48cbcab4858dee8175b526a6fa21768e773/index.js#L105)
  * The framework `__mock__` folder final location, so that the plugins mock can be generated [here](https://github.com/cgkineo/jest-environment-adapt/blob/b3bef48cbcab4858dee8175b526a6fa21768e773/index.js#L75)
* The library needs documenting properly with jsdoc and a readme
* The library needs publishing on npm

oliverfoster:
# Update

#### Commands
image

### Test file layouts
1. Cypress/e2e tests should live in each plugin at the folder `test/e2e/name.cy.js`
2. Jest/unit tests should live in each plugin at the folder `test/unit/name.js`
3. Jest global mocks should live at  `test/unit/__mock__/name.js`
image

See [example test folder](https://github.com/adaptlearning/adapt_framework/tree/Framework-unit-testing/test).

### Notes
I've moved all of the code for executing tests into [`test.js`](https://github.com/adaptlearning/adapt_framework/blob/Framework-unit-testing/test.js) to keep things very simple and the testing architecture out of the test definition folders.

### Libraries
[jest-environment-adapt](https://www.npmjs.com/package/jest-environment-adapt) contains all of the code required to setup jest to be able to load Adapt into node.

test/unit/header.js
```diff
@@ -0,0 +1,182 @@
...
+  it('should display the component description when the title, display title or instruction is not specified ' +
+    'and the _isA11yComponentDescriptionEnabled is enabled and the aria region is set', () => {
+    const props = {
+      _isA11yComponentDescriptionEnabled: true,
+      _component: 'text'
+    };
+    act(() => {
+      render(<Header {...props} />, container);
+    });
+    const ariaRegion = 'Text Component Aria Text';
```
cahirodoherty-learningpool, line 98: Where do these defaults/fallbacks come form? Can we import them possibly?

test/unit/header.js
```diff
@@ -0,0 +1,182 @@
...
+  global.console = console;
+  // await import('core/js/app.js');
+  // await new Promise((resolve, reject) => {
+  //   const a = setInterval(() => {
+  //     if (!Adapt.get('_isStarted')) return;
+  //     clearInterval(a);
+  //     resolve(Promise.resolve());
+  //   }, 250);
+  // });
+});
```
eleanor-heath, line 20: is this still needed?

test/unit/header.js
oliverfoster: We could. I'm not sure if it's wiser to have defined, uniform test text, or if they should be made up for each test.

cahirodoherty-learningpool: Happy to leave as-is for now. Might be something we tighten up later 👍

cahirodoherty-learningpool (approved):
👀 
Excellent!

eleanor-heath:
Approved ✅ extremely happy with this and great we will now have both e2e and unit tests within the Adapt Framework.

oliverfoster (approved):

ethan-lp (approved):

PR #3408, opened 2023-04-28
New: Breakpoints derived from schema defaults

oliverfoster:
fixes [#362](https://github.com/adaptlearning/adapt-contrib-core/issues/362)

The process now applies defaults from the schema to the config.json where missing, then reads the values from the config and produces less and css variables from the name/values pairs in the config. Breakpoints are therefore no longer hardcoded.

### New
* Breakpoints are derived from schema defaults and config

### DIFF
--- a/grunt/helpers/Framework.js
+++ b/grunt/helpers/Framework.js
@@ -185,6 +185,23 @@ class Framework {
     return this;
   }
 
+  /** @returns {Framework} */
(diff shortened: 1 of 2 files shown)


cahirodoherty-learningpool (approved):
👀

StuartNicholls (approved):
👀

swashbuck (approved):
👍

PR #3413, opened 2023-05-09
Fix: Apply screenSize defaults after json copy

### LINKED ISSUE #3412: Apply screenSize defaults with globals defaults instead of with less compilet
### Subject of the issue

`framework.applyScreenSizeDefaults();` is in the wrong place.

Move from:
https://github.com/adaptlearning/adapt_framework/blob/34e39ff815047024afb0a2d14ae8d2c482077cdc/grunt/tasks/less.js#L5-L24

To:
https://github.com/adaptlearning/adapt_framework/blob/34e39ff815047024afb0a2d14ae8d2c482077cdc/grunt/tasks/schema-defaults.js#L1-L7

For better, more predicatable coverage.

oliverfoster:
fixes #3412 

### Fix
* Apply screenSize defaults after json copy rather than with less compile, as the json copy can overwrite the defaults from src/course > build/course

### DIFF
--- a/grunt/tasks/less.js
+++ b/grunt/tasks/less.js
@@ -1,6 +1,5 @@
 module.exports = function(grunt) {
   const convertSlashes = /\\/g;
-  const Helpers = require('../helpers')(grunt);
 
   grunt.registerMultiTask('less', 'Compile LESS files to CSS', function() {
     const less = require('less');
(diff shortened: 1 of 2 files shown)


cahirodoherty-learningpool (approved):
👀

swashbuck (approved):
👍

PR #3416, opened 2023-05-16
Fix: Apply schema defaults as a safeguard for AAT missing defaults

### LINKED ISSUE #3415: 'server-build' grunt task fails since 5.31.3
When performing an Adapt authoring tool build of a course using the new 'adapt_framework' v5.31.3, the build fails (error given below in 'actual behaviour').  This appears to be because the change that was applied in v5.31.3 was applied to the 'build' grunt task, but not the 'server-build' grunt task, the latter of which assumes the screen config steps are still performed as part of the 'less' build task.

### Your environment
* version (AT/Framework) : 5.31.3
* which browser and its version: Chrome, 113.0.5672.93
* device(s) + operating system(s): Windows desktop PC

### Steps to reproduce
Install the adapt authoring tool (0.11.1), and link it to the adapt_framework v5.31.3.  Create a new course, and immediately attempt to publish/build.  After a short delay, the error mentioned above will be presented.

### Expected behaviour
The course should build! - or rather the server-build process should complete in the same manner that the build process does.

### Actual behaviour
The course fails to build, with the following error:

error [16 May 23 08:11:46 +01:00] Command failed: npx grunt server-build:prod --outputdir=courses\6463226d1cf94b72306f8679\6463230edd525d608017feed\build --theme=646322c31cf94b72306f870d --menu=adapt-contrib-boxMenu
Error: No screenSize defined in config.json

oliverfoster:
fixes #3415 

### Fix
* Apply schema defaults as a safeguard for AAT missing defaults

### DIFF
--- a/grunt/tasks/server-build.js
+++ b/grunt/tasks/server-build.js
@@ -10,6 +10,7 @@ module.exports = function(grunt) {
       'build-config',
       'copy',
       'scripts:adaptpostcopy',
+      'schema-defaults',
       'language-data-manifests',
       'less:' + requireMode,
       'handlebars',


simondate (approved):
👀

guywillis (approved):
👀

PR #3428, opened 2023-06-06
Custom jest environment - Allow ES Class unit tests

oliverfoster:
### Fix
* Reduced react and react-dom versions to 17.0.2 to be inline with the adapt varieties
* Removed mocks for adapt es classes and used classes instead

### New
* Created and used [jest-environment-adapt](https://github.com/cgkineo/jest-environment-adapt) to load `scorm_test_harness.html` and all associated files, up until `adapt.min.js`, whereby tests can `import Adapt from 'core/js/adapt'` etc, and test the code directly.
* Mocked `backbone`, `html-react-parser` to steal from the Adapt loaded versions 
* Mocked `bowser` to override appropriately
* Mocked `plugins` dynamically, to allow `import 'core/js/app';` to load the compiled plugins, which will allow loading a full version of adapt into the testing environment for more complex json configured tests

### Testing
* One must run `grunt dev` before running `npm run jesttest` as files are loaded from the `build/` folder to setup the test environment
* Libraries are loaded in a `use strict` environment (have yet to figure out how to stop this), so mediaelement-and-player.js has some `delete` statements which fail
* adapt-contrib-spoor.js uses `require` from the global scope, it should use `window.require` instead
* wrapper.js errors when trying to find `Adapt.course` as it's not loaded yet, so a return early must be inserted to fix

image


cahirodoherty-learningpool:
Great work Oliver!
I was wondering what the maintenance plan for [jest-environment-adapt](https://github.com/cgkineo/jest-environment-adapt) would be? Is it at all possible to bundle it directly in with the FW code?

Also, to handle the build/ folder requirement, could the package.json script be updated to ```"jesttest": "grunt dev && jest",```

eleanor-heath (approved):
this is fantastic, so happy to see :)

oliverfoster:
> Great work Oliver! I was wondering what the maintenance plan for [jest-environment-adapt](https://github.com/cgkineo/jest-environment-adapt) would be? Is it at all possible to bundle it directly in with the FW code?

I don't think it's possible to bundle the code directly, as one has to specify a module name in the [config](https://jestjs.io/docs/configuration#testenvironment-string), I'll have a look see if there are alternative ways of extending the `jest-environment-jsdom` module which are more integrated. Obviously, this code will be fw version specific, but its version is specified in the fw package.json, so it shouldn't be too difficult to maintain alongside.

It's not finished at the moment, I'd like all of the hardcoded strings to be configurable from the jest config. The `build/` folder should be configurable in a similar way to the `--outputdir=` for grunti.

> Also, to handle the build/ folder requirement, could the package.json script be updated to `"jesttest": "grunt dev && jest",`

Probably `grunt diff` as it doesn't wait at the end. The `jesttest` command is under contention from [my comments on the main pr](https://github.com/adaptlearning/adapt_framework/pull/3383#issuecomment-1576617186). I think we should have one unified testing api, even though we're using two frameworks to pull it off. `npm test unit` and `npm test e2e` rather than `npm run jesttest` and `npm test`.

__tests__/header.js
```diff
@@ -1,16 +1,48 @@
...
+
 let container = null;
 beforeEach(() => {
   jest.clearAllMocks();
   // setup a DOM element as a render target
   container = document.createElement('div');
+
   document.body.appendChild(container);
+  Adapt.course = {
+    get: jest.fn().mockReturnValue(
```
oliverfoster, line 30: @eleanor-heath I'm pretty sure there is a much better way of mocking this Adapt.course.get function, but I didn't manage to read that far into the jest api.

oliverfoster:
We should probably have a meeting to discuss what this code actually does?

eleanor-heath:
> We should probably have a meeting to discuss what this code actually does?

Yes please, I read your previous comment here https://github.com/adaptlearning/adapt_framework/pull/3383#issuecomment-1576617186 so understand a lot of what has been done based on the comment you have added but a call to discuss unit testing in general and our plans would be ideal.

joe-allen-89 (approved):
Really good work all involved on this, it is fairly straight forward to get up to speed with and works as expected, going to be very useful going forward.

PR #3471, opened 2023-10-06
Chore: Added drawer extension e2e tests (Issue/3470)

lemmyadams:

addresses adaptlearning/adapt-contrib-resources#111

### DIFF
--- a/test/e2e/extensions/drawer.cy.js
+++ b/test/e2e/extensions/drawer.cy.js
@@ -38,7 +38,7 @@ describe('Drawer', () => {
         checkDrawerLength(2)
     });
 
-    it('ahouls display the correct drawer items', () => {
+    it('should display the correct drawer items', () => {
         cy.get('.drawer__item').each(($item, index) => {
             cy.get($item).within(() => {
                 cy.get('.drawer__item-title').should('contain', Course._resources._resourcesItems[index].title)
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

test/e2e/extensions/drawer.cy.js
```diff
@@ -0,0 +1,63 @@
+import Course from '../../../src/course/en/course.json'
```
oliverfoster, line 1: Interesting... very interesting...

The tests can be run against a specified build output folder, which wouldn't have a `src/course` but an `{outputdir}/{coursedir}/` instead. So we're going to need some way of grabbing that stuff in the test runtime such that the json can be loaded from the correct place.

oliverfoster: A few people have mentioned they'd like to perform tests based upon the course content, effectively removing some Quality Assurance tasks. We should start from facilitating that point of view, even if we never get there. As the first test, this already relies on the defined course data.

oliverfoster: The resource specific e2e tests should live in resources.

joe-allen-89:
New data loading needs to be implemented.

cahirodoherty-learningpool:
Replaced with https://github.com/adaptlearning/adapt-contrib-boxMenu/pull/168 and https://github.com/adaptlearning/adapt-contrib-resources/pull/112

Issue #3472, opened 2023-10-10
Testing bugs


oliverfoster:
### Subject of the issue
1. Need to split the `adapt install` and `grunt diff` into two separate prepare tasks as an `adapt devinstall` will often be used for test creation
2. Fix the `npm test **/globForTestsToRun/**` as it doesn't currently work
3. Resolve the issue of loading json from the outputdir for cypress jobs

PR #3489, opened 2023-12-04
New: Cypress task log added, cypress commands file added, cypress get…

joe-allen-89:
…Data command added.

Part 3 of the issue #3472

### Description
* Automatically Loads a `/test/e2e/commands.js` file if available through the `supportFile` directive in the cypress config
* Adds cypress globals to eslint so that the linter doesn't warn unnecessarily on cypress test files
* Allowed the `outputdir` to be the fixtures root for cypress, such that any file can be loaded from the `outputdir` into the cypress environment
* Expose `cy.getBuild()` to return the `adapt/js/build.min.js` json
* Expose `cy.getConfig()` to return the `{build.coursedir}/config.json`
* Expose `cy.getData(languagecode)` to return an array of all the data from `{build.coursedir}/{languagecode}/`
* Expose the array `this.data` inside tests, containing all of the data items from the adapt course language folder
* Expose `this.build, this.config, this.data.course, this.data.contentObjects, this.data.articles, this.data.blocks, this.data.components` as shorthands as they are in adapt
* Expose `cy.getBuild().then(build => {}), cy.getConfig().then(config => {}),  cy.getData('en').then(data=> {})`

### New
* Cypress tasks added to config to log messages
* Single commands file added to e2e
* getBuild, getConfig and getData command added to load data using outputDir

### Fix
* grunt language loader to fail silently on non-language subfolders in coursedir 

### Update
* Added cypress globals to eslint


test/e2e/menuPage.cy.js
```diff
@@ -1,8 +1,13 @@
+import {getData} from './commands'
+
 describe('Menu Page', () => {
-  const pageTitle = 'Adapt Version 5';
 
-  it(`should have the title ${pageTitle}`, () => {
+  beforeEach(() => {
```
cahirodoherty-learningpool, line 5: I think we could reduce calls here by just making this a `before` for one run instead of on each test scenario

cahirodoherty-learningpool, line 5:
```suggestion
  before(() => {
```

I think we could reduce calls here by just making this a `before` for one run instead of on each test scenario

oliverfoster: You'll want clean data for each test. If any test modifies the data, the following test will otherwise have to be aware of changes in the preceding tests. The best practise would be to load the data fresh for each test or at least clone it each time.

joe-allen-89: Just to add to this, Cypress will also clear aliases after each test to avoid data manipulation issues.

eleanor-heath (approved):

test/e2e/menuPage.cy.js
```diff
@@ -1,8 +1,13 @@
+import {getData} from './commands'
```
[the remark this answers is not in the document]
joe-allen-89: The import of the command file is still needed, but not the naming of the function. That's been amended.

cahirodoherty-learningpool (approved):
👀

test/e2e/commands.js
```diff
@@ -0,0 +1,46 @@
...
+    const allFileData = [];
+    cy.fixture(manifest).as('manifest_data').then((data) => {
+      data.forEach((item) => {
+        const name = path.parse(item).name;
+        cy.fixture(item).as(`${name}Data`).then((data) => {
+          allFileData.push(data);
+        });
+      });
+    }).then(() => {
+      // Potentially flatten data as with FW data.js
```
oliverfoster, line 33: Flattening nested arrays is much easier now.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

```js
[[1,2,3,4],5,6,[[7,8,9],10]].flat(Infinity)
```
It is very much easier to sort and filter the data if it's flattened.
In adapt we add the file path to __path__ and the index for each path at __index__:

https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/data.js#L114C9-L114C30

https://github.com/adaptlearning/adapt-contrib-core/blob/852ef1269dcdfc220361d500c473c9042ffa65f4/js/data.js#L147-L148

oliverfoster (commented):
Is it possible to put the commands.js contents at the bottom of cypress.config.js?

test/e2e/commands.js
joe-allen-89: @oliverfoster best practice according to Cypress is to have custom commands in a commands file, I think the config file is loaded too early to be able to set aliases, but I will double check that.

oliverfoster:
These two issues probably need fixing:

* If the course folder name is changed then the data won't load as it's hard coded [here](https://github.com/adaptlearning/adapt_framework/blob/9e4ece1be23def40704714bd4d4b42ba7959b780/test/e2e/commands.js#L47)
* If the language is changed in the tests the test data will only be in the language specific as [_defaultLanguage](https://github.com/adaptlearning/adapt_framework/blob/9e4ece1be23def40704714bd4d4b42ba7959b780/test/e2e/commands.js#L54C1-L54C1) in the config.json

test/e2e/menuPage.cy.js
```diff
@@ -1,8 +1,11 @@
 describe('Menu Page', () => {
-  const pageTitle = 'Adapt Version 5';
 
-  it(`should have the title ${pageTitle}`, () => {
+  beforeEach(() => {
+    cy.getData();
+  });
+
+  it('should have the correct title', function () {
```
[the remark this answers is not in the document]
oliverfoster: This whole test should be in boxMenu. It really needs to be deleted from here.

cahirodoherty-learningpool (approved):
👀

oliverfoster:
> These two issues probably need fixing:
> 
> * If the course folder name is changed then the data won't load as it's hard coded [here](https://github.com/adaptlearning/adapt_framework/blob/9e4ece1be23def40704714bd4d4b42ba7959b780/test/e2e/commands.js#L47)
> * If the language is changed in the tests the test data will only be in the language specified as the [_defaultLanguage](https://github.com/adaptlearning/adapt_framework/blob/9e4ece1be23def40704714bd4d4b42ba7959b780/test/e2e/commands.js#L54C1-L54C1) in the config.json

These are fixed. Added `getBuild` to be able to use the `availableLanguageNames` and `coursedir` from `adapt/js/build.min.js`. Added `getConfig` to be able to load the `${coursedir}/config.json`. Allowed `getData()` to load a specified or the default language.

cahirodoherty-learningpool (approved):
Nice job 👌

cahirodoherty-learningpool:
@oliverfoster @joe-allen-89 
You both have contributed to this, so not sure what our approval process should be. Are you both happy with it in it's current form?

@oliverfoster  Once this is merged, are there any more blockers outstanding or can the currently existing testing PRS be moved to their relevant plugins instead and reviewed there?

oliverfoster:
I'm happy if @joe-allen-89 is. 

The only other blocker is to provide a similar interface for the unit testing. e2e tests can proceed after this pr is merged I think.

I'm sure we'll see what other issues arise thereafter.

ethan-lp (approved):

eleanor-heath (approved):

cahirodoherty-learningpool:
Thank you both!
🙌

PR #3495, opened 2023-12-15
Fix: return early added to applyDefaults to protect against null values in AAT (fixes #3491)

### LINKED ISSUE #3491: Grunt build error on AT instances
AT installs are throwing an error on build for newer versions of the framework (not sure exactly when the issue started, but has cropped up in the last few upgrades). 

Error calling `hasOwnProperty` on `undefined` in [Schemas.js applyDefaults](https://github.com/adaptlearning/adapt_framework/blob/master/grunt/helpers/schema/Schema.js#L111) - the `output` var looks to be the culprit. I'm not sure why this would be undefined, let me know if it requires an AT fix.

### Your environment
* AT: 0.11.2, FW: 5.33.5+

### Steps to reproduce
Run a full/forced rebuild in an offending AT instance.

### Expected behaviour
Course should build successfully.

### Actual behaviour
Course does not build successfully.

joe-allen-89:

#3491 

### Fix
* Return early added to applyDefaults to protect against null values in AAT (fixes #3491)

### DIFF
--- a/grunt/helpers/schema/Schema.js
+++ b/grunt/helpers/schema/Schema.js
@@ -126,6 +126,9 @@ class Schema {
       let hasChanged = false;
       let haveChildenChanged = false;
       let defaultValue;
+
+      if (description === null || output === null) return;
+
       switch (description.type) {
         case 'object':
           defaultValue = description.hasOwnProperty('default') && options.fillObjects ? description.default : {};


taylortom (approved):
Works for me

oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

PR #3528, opened 2024-03-12
Fix: Warn if incompatible plugins are being compiled (fixes #3321)

### LINKED ISSUE #3321: warn if incompatible plugins are being compiled
plugin package.json:framework vs framework package.json and warn if semver doesn't match

swashbuck:
Fixes #3321

### New
* Throws an error if required framework version in a plugin is not satisfied by the current framework version

### Testing
1. In a plugin's _bower.json_, set `framework` to a high value like `>=7.0.0`

### DIFF
--- a/grunt/helpers/plugins/Plugin.js
+++ b/grunt/helpers/plugins/Plugin.js
@@ -1,5 +1,4 @@
 const semver = require('semver');
-
 const globs = require('globs');
 const JSONFile = require('../lib/JSONFile');
 
--- a/grunt/helpers/plugins/Plugin.js
+++ b/grunt/helpers/plugins/Plugin.js
@@ -1,3 +1,5 @@
+const semver = require('semver');
+
 const globs = require('globs');
 const JSONFile = require('../lib/JSONFile');
 
--- a/grunt/helpers/plugins/Plugin.js
+++ b/grunt/helpers/plugins/Plugin.js
@@ -108,6 +108,8 @@ class Plugin {
 
   /** @returns {boolean} */
   get isFrameworkCompatible() {
+    if (!this.framework || !this.framework.version) return;
+
     return semver.satisfies(this.framework.version, this.requiredFramework);
   }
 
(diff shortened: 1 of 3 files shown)


oliverfoster:
Awesome

oliverfoster (approved):

oliverfoster:
Error or warn? Thinking AAT...

swashbuck:
> Error or warn? Thinking AAT...

@oliverfoster Is the concern that throwing an error would prevent a course from building, but a warning would not interfere since it just outputs a message?

oliverfoster:
Only in theory. I don't actually know what would happen or what's worse or if the user would get the appropriate message.

swashbuck:
> Only in theory. I don't actually know what would happen or what's worse or if the user would get the appropriate message.

Doesn't `adapt-cli` prevent incompatible plugins from being uploaded? So, in theory, you would never get to the point where you're trying to build a course in the AAT with an incompatible plugin.

oliverfoster:
You can use the `adapt-cli` or the AAT plugin upload to install a plugin in a course. Or you can download a plugin and unzip it in the right place.

swashbuck:
> You can use the `adapt-cli` or the AAT plugin upload to install a plugin in a course. Or you can download a plugin and unzip it in the right place.

Ok, probably safer to just use a warn. Same as the "Plugin folder name ${pathDerivedName} does not match package name" message.

joe-allen-89 (approved):
👍

grunt/helpers/plugins/Plugin.js
```diff
@@ -96,6 +100,18 @@ class Plugin {
...
   }
 
+  /** @returns {string} */
+  get requiredFramework() {
+    return this.packageJSONFile.firstFileItem.item.framework;
+  }
+
+  /** @returns {boolean} */
+  get isFrameworkCompatible() {
+    if (!this.framework || !this.framework.version) return;
```
cahirodoherty-learningpool, line 110: Can we return `true` here and then apply a boolean check on line 66? (i.e. `!this.isFrameworkCompatible`)

cahirodoherty-learningpool (approved):
👀

Issue #3532, opened 2024-03-27
Cypress helper function for stripping html


lemmyadams:
### Subject of the issue/enhancement/features
With more tests in place, there's some reused code that would be best placed in a custom cypress command to reduce the risk of errors. Such as:
```
function stripHtml(text) {
  const textWithoutHtml = text.replace(/<[^>]*>/g, '');
  cy.wrap(textWithoutHtml).as('text');
}
```

oliverfoster:
```js
function testQuestionButtons(buttonsObject = {btn__action: 'Submit', btn__feedback: 'Show feedback'}) {
  Object.keys(buttonsObject).forEach((key) => {
    cy.get(`.${key}`).should('contain', buttonsObject[key])
  })
}
```
This is much more generic than it looks:
```js
function hashContains(hash) { // the name needs changing
  Object.entries(hash).forEach(([key, value]) => {
    cy.get(`.${key}`).should('contain', value)
  })
}
```
 The default values should not be hardcoded in English but instead be passed from the loaded JSON.

oliverfoster:
There are 1001 ways of doing the stripHtml function:
```js
function stripHtml(text) {
  const textWithoutHtml = text.replace(/<[^>]*>/g, '');
  cy.wrap(textWithoutHtml).as('text');
}
```
As you can see here: https://stackoverflow.com/questions/822452/strip-html-tags-from-text-using-plain-javascript

What are you going to use this for?

lemmyadams:
> There are 1001 ways of doing the stripHtml function:
> 
> ```js
> function stripHtml(text) {
>   const textWithoutHtml = text.replace(/<[^>]*>/g, '');
>   cy.wrap(textWithoutHtml).as('text');
> }
> ```
> 
> As you can see here: https://stackoverflow.com/questions/822452/strip-html-tags-from-text-using-plain-javascript
> 
> What are you going to use this for?

I'm going to use it for stripping the html. On that link currently its the 2nd bit of code, it was just the simplest design which I chose. I can pick the first one though if thats preferred or any other option.

lemmyadams:
> ```js
> function testQuestionButtons(buttonsObject = {btn__action: 'Submit', btn__feedback: 'Show feedback'}) {
>   Object.keys(buttonsObject).forEach((key) => {
>     cy.get(`.${key}`).should('contain', buttonsObject[key])
>   })
> }
> ```
> 
> This is much more generic than it looks:
> 
> ```js
[...]

PR #3533, opened 2024-03-27
Chore: Added 2 helper functions for e2e tests

lemmyadams:
Addresses #3532 

### Chore
* Added stripHtml helper function for e2e tests

### DIFF
--- a/test/e2e/commands.js
+++ b/test/e2e/commands.js
@@ -112,7 +112,20 @@ function testContainsOrNotExists(target, value) {
   }
 }
 
+function testQuestionButtons(buttonsObject = {btn__action: 'Submit', btn__feedback: 'Show feedback'}) {
+  Object.keys(buttonsObject).forEach((key) => {
+    cy.get(`.${key}`).should('contain', buttonsObject[key])
+  })
+}
+
+function stripHtml(text) {
+  const textWithoutHtml = text.replace(/<[^>]*>/g, '');
+  cy.wrap(textWithoutHtml).as('text');
+}
+
 Cypress.Commands.add('getBuild', getBuild);
 Cypress.Commands.add('getConfig', getConfig);
 Cypress.Commands.add('getData', getData);
 Cypress.Commands.add('testContainsOrNotExists', testContainsOrNotExists);
+Cypress.Commands.add('testQuestionButtons', testQuestionButtons);
+Cypress.Commands.add('stripHtml', stripHtml);
--- a/test/e2e/commands.js
+++ b/test/e2e/commands.js
@@ -112,15 +112,17 @@ function testContainsOrNotExists(target, value) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


test/e2e/commands.js
```diff
@@ -112,7 +112,20 @@ function testContainsOrNotExists(target, value) {
...
+function testQuestionButtons(buttonsObject = {btn__action: 'Submit', btn__feedback: 'Show feedback'}) {
+  Object.keys(buttonsObject).forEach((key) => {
+    cy.get(`.${key}`).should('contain', buttonsObject[key])
+  })
+}
+
+function stripHtml(text) {
+  const textWithoutHtml = text.replace(/<[^>]*>/g, '');
+  cy.wrap(textWithoutHtml).as('text');
+}
```
cahirodoherty-learningpool, line 124: Could this work as a simple return without the need to involve Cypress complexity?

```suggestion
function stripHtml(text) {
  return text.replace(/<[^>]*>/g, '');
}
```

It may be that we need to move it into a helpers file outside of the test files and reference it from there?
It just feels like the Cypress wrapper is more than would be necessary

cahirodoherty-learningpool: Upon discussion, moving to the helpers file would mean touching core and in a quite confusing way.
Disregarding the above suggestion

oliverfoster: Did you agree that the wrapper context is necessary? 

Using `const stripped = this.text` rather than `const stripped = cy.stripText()` seems really janky. The `this` variable is for context and not for parsing local variables.

lemmyadams: Yes it is necessary. Within cypress setting a const to a function doesn't work, even if that function returns a variable. I would prefer to do that of course, but it is not possible unfortunately.

oliverfoster: If you define it as:

`test/e2e/helpers.js`
```js
cy.helpers = {
  stripHtml(text) {
    let tmp = document.createElement("DIV");
    tmp.innerHTML = text;
    const textWithoutHtml = tmp.textContent || tmp.innerText || "";
    return textWithoutHtml
  }
}
```

Then at the top of the `commands.js` file put:
```js
import './helpers` 
```

You should be able to refer to it as:
```js
      const bodyWithoutHtml = cy.helpers.stripHtml(textInputComponent.body);
      cy.testContainsOrNotExists('.textinput__body', bodyWithoutHtml)
```

lemmyadams: I personally think this is more janky to have a separate file for this but I'll add it

oliverfoster: It doesn't need to be a separate file if you prefer. I don't mind. I guess if we add more we can reorganise at that point.

cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
👍

ethan-lp (approved):

PR #3548, opened 2024-04-24
New: Added server-diff and --cacheage (fixes 3547)

### LINKED ISSUE #3547: AAT compilation improvements
### Subject of the issue
* Bring `grunt diff` to the server build process, use `newer` to compile only when necessary. 
https://github.com/adaptlearning/adapt_framework/blob/5663a1380dc49c486af4b4ab7dd368ac66c7ee18/grunt/tasks/diff.js#L14-L16

* Allow the javascript compilation cache to live for longer than a week
https://github.com/adaptlearning/adapt_framework/blob/5663a1380dc49c486af4b4ab7dd368ac66c7ee18/grunt/tasks/javascript.js#L23
https://github.com/adaptlearning/adapt_framework/blob/5663a1380dc49c486af4b4ab7dd368ac66c7ee18/grunt/helpers/CacheManager.js#L11-L13

oliverfoster:
fixes #3547 

Speed up AAT compilation by only compiling when necessary and allowing javascript compilation caches to live for longer.

### New
* Added `grunt server-diff:dev` which will only compile less, handlebars and javascript if those files are newer than their output and will only copy most files if they are also newer than their output
* Allow the age of the javascript compilation cache to be changed from the command line using a millisecond integer at `grunt server-diff:dev --cacheage=1000000000000`

### Fixes
* Expanded newer configuration in javascript config to improve glob accuracy and increase speed
* Allowed schema and data load caching for schema-defaults task to increase processing speed
* Allowed the diff tasks to force build when switching from production to development and back
* Improved `grunt diff` along the same lines as `grunt server-diff`

### Notes
* The copy task is split so that the [`replace`](https://github.com/adaptlearning/adapt_framework/blob/58340235b8bb2f401dfa4a24d4c131bd765deec3/grunt/tasks/server-diff.js#L25) task always has a fresh `@@` directives to string replace
https://github.com/adaptlearning/adapt_framework/blob/58340235b8bb2f401dfa4a24d4c131bd765deec3/grunt/tasks/server-diff.js#L11-L18
* The javascript caching seems stable enough to extend the life of the files beyond one week, this is now configurable at the terminal only. Each cached file can be 2.5mb+ in size, there is one per course and they will live on the system for the defined length of time.

### Testing
1. Setup
```js
git clone https://github.com/adaptlearning/adapt_framework/ 3547-diff
cd 3547-diff
git checkout issue/3547
mkdir build/
cp -r src/course/ build/course/
grunt server-diff:dev --cacheage=2419200000
```
2. Rediff no changes
```js
grunt server-diff:dev --cacheage=2419200000
```
3. Rediff with changes
```js
grunt server-diff:dev --cacheage=2419200000
```
4. Switch from development to production
```js
grunt server-diff:build --cacheage=2419200000
```

### Caveats
* A full rebuild `grunt server-build` is required for:
  * Files copied into the build output from plugins that are subsequently removed will not be cleared up
  * Files copied in by spoor at scorm 1.2 will not be cleared when switching to scorm 2004


chris-steele:
Looks slick 👍 @oliverfoster just remind me what was the reason for not using `newer:` on `required` files?

oliverfoster:
> Looks slick 👍 @oliverfoster just remind me what was the reason for not using `newer:` on `required` files?

Ensuring that the `@@` string replacement directives for the `replace` task work properly every time. Replace can only function on freshly copied original files where the directives haven't yet been replaced.

Like: https://github.com/adaptlearning/adapt-contrib-spoor/blob/df90b8dacbe8c20e12e1d57dfd1cdfc37de82903/required/index.html#L2

cahirodoherty-learningpool (approved):
👀

taylortom (approved):
Works well for me :fire:

chris-steele (approved):

PR #3612, opened 2024-10-03
Fix: Tightening glob pattern for FW-level tests

### LINKED ISSUE #3611: Tests: Glob to target FW tests was too loose
The glob pattern used to target FW tests was too loose. As a result, all plugin tests were being picked up (undoing the targeted approach that was desired)

cahirodoherty-learningpool:

Fixes #3611 

### Fix
* Tightening glob pattern for FW-level tests

### DIFF
--- a/src/course/config.json
+++ b/src/course/config.json
@@ -79,7 +79,7 @@
   "build": {
     "strictMode": true,
     "includes": [
-      "adapt-contrib-according",
+      "adapt-contrib-accordion",
       "adapt-contrib-assessment",
       "adapt-contrib-assessmentResults",
       "adapt-contrib-blank",
--- a/src/course/config.json
+++ b/src/course/config.json
@@ -77,6 +77,35 @@
     "_limitToSelector": ""
   },
   "build": {
-    "strictMode": true
+    "strictMode": true,
+    "includes": [
+      "adapt-contrib-according",
+      "adapt-contrib-assessment",
+      "adapt-contrib-assessmentResults",
+      "adapt-contrib-blank",
+      "adapt-contrib-bookmarking",
+      "adapt-contrib-boxMenu",
+      "adapt-contrib-branching",
+      "adapt-contrib-confidenceSlider",
+      "adapt-contrib-gmcq",
+      "adapt-contrib-glossary",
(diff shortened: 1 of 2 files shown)


src/course/config.json
```diff
@@ -77,6 +77,35 @@
     "_limitToSelector": ""
   },
   "build": {
-    "strictMode": true
+    "strictMode": true,
```
oliverfoster, line 80: I'm not sure about adding these in the default course. They shouldn't be necessary in course where all of the plugins are to be compiled. The vanilla course is used a lot for testing new plugins and these would now newly need to be updated by hand when doing that.

cahirodoherty-learningpool: > I'm not sure about adding these in the default course. They shouldn't be necessary in course where all of the plugins are to be compiled. The vanilla course is used a lot for testing new plugins and these would now newly need to be updated by hand when doing that.

I've removed these again, but as a result our PR test runner (and scheduled one) don't actually know which plugins to run tests against. There isn't any reference point for them to check so the only automated tests are those in the FW test folder

oliverfoster: Wider glob if no build includes or empty build.includes?

cahirodoherty-learningpool: That did the trick. Thanks Oli

oliverfoster (approved):

test.js
```diff
@@ -87,16 +87,24 @@ function populateTestFiles(testFormat) {
...
+  const plugins = config?.build?.includes;
   const globSuffix = testFormat === 'e2e' ? 'e2e/*.cy.js' : 'unit/*.js';
 
-  const testFiles = plugins.map(plugin => {
-    return `**/${plugin}/test/${globSuffix}`;
-  });
+  // Set a wider glob as default and limit to included plugins if that is set
+  let testFiles = [`**/*/test/${globSuffix}`];
+
+  if (plugins) {
```
oliverfoster, line 96:
```suggestion
  if (plugins?.length) {
```

This will return undefined or 0 if plugins isn't defined or length is 0. Might be better just to step over and empty build.includes array.

ethan-lp (approved):

lemmyadams (approved):

PR #3645, opened 2025-01-22
Removed now unnecessary supplementEntry

oliverfoster:
* Due to https://github.com/cgkineo/adapt-migrations/pull/22 supplementEntry is now no longer a passable option


joe-allen-89 (approved):
👍

PR #3672, opened 2025-04-14
Fix: Path resolution to root in migrations scripts (fixes #3671)

### LINKED ISSUE #3671: Migations path.join vs path.resolve giving `C:/path/C:/path`.
> Have just come across an AAT issue related to this PR - specifically [these lines](https://github.com/adaptlearning/adapt_framework/blob/master/grunt/helpers/Framework.js#L45-L47). The AAT passes absolute paths, so we were getting something like `C:/path/C:/path`.

Switching from `path.join` to `path.resolve` fixes this for me

If we should be passing the `rootDir` somehow via the AAT, let me know (not sure if this is hooked up to the CLI API)

_Originally posted by @taylortom in https://github.com/adaptlearning/adapt_framework/issues/3668#issuecomment-2797608883_

oliverfoster:
fixes #3671 

### Fix
* Changed from path.join to path.resolve

image

### DIFF
--- a/grunt/helpers/Framework.js
+++ b/grunt/helpers/Framework.js
@@ -42,9 +42,9 @@ class Framework {
     /** @type {string} */
     this.rootPath = rootPath.replace(/\\/g, '/');
     /** @type {string} */
-    this.outputPath = path.join(this.rootPath, outputPath).replace(/\\/g, '/');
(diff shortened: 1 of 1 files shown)


taylortom (approved):
Thanks for jumping on that one so quickly!

cahirodoherty-learningpool (approved):
👀

PR #3680, opened 2025-05-01
New: customscriptsdir added to migrations grunt task (fixes #3677)

### LINKED ISSUE #3677: Add the ability to pass in a folder containing migration scripts.
Currently the only way to include migration scripts is within a plugin, it would be useful, especially in the authoring tool, to be able to have a seperate folder containing multiple scripts, for example updating the start controller on course import.

joe-allen-89:
Adding the customscriptdir is to allow a custom folder to be specified, this allows us to load custom migration scripts for the authoring tool. 

Changed the filNameIncludes from .includes to RegExp, this is to make it more accurate when targetting a specific file. For example when using .includes the flag --file=text will also pull through textInput. 

#3677 

### Fix
* match fileNameIncludes changed to RegExp

### New
* customscriptsdir added to migrations grunt task (#3677)

### DIFF
--- a/grunt/tasks/migration.js
+++ b/grunt/tasks/migration.js
@@ -32,6 +32,7 @@ module.exports = function(grunt) {
     const next = this.async();
     const buildConfig = Helpers.generateConfigData();
     const fileNameIncludes = grunt.option('file');
+    const customScriptsDir = grunt.option('customscriptsdir');
 
     (async function() {
       const migrations = await import('adapt-migrations');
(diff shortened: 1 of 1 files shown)


oliverfoster:
```js
minimatch(filePath, '**/' + fileNameIncludes)
```

This should do what you're doing with the rented Vs includes already. It uses globs to define the regex.

joe-allen-89:
My thought process was to avoid any potential issue with naming custom tasks in the authoring tool. Regex change reveretd.

oliverfoster:
There is currently no API for getting all of the existing tasks and running a specific one (other than --file). If you need the AAT to be able do that then you can add new stuff in here for that specific purpose?

`grunt migration:gettasks` to output JSON or such?

oliverfoster (approved):

PR #3822, opened 2026-08-04
Fix: Raise minimum Node version to 22 (fixes #3821)

### LINKED ISSUE #3821: Fix "crypto is not defined" build failure on Node 16 and 18 introduced in v5.56.0

Since v5.56.0, `grunt build` fails on Node 16 and Node 18 with `crypto is not defined`. This includes Node 18, which is the framework's own declared minimum in _package.json_ (`engines.node: ">=18"`).

v5.56.2 is still the current release, so anyone running `adapt create course` today on Node 16 or 18 gets a course that will not build. This affects command-line and CI users as well as authoring tool installs. It has been broken since 2026-04-13.

### Your environment

* Framework v5.56.0 through v5.56.2 (current release)
* Node 16.14.0, 16.20.2, 18.20.8 — all fail. Node 20.20.0, 22.22.0, 24.13.0 — all pass
* Not browser or device specific; the build never completes

### Steps to reproduce

Two commands, no authoring tool involved:

```bash
nvm use 18
npx adapt-cli@3.4.0 create course my-course v5.56.2 true
cd my-course && npx grunt build
```

`adapt create course` itself succeeds. The failure is on the very next step the CLI tells you to run.

Switching to Node 20 or above in the same directory, with nothing else changed, then builds successfully.

### Expected behaviour

The build completes and produces a working _adapt.min.js_, as it does on v5.55.2 and as it does on Node 20+.

### Actual behaviour

Exit code 3:

```
Warning: crypto is not defined Use --force to continue.

Aborted due to warnings.
```

Every build mode is affected — `build`, `build:dev`, `build:prod`, `server-build`, `dev` and `server` — because the error happens when the task loads, before any minification decision is made. Tasks that do not pull in the `javascript` task (`help`, `check-json`, `clean`, `less`) still work.

**Cause.** #3791 (Dependabot, merged 2026-04-13, shipped in v5.56.0) bumped `@rollup/plugin-terser` from `^0.4.0` to `^1.0.0`, which pulls in `serialize-javascript@7`. That version calls the global `crypto.getRandomValues()` at module scope, where version 6 used `require('randombytes')` instead. The global `crypto` object does not exist in Node before version 19/20. `grunt/tasks/javascript.js:15` requires the plugin when the task loads, so it throws immediately.

**Why it went unnoticed for nearly four months.** CI pins `node-version: 'lts/*'` in both _test.yml_ and _build-and-deploy.yml_, so Node 16 and 18 are never exercised despite the declared minimum of 18. And `npm install` only warns about the mismatch, then exits successfully:

```
npm warn EBADENGINE   package: '@rollup/plugin-terser@1.0.0',
npm warn EBADENGINE   required: { node: '>=20.0.0' },
npm warn EBADENGINE   current: { node: 'v18.20.8' }
npm warn EBADENGINE   package: 'serialize-javascript@7.0.5',
npm warn EBADENGINE   required: { node: '>=20.0.0' },
```

Nothing fails until build time.

### Please do not use `--force`

Grunt suggests it, and it makes things worse. `grunt build:prod --force` exits **0** with "Done, but with warnings" and writes a build folder that looks complete — _index.html_, compiled LESS, _scriptLoader.js_ — but with **no _adapt.min.js_**. _scriptLoader.js_ still asks for it, so the course returns a 404 and never starts, while CI reports success and the authoring tool reports a successful publish. That is worse than the visible failure, so it is worth flagging to anyone triaging this.

### Suggested fix

Keep the security bump and raise `engines.node` to `>=22`.

A revert is not free. #3791 was a security fix: `serialize-javascript` 6.0.2 carries GHSA-5c6j-r48x-rmvq (high, remote code execution, patched in 7.0.3) and GHSA-qj8w-gfj5-8c6v (moderate, denial of service, patched in 7.0.5). Pinning back reintroduces both, and Dependabot would simply reopen the PR.

Raising the floor to `>=20` is not the answer either, because Node 20 reached end of life in March 2026. Every Node version this bug breaks is already unsupported — v16 since August 2023, v18 since March 2025, v20 since March 2026. The current LTS lines are 22 and 24, so `>=22` is the floor that matches what Node actually supports.

There is no way to fix this from inside the framework. `@rollup/plugin-terser@1.0.0` runs terser in a worker thread pool, and each worker requires `serialize-javascript` again in its own global context. Setting `globalThis.crypto` at the top of _Gruntfile.js_ does not reach the workers — tested, and it just moves the same error from load time into `javascript:compile`. Pushing `--require` onto `process.execArgv` does not reach them either. Only a process-level `NODE_OPTIONS=--require` preload propagates, which has to come from outside the repo.

Bumping forward does not help either: `serialize-javascript` 7.0.6 and 7.0.7 both still use the global, and `@rollup/plugin-terser` has no release above 1.0.0.

If a transitional patch is wanted for people still on older Node, `"overrides": { "serialize-javascript": "^6.0.2" }` does work — verified building successfully on Node 16.14.0 — at the cost of reintroducing both advisories.

Separately, it is worth replacing `node-version: 'lts/*'` in CI with an explicit matrix on the supported lines. That is the change that stops this recurring: `lts/*` silently tracks whatever is newest, which is exactly why a declared minimum of 18 went untested for nearly four months.

Surfaced via https://github.com/adaptlearning/adapt_authoring/issues/2847.

Posted via collaboration with Claude Code

swashbuck:
Fixes #3821

### Fix
* Raises `engines.node` to `>=22`, so the declared minimum matches a Node version where `globalThis.crypto` exists and `grunt build` actually completes. Since v5.56.0 the declared minimum of `>=18` named a version the build cannot run on at all — `@rollup/plugin-terser@1.0.0` pulls `serialize-javascript@7`, which calls the global `crypto.getRandomValues()` at module scope, and _grunt/tasks/javascript.js_ requires the plugin at task-load time. Both of those packages already declare `>=20.0.0` themselves.
* Updates the Requirements line in the README, which still said Node >=16 and was already two floors out of date.

### Testing
1. On Node 22 or 24, `npm ci && npx -y -p adapt-cli adapt install && npx grunt build` completes and produces _adapt.min.js_.
2. On Node 18 or 20, `npm install` now reports the engine mismatch for the framework itself (`npm warn EBADENGINE`) rather than only for its dependencies, so the incompatibility is visible at install time instead of surfacing as `Warning: crypto is not defined` at build time.

### Notes for reviewers
* **A revert is not an option here.** #3791 was a security bump: `serialize-javascript` 6.0.2 carries GHSA-5c6j-r48x-rmvq (high, remote code execution, patched in 7.0.3) and GHSA-qj8w-gfj5-8c6v (moderate, denial of service, patched in 7.0.5). Pinning back reintroduces both and Dependabot would reopen the PR. There is also no in-repo shim: `@rollup/plugin-terser@1.0.0` runs terser in a worker thread pool and each worker requires `serialize-javascript` again in its own global context, so setting `globalThis.crypto` in _Gruntfile.js_ does not reach the workers — only a process-level `NODE_OPTIONS=--require` preload propagates.
* **Why 22 and not 20.** Every Node version this bug breaks is already end of life — v16 since August 2023, v18 since March 2025, v20 since March 2026. Setting the floor to `>=20` would pin to a line that no longer receives security patches. 22 and 24 are the current LTS lines.
* **Release type is a maintainer call.** This is submitted as `Fix` on the grounds that nothing works today that stops working — builds on Node 18 and 20 already fail. If you would rather the support-floor change drive a major version, the squash-merge subject needs to be `Breaking` instead.
* Worth stating the new floor in the release notes and the authoring tool compatibility matrix. Legacy `adapt_authoring` 0.11.5 declares `engines: { node: "16 || 18" }`, so it sits entirely below this floor; its declared range is advisory rather than enforced, and a fresh install has been verified to run and publish successfully on Node 22.
* Two related follow-ups deliberately left out of this PR, both mentioned in #3821 and neither needed to fix it. Happy to raise either separately if wanted. The CI workflows still pin `node-version: 'lts/*'`, so the declared floor is never the version actually exercised — that is what let a minimum of 18 go untested for nearly four months. And `Framework.getSchemas()` does not forward `warn`, so plugin warnings fall back to `console.warn` on stderr, which is why an unrelated warning was mistaken for the failure in this bug's reports.

Posted via collaboration with Claude Code


simondate:
~~Is there any potential side effects introduced by going up to Node 22? I thought the AAT legacy branch was to be kept the same? Couldn't we just keep that product on a stable Node 16/18 version of the FW?~~

EDIT: Seems like the AAT works fine with the newer node version (from my brief testing of features and builds).

Could be some other side effect with in the FW though?

swashbuck:
> Is there any potential side effects introduced by going up to Node 22?
> 
> I thought the AAT legacy branch was to be kept the same? Couldn't we just keep that product on a stable Node 16/18 version of the FW?

@oliverfoster @taylortom Any thoughts here?

simondate (approved):

oliverfoster (approved):

